PR #4, opened 2019-04-04
[README] minor build steps

lijiansong:
specify gcc and g++ version


lattner:
@joker-eph FYI

joker-eph:
@lijiansong : do you mind giving a bit more information on your setup? I would expect this to be close to the default behavior of CMake

PR #8, opened 2019-04-04
Fix the link to the next chapter

sylvestre:


joker-eph:
Thanks for the fix!

Feel free to report bugs though.

(this particular case was already reported and we have a fix that should be pushed soon)

PR #12, opened 2019-04-04
Add ninja to the build requirements

ddutta:
Its a small doc change since I needed to install ninja (am a llvm/ninja newbie).


ddutta:
I have signed the CLA already but did an anonymous push to my fork because it forced me to do so.

PR #14, opened 2019-04-05
Fix typo in tutorials

mshr-h:
Fix small typos in the tutorials.


joker-eph:
Thanks for the fix!

(we have an internal fix for these, it should be mirrored soon).

Issue #6, opened 2019-04-04
Fix LangRef.md, remove erroneous characters


madhur13490:
The "Context" section has some erroneous characters (may be an effect of copying text from LaTex). Better to remove them to have a clean reading experience.

joker-eph:
Fixed in 792d9b55670abc2169ae93408f4bdfa2825ce0e1

PR #15, opened 2019-04-05
Fix links in LangRef.md

mshr-h:
As mentioned in #6


joker-eph:
Thanks for the fix!

We know about this issue where our internal markdown formatter is different from the one used by GitHub. Removing the anchor may break links that we have internally using these anchors, this is why I haven't removed them (yet). We need to audit and replace the links using them before removing them here.

I'm closing this PR, but we'll track the issue your opened.

PR #22, opened 2019-04-19
Fix build for MSVC

loic-joly-sonarsource:
This contains changes to make MLIR build on MSVC(19). It also require one change in LLVM (https://github.com/loic-joly-sonarsource/llvm-project/pull/1). 

Currently, with thos changes, the release mode is built correctly (there are some issues with the debug mode not picking the command line parameters). When running check-mlir, we have:
```
57>  Expected Passes    : 116
57>  Unexpected Failures: 25
```


jpienaar:
I believe this has now been addressed in various changes post. Closing now, thanks!

PR #23, opened 2019-04-20
docs: minor spelling tweaks

brettkoonce:


joker-eph:
Thanks for the fixes!

We don't accept contributions right now (mostly for tooling reason), but we should soon. I'll leave this PR open here until we are set-up to process it.

brettkoonce:
🥳

PR #40, opened 2019-07-02
Introduce parser library method to parse list of region arguments

bondhugula:
- introduce parseRegionArgumentList (similar to parseOperandList) to parse a
  list of region arguments with a delimiter
- allows defining custom parse for op's with multiple/variadic number of
  region arguments
- use this on the gpu.launch op (although the latter has a fixed number
  of region arguments)

### DIFF
--- a/include/mlir/IR/OpImplementation.h
+++ b/include/mlir/IR/OpImplementation.h
@@ -315,8 +315,8 @@ public:
   /// Parse a single operand.
   virtual ParseResult parseOperand(OperandType &result) = 0;
 
-  /// These are the supported delimiters around operand lists, used by
-  /// parseOperandList.
+  /// These are the supported delimiters around operand lists and region
+  /// argument lists, used by parseOperandList and parseRegionArgumentList.
   enum class Delimiter {
     /// Zero or more operands with no delimiters.
     None,
@@ -410,9 +410,24 @@ public:
                                           ArrayRef<OperandType> arguments,
                                           ArrayRef<Type> argTypes) = 0;
 
-  /// Parse a region argument.  Region arguments define new values, so this also
-  /// checks if the values with the same name has not been defined yet.
+  /// Parse a region argument.  Region arguments define new values; so this also
+  /// checks if values with the same name have not been defined yet.
   virtual ParseResult parseRegionArgument(OperandType &argument) = 0;
+
+  /// Parse zero or more region arguments with a specified surrounding
+  /// delimiter, and an optional required argument count. Region arguments
+  /// define new values; so this also checks if values with the same names have
+  /// not been defined yet.
+  virtual ParseResult
+  parseRegionArgumentList(SmallVectorImpl<OperandType> &result,
+                          int requiredOperandCount = -1,
+                          Delimiter delimiter = Delimiter::None) = 0;
+  virtual ParseResult
+  parseRegionArgumentList(SmallVectorImpl<OperandType> &result,
+                          Delimiter delimiter) {
+    return parseRegionArgumentList(result, /*requiredOperandCount=*/-1,
+                                   delimiter);
+  }
 
   /// Parse a region argument if present.
   virtual ParseResult parseOptionalRegionArgument(OperandType &argument) = 0;
@@ -486,6 +501,14 @@ public:
     result.append(types.begin(), types.end());
     return success();
   }
+
+private:
+  /// Parse either an operand list or a region argument list depending on
+  /// whether `operand' is true.
+  ParseResult parseOperandOrRegionArgList(SmallVectorImpl<OperandType> &result,
+                                          bool isOperandList,
+                                          int requiredOperandCount,
+                                          Delimiter delimiter);
 };
 
 } // end namespace mlir
--- a/lib/GPU/IR/GPUDialect.cpp
+++ b/lib/GPU/IR/GPUDialect.cpp
@@ -221,12 +221,13 @@ parseSizeAssignment(OpAsmParser *parser,
                     MutableArrayRef<OpAsmParser::OperandType> sizes,
                     MutableArrayRef<OpAsmParser::OperandType> regionSizes,
                     MutableArrayRef<OpAsmParser::OperandType> indices) {
-  if (parser->parseLParen() || parser->parseRegionArgument(indices[0]) ||
-      parser->parseComma() || parser->parseRegionArgument(indices[1]) ||
-      parser->parseComma() || parser->parseRegionArgument(indices[2]) ||
-      parser->parseRParen() || parser->parseKeyword("in") ||
-      parser->parseLParen())
+  assert(indices.size() == 3 && "space for three indices expected");
+  SmallVector<OpAsmParser::OperandType, 3> args;
+  if (parser->parseRegionArgumentList(args, /*requiredOperandCount=*/3,
+                                      OpAsmParser::Delimiter::Paren) ||
+      parser->parseKeyword("in") || parser->parseLParen())
     return failure();
+  std::move(args.begin(), args.end(), indices.begin());
 
   for (int i = 0; i < 3; ++i) {
     if (i != 0 && parser->parseComma())
--- a/lib/Parser/Parser.cpp
+++ b/lib/Parser/Parser.cpp
@@ -3344,6 +3344,16 @@ public:
   ParseResult parseOperandList(SmallVectorImpl<OperandType> &result,
                                int requiredOperandCount = -1,
                                Delimiter delimiter = Delimiter::None) override {
+    return parseOperandOrRegionArgList(result, /*isOperandList=*/true,
+                                       requiredOperandCount, delimiter);
+  }
+
+  /// Parse zero or more SSA comma-separated operand references with a specified
+  /// surrounding delimiter, and an optional required operand count.
+  ParseResult
+  parseOperandOrRegionArgList(SmallVectorImpl<OperandType> &result,
+                              bool isOperandList, int requiredOperandCount = -1,
+                              Delimiter delimiter = Delimiter::None) {
     auto startLoc = parser.getToken().getLoc();
 
     // Handle delimiters.
@@ -3382,10 +3392,11 @@ public:
     // Check for zero operands.
     if (parser.getToken().is(Token::percent_identifier)) {
       do {
-        OperandType operand;
-        if (parseOperand(operand))
+        OperandType operandOrArg;
+        if (isOperandList ? parseOperand(operandOrArg)
+                          : parseRegionArgument(operandOrArg))
           return failure();
-        result.push_back(operand);
+        result.push_back(operandOrArg);
       } while (parser.consumeIf(Token::comma));
     }
 
@@ -3537,6 +3548,14 @@ public:
     if (parser.getToken().isNot(Token::percent_identifier))
       return success();
     return parseRegionArgument(argument);
+  }
+
+  ParseResult
+  parseRegionArgumentList(SmallVectorImpl<OperandType> &result,
+                          int requiredOperandCount = -1,
+                          Delimiter delimiter = Delimiter::None) override {
+    return parseOperandOrRegionArgList(result, /*isOperandList=*/false,
+                                       requiredOperandCount, delimiter);
   }
 
   //===--------------------------------------------------------------------===//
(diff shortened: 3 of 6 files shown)


include/mlir/IR/OpImplementation.h
```diff
@@ -404,10 +404,25 @@ class OpAsmParser {
...
+  /// delimiter, and an optional required argument count. Region arguments
+  /// define new values; so this also checks if values with the same names have
+  /// not been defined yet.
+  virtual ParseResult
+  parseRegionArgumentList(SmallVectorImpl<OperandType> &result,
+                          int requiredOperandCount = -1,
+                          Delimiter delimiter = Delimiter::None) = 0;
+  virtual ParseResult
+  parseRegionArgumentList(SmallVectorImpl<OperandType> &result,
+                          Delimiter delimiter = Delimiter::None) {
```
River707: Drop the default argument here, it won't be able to resolve `parseRegionArgumentList(result)` otherwise.

include/mlir/IR/OpImplementation.h
```diff
@@ -480,6 +495,14 @@ class OpAsmParser {
     result.append(types.begin(), types.end());
     return success();
   }
+
+private:
+  /// Parse either an operand list or a region argument list depending on
+  /// whether `operand' is true.
+  ParseResult parseOperandOrRegionArgList(SmallVectorImpl<OperandType> &result,
+                                          bool operand,
```
River707: I'd change 'operand' to 'isOperandList' to be more descriptive.

lib/Parser/Parser.cpp
```diff
@@ -3360,10 +3370,12 @@ class CustomOpAsmParser : public OpAsmParser {
     // Check for zero operands.
     if (parser.getToken().is(Token::percent_identifier)) {
       do {
-        OperandType operand;
-        if (parseOperand(operand))
-          return failure();
-        result.push_back(operand);
+        OperandType operandOrArg;
+        if (operand && parseOperand(operandOrArg))
```
River707: if (isOperandList ? parseOperand(...) : parseRegionArgument(...))?

include/mlir/IR/OpImplementation.h
bondhugula: Done.

include/mlir/IR/OpImplementation.h
bondhugula: Done, thanks.

lib/Parser/Parser.cpp
bondhugula: Done, thanks.

joker-eph:
Please squash and force push to keep your changes in a single commit

include/mlir/IR/OpImplementation.h
```diff
@@ -410,10 +410,25 @@ class OpAsmParser {
...
+  /// checks if values with the same name have not been defined yet.
   virtual ParseResult parseRegionArgument(OperandType &argument) = 0;
 
+  /// Parse zero or more region arguments with a specified surrounding
+  /// delimiter, and an optional required argument count. Region arguments
+  /// define new values; so this also checks if values with the same names have
+  /// not been defined yet.
+  virtual ParseResult
+  parseRegionArgumentList(SmallVectorImpl<OperandType> &result,
+                          int requiredOperandCount = -1,
```
jpienaar: Is there a test case for Delimiter::None?

bondhugula: Actually, there isn't. The only way I can think of testing it is by introducing an op that has such a custom form. AFAIK, the gpu.launch is the only op in MLIR that has multiple region arguments; I have another experimental one on the side, but that isn't meant for mainstream yet and it too uses a parenthesis enclosed list.

bondhugula:
> Please squash and force push to keep your changes in a single commit

Done.

include/mlir/IR/OpImplementation.h
River707: You should be able to test by adding a fake operation to the TestDialect. I added that specifically for the things that don't have a use case in-tree.

https://github.com/tensorflow/mlir/blob/master/test/lib/TestDialect/TestOps.td

bondhugula: Just looked at TestOps.td:

bondhugula: Just looked at TestOps.td. Does "Arguments< ..>" there mean operands (cf. appended lines for example)? An operand is expected to have a def prior to its use, while in contrast a region argument is expected not to have a def at all. If "Arguments" in the .td means operands, what are the region arguments referred to? It looks like the naming is confusing here?

> def OpD : TEST_Op<"op_d">, Arguments<(ins I32:$arg)>, Results<(outs I32:$res)>;
> def OpE : TEST_Op<"op_e">, Arguments<(ins I32:$arg)>, Results<(outs I32:$res)>;
> def OpF : TEST_Op<"op_f">, Arguments<(ins I32:$arg)>, Results<(outs I32:$res)>;
> def OpG : TEST_Op<"op_g">, Arguments<(ins I32:$arg)>, Results<(outs I32:$res)>;

bondhugula: > You should be able to test by adding a fake operation to the TestDialect. I added that specifically for the things that don't have a use case in-tree.
> 
> https://github.com/tensorflow/mlir/blob/master/test/lib/TestDialect/TestOps.td

Is there an example of a test op that has a parse method for its custom form? I couldn't immediately find any. (Just adding a test op that takes multiple region arguments isn't sufficient here.)

jpienaar: > Just looked at TestOps.td. Does "Arguments< ..>" there mean operands (cf. appended lines for example)? An operand is expected to have a def prior to its use, while in contrast a region argument is expected not to have a def at all. If "Arguments" in the .td means operands, what are the region arguments referred to? It looks like the naming is confusing here?
> 

Arguments here refer to operands and attributes (consult the docs directory for ODS).

> Is there an example of a test op that has a parse method for its custom form? I couldn't immediately find any. (Just adding a test op that takes multiple region arguments isn't sufficient here.)

There isn't feel free to add one that exercises this path.

include/mlir/IR/OpImplementation.h
```diff
@@ -410,10 +410,25 @@ class OpAsmParser {
...
+                          Delimiter delimiter = Delimiter::None) = 0;
+  virtual ParseResult
+  parseRegionArgumentList(SmallVectorImpl<OperandType> &result,
+                          Delimiter delimiter) {
```
jpienaar: I'd prefer to not have this overload, it doesn't add much. If useful, we could interchange delimiter and requiredOperandCount based on usage and that would then cover this case.

include/mlir/IR/OpImplementation.h
```diff
@@ -486,6 +501,14 @@ class OpAsmParser {
     result.append(types.begin(), types.end());
     return success();
   }
+
+private:
+  /// Parse either an operand list or a region argument list depending on
+  /// whether `operand' is true.
+  ParseResult
+  parseOperandOrRegionArgList(SmallVectorImpl<OperandType> &result,
+                              bool isOperandList, int requiredOperandCount = -1,
```
jpienaar: These default arguments should not be needed. This function is only invoked internally from others that already have the parameters specified.

lib/GPU/IR/GPUDialect.cpp
```diff
@@ -220,12 +220,13 @@ parseSizeAssignment(OpAsmParser *parser,
...
-      parser->parseComma() || parser->parseRegionArgument(indices[2]) ||
-      parser->parseRParen() || parser->parseKeyword("in") ||
-      parser->parseLParen())
+  assert(indices.size() == 3 && "space for three indices expected");
+  SmallVector<OpAsmParser::OperandType, 3> args;
+  if (parser->parseRegionArgumentList(args, /*requiredOperandCount=*/3,
+                                      OpAsmParser::Delimiter::Paren) ||
+      parser->parseKeyword("in") || parser->parseLParen())
     return failure();
+  std::copy(args.begin(), args.end(), indices.begin());
```
jpienaar: Could this copy be avoided by parsing into indices directly?

lib/Parser/Parser.cpp
```diff
@@ -3332,6 +3332,16 @@ class CustomOpAsmParser : public OpAsmParser {
...
                                Delimiter delimiter = Delimiter::None) override {
+    return parseOperandOrRegionArgList(result, /*isOperandList=*/true,
+                                       requiredOperandCount, delimiter);
+  }
+
+  /// Parse zero or more SSA comma-separated operand references with a specified
+  /// surrounding delimiter, and an optional required operand count.
+  ParseResult
+  parseOperandOrRegionArgList(SmallVectorImpl<OperandType> &result,
+                              bool isOperandList, int requiredOperandCount = -1,
```
jpienaar: I'm not a fan of isOperandList: its enabling code reuse but at the cost of the modelling IMHO. Could you try a different refactoring that enables the reuse (most of the reuse) but avoids the flag & if below?

include/mlir/IR/OpImplementation.h
bondhugula: In the ODS, I see "operands and attributes" but nothing on regions and thus region arguments. So any op that has a region can't be tablegen'd currently(?), and this test op will have to be done via TestDialect.h and TestDialect.cpp?

lib/GPU/IR/GPUDialect.cpp
bondhugula: Unfortunately, no AFAIU. The 'indices' is a MutableArrayRef, while the parse method takes a SmallVector. The former's length can't be changed although its contents can be, and so there can't be an implicit conversion to the latter.

include/mlir/IR/OpImplementation.h
bondhugula: I was just following parseOperandList, which is ditto in spirit, and has this overload. (If we don't want this, we should remove it there too to be consistent IMO. Let me know.)

lib/Parser/Parser.cpp
bondhugula: I had the same thoughts. But I couldn't find another refactoring that was clean and allowed all / most of the reuse. If we think about it, the parsing is really the same for both except for how each element is treated (for one, the SSA use has to be resolved, and for the other, the absence of def's to the same name has to be checked); so this made sense.

include/mlir/IR/OpImplementation.h
bondhugula: Thanks, done.

include/mlir/IR/OpImplementation.h
jpienaar: There are regions there, but the region arguments one seems to be pending still, so via C++ SGTM.

include/mlir/IR/OpImplementation.h
jpienaar: That could be done in a follow up yes, not along with this.

bondhugula: Thanks.

lib/GPU/IR/GPUDialect.cpp
ftynse: Can you make this function take `SmallVectorImpl &` instead, forward it to the parse method, and then concatenate vectors using move iterators at the top call site?

bondhugula: The declaration and concatenation of six different vectors is going to be just too verbose at the call site! Did you want that?

bondhugula: Also, IIUC, using move iterators to concatenate three SmallVectors won't prevent any copies - in fact, it'll lead to 14 copies here instead of the six right now. (SmallVectors store elements contiguously in a single allocation.) Using move on the OperandType itself isn't more efficient since all its fields are POD or POD-like, and one could use move anyway separately on these three here.

bondhugula:
Any further comments here?

jpienaar:
Just as mentioned a test as this introducing a default case that isn't tested.

bondhugula:
Looks like there's a typo in your comment - sorry, I can't parse it :) Did you mean testing the default delimiter? This will require doing things outside of the tablegen approach -- adding print, parse, verify, and the c++ class for the test op. Did you want this?

bondhugula:
> Just as mentioned a test as this introducing a default case that isn't tested.

Done. PTAL. (added a test op to test this case).

include/mlir/IR/OpImplementation.h
bondhugula: Done now (via TestDialect.cpp).

Issue #42, opened 2019-07-05
Add spv.CompositeExtract


antiagainst:
The spec for this op is at https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpCompositeExtract.

We use Op Definition Spec (ODS) to describe the op, the doc for ODS is here: https://github.com/tensorflow/mlir/blob/master/g3doc/OpDefinitions.md

Useful SPIR-V ODS definitions: https://github.com/tensorflow/mlir/blob/34f7a2d9f8c00f596990ee5385cda2561d8fe202/include/mlir/SPIRV/SPIRVBase.td#L86

It would be nice to define the custom assembly format to look like `std.extract_element`: https://github.com/tensorflow/mlir/blob/34f7a2d9f8c00f596990ee5385cda2561d8fe202/lib/StandardOps/Ops.cpp#L1579-L1594
test: https://github.com/tensorflow/mlir/blob/34f7a2d9f8c00f596990ee5385cda2561d8fe202/test/IR/core-ops.mlir#L357-L368

So for `spv.CompositeExtract`, we will have something like
```
spv.CompositeExtract %composite[%index0, %index1, ...] : <extracted-type>
```

It would be nice to implement the validation rules specified in the op description. You may want to use `spv.struct`, `spv.array` and other types for writing tests.

See the code for adding `spv.Load/spv.Store` as reference: https://github.com/tensorflow/mlir/commit/0c0217d5a76a8a10cc587b6dfd97a40b4d386243

antiagainst:
@denis0x0D is interested in taking this. @denis0x0D, please let me know if you have any questions. :)

denis0x0D:
@antiagainst, I have some thoughts about implementation and your commments, can you please take a look and please fix me if I'm wrong:

1. Seems like the ebnf form should look like this:
```
// <operation> ::= ssa-id `=`  `spv.CompositeExtract` ssa-use
//                 `[` integer-literal (`,` integer-literal)* `]`
//                  `:`  composite-type
```
Regarding to official documentation, the indices are literals and are not in SSA form.
For example:
```
func @composite_extract_2D_array() -> () {
  %0 = spv.Variable : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
  %1 = spv.Load "Function" %0 ["Volatile"] : !spv.array<4x!spv.array<4xf32>>
[...]

PR #44, opened 2019-07-09
[spirv] Add CompositeExtractOp operation.

denis0x0D:
Hello, this patch is addressing https://github.com/tensorflow/mlir/issues/42 issue.

Mostly all verifying checks are added to parser, for some reasons:
1. To get the result type of operation, we have to iterate all over the indices, seems like the better way to do it once, instead two times in parser and verifier.
2. To get the element type of the struct type we have to access it by index, to make it save the check for out of bounds access must be added and it seems like the better way do it once only in the parser.

BTW, I might miss something and do not see the possible side effects of this, please fix me if I'm wrong.

The verifier only checks that extracted type is not the same as composite type, to handle situation like this
`%0 = spv.CompositeExtract %arg0[] : !spv.array<4x!spv.array<4xf32>>`

The tests for different composite types added, also StructType is added to composite and spirv types.

@antiagainst can you please take a look. Thanks!

### DIFF
--- a/include/mlir/SPIRV/SPIRVBase.td
+++ b/include/mlir/SPIRV/SPIRVBase.td
@@ -65,6 +65,7 @@ def SPV_Dialect : Dialect {
 def SPV_IsPtrType : CPred<"$_self.isa<::mlir::spirv::PointerType>()">;
 def SPV_IsArrayType : CPred<"$_self.isa<::mlir::spirv::ArrayType>()">;
 def SPV_IsRTArrayType : CPred<"$_self.isa<::mlir::spirv::RuntimeArrayType>()">;
+def SPV_IsStructType : CPred<"$_self.isa<::mlir::spirv::StructType>()">;
 
 // See https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_types
 // for the definition of the following types and type categories.
@@ -79,14 +80,15 @@ def SPV_Vector : VectorOf<[SPV_Bool, SPV_Integer, SPV_Float]>;
 def SPV_AnyPtr : Type<SPV_IsPtrType, "any SPIR-V pointer type">;
 def SPV_AnyArray : Type<SPV_IsArrayType, "any SPIR-V array type">;
 def SPV_AnyRTArray : Type<SPV_IsRTArrayType, "any SPIR-V runtime array type">;
+def SPV_AnyStruct : Type<SPV_IsStructType, "any SPIR-V struct type">;
 
 def SPV_Numerical : AnyTypeOf<[SPV_Integer, SPV_Float]>;
 def SPV_Scalar : AnyTypeOf<[SPV_Numerical, SPV_Bool]>;
-def SPV_Aggregrate : AnyTypeOf<[SPV_AnyArray]>;
-def SPV_Composite: AnyTypeOf<[SPV_Vector, SPV_AnyArray]>;
+def SPV_Aggregrate : AnyTypeOf<[SPV_AnyArray, SPV_AnyStruct]>;
+def SPV_Composite: AnyTypeOf<[SPV_Vector, SPV_AnyArray, SPV_AnyStruct]>;
 def SPV_Type : AnyTypeOf<[
     SPV_Void, SPV_Bool, SPV_Integer, SPV_Float, SPV_Vector,
-    SPV_AnyPtr, SPV_AnyArray, SPV_AnyRTArray
+    SPV_AnyPtr, SPV_AnyArray, SPV_AnyRTArray, SPV_AnyStruct
   ]>;
 
 class SPV_ScalarOrVectorOf<Type type> :
--- a/include/mlir/SPIRV/SPIRVOps.td
+++ b/include/mlir/SPIRV/SPIRVOps.td
@@ -324,4 +324,48 @@ def SPV_VariableOp : SPV_Op<"Variable"> {
   let opcode = 59;
 }
 
+def SPV_CompositeExtractOp : SPV_Op<"CompositeExtract", [NoSideEffect]> {
+  let summary = "Extract a part of a composite object.";
+
+  let description = [{
+    Result Type must be the type of object selected by the last provided index.
+    The instruction result is the extracted object.
+
+    Composite is the composite to extract from.
+
+    Indexes walk the type hierarchy, potentially down to component granularity,
+    to select the part to extract. All indexes must be in bounds.
+    All composite constituents use zero-based numbering, as described by their
+    OpType… instruction.
+
+    ### Custom assembly form
+
+    ``` {.ebnf}
+    composite-extract-op ::= ssa-id `=` `spv.CompositeExtract` ssa-use
+                             `[` integer-literal (',' integer-literal)* `]`
+                             `:` composite-type
+    ```
+
+    For example:
+
+    ```
+    %0 = spv.Variable : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
+    %1 = spv.Load "Function" %0 ["Volatile"] : !spv.array<4x!spv.array<4xf32>>
+    %2 = spv.CompositeExtract %1[1 : i32] : !spv.array<4x!spv.array<4xf32>>
+    ```
+
+  }];
+
+  let arguments = (ins
+    SPV_Composite:$composite,
+    I32ArrayAttr:$indices
+  );
+
+  let results = (outs
+    SPV_Type:$component
+  );
+
+  let opcode = 81;
+}
+
 #endif // SPIRV_OPS
--- a/include/mlir/SPIRV/SPIRVTypes.h
+++ b/include/mlir/SPIRV/SPIRVTypes.h
@@ -22,6 +22,7 @@
 #ifndef MLIR_SPIRV_SPIRVTYPES_H_
 #define MLIR_SPIRV_SPIRVTYPES_H_
 
+#include "mlir/IR/StandardTypes.h"
 #include "mlir/IR/TypeSupport.h"
 #include "mlir/IR/Types.h"
 
@@ -52,9 +53,26 @@ enum Kind {
 };
 }
 
+// Provides a common base class between VectorType, SPIR-V ArrayType, SPIR-V
+// StructType.
+class CompositeType : public Type {
+public:
+  using Type::Type;
+
+  static bool classof(Type type) {
+    return (type.getKind() == TypeKind::Array ||
+            type.getKind() == TypeKind::Struct ||
+            type.getKind() == StandardTypes::Vector);
+  }
+
+  uint64_t getNumMembers() const;
+
+  Type getMemberType(uint64_t) const;
+};
+
 // SPIR-V array type
-class ArrayType
-    : public Type::TypeBase<ArrayType, Type, detail::ArrayTypeStorage> {
+class ArrayType : public Type::TypeBase<ArrayType, CompositeType,
+                                        detail::ArrayTypeStorage> {
 public:
   using Base::Base;
 
@@ -147,9 +165,8 @@ public:
 };
 
 // SPIR-V struct type
-class StructType
-    : public Type::TypeBase<StructType, Type, detail::StructTypeStorage> {
-
+class StructType : public Type::TypeBase<StructType, CompositeType,
+                                         detail::StructTypeStorage> {
 public:
   using Base::Base;
 
--- a/lib/SPIRV/SPIRVOps.cpp
+++ b/lib/SPIRV/SPIRVOps.cpp
@@ -38,6 +38,7 @@ using namespace mlir;
 static constexpr const char kAlignmentAttrName[] = "alignment";
 static constexpr const char kBindingAttrName[] = "binding";
 static constexpr const char kDescriptorSetAttrName[] = "descriptor_set";
+static constexpr const char kIndicesAttrName[] = "indices";
 static constexpr const char kValueAttrName[] = "value";
 static constexpr const char kValuesAttrName[] = "values";
 static constexpr const char kFnNameAttrName[] = "fn";
@@ -195,6 +196,115 @@ static LogicalResult verifyLoadStorePtrAndValTypes(LoadStoreOpTy op, Value *ptr,
 static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
   *printer << op->getName();
   printer->printOptionalAttrDict(op->getAttrs());
+}
+
+//===----------------------------------------------------------------------===//
+// spv.CompositeExtractOp
+//===----------------------------------------------------------------------===//
+
+static ParseResult parseCompositeExtractOp(OpAsmParser *parser,
+                                           OperationState *state) {
+  OpAsmParser::OperandType compositeInfo;
+  Attribute indicesAttr;
+  Type compositeType;
+  llvm::SMLoc attrLocation;
+  int32_t index;
+
+  if (parser->parseOperand(compositeInfo) ||
+      parser->getCurrentLocation(&attrLocation) ||
+      parser->parseAttribute(indicesAttr, kIndicesAttrName,
+                             state->attributes) ||
+      parser->parseColonType(compositeType) ||
+      parser->resolveOperand(compositeInfo, compositeType, state->operands)) {
+    return failure();
+  }
+
+  auto indicesArrayAttr = indicesAttr.dyn_cast<ArrayAttr>();
+  if (!indicesArrayAttr) {
+    return parser->emitError(
+        attrLocation,
+        "expected an 32-bit integer array attribute for 'indices'");
+  }
+
+  if (!indicesArrayAttr.size()) {
+    return parser->emitError(
+        attrLocation, "expected at least one index for spv.CompositeExtract");
+  }
+
+  Type resultType = compositeType;
+  for (auto indexAttr : indicesArrayAttr) {
+    if (auto indexIntAttr = indexAttr.dyn_cast<IntegerAttr>()) {
+      index = indexIntAttr.getInt();
+    } else {
+      return parser->emitError(
+                 attrLocation,
+                 "expexted an 32-bit integer for index, but found '")
+             << indexAttr << "'";
+    }
+    if (resultType.isa<spirv::CompositeType>()) {
+      if (index < 0 ||
+          static_cast<uint32_t>(index) >=
+              resultType.cast<spirv::CompositeType>().getNumMembers()) {
+        return parser->emitError(attrLocation, "index ")
+               << index << " out of bounds for " << resultType;
+      }
+      resultType = resultType.cast<spirv::CompositeType>().getMemberType(index);
+    } else {
+      return parser->emitError(attrLocation, "invalid type to extract from");
+    }
+  }
+
+  if (!resultType) {
+    return parser->emitError(attrLocation,
+                             "can not extract from composite type");
+  }
+
+  state->addTypes(resultType);
+  return success();
+}
+
+static void print(spirv::CompositeExtractOp compositeExtractOp,
+                  OpAsmPrinter *printer) {
+  *printer << spirv::CompositeExtractOp::getOperationName() << ' '
+           << *compositeExtractOp.composite() << compositeExtractOp.indices()
+           << " : " << compositeExtractOp.composite()->getType();
+}
+
+static LogicalResult verify(spirv::CompositeExtractOp compExOp) {
+  auto resultType = compExOp.composite()->getType();
+  auto indicesArrayAttr = compExOp.indices().dyn_cast<ArrayAttr>();
+
+  if (!indicesArrayAttr.size()) {
+    return compExOp.emitOpError(
+        "expexted at least one index for spv.CompositeExtractOp");
+  }
+
+  int32_t index;
+  for (auto indexAttr : indicesArrayAttr) {
+    if (!resultType) {
+      return compExOp.emitError("invalid type to extract from");
+    }
+
+    index = indexAttr.dyn_cast<IntegerAttr>().getInt();
+    if (resultType.isa<spirv::CompositeType>()) {
+      if (index < 0 ||
+          static_cast<uint32_t>(index) >=
+              resultType.cast<spirv::CompositeType>().getNumMembers()) {
+        return compExOp.emitOpError("index ")
+               << index << " out of bounds for " << resultType;
+      }
+      resultType = resultType.cast<spirv::CompositeType>().getMemberType(index);
+    } else {
+      return compExOp.emitOpError("invalid type to extract from");
+    }
+  }
+
+  if (resultType != compExOp.getType()) {
+    return compExOp.emitOpError("invalid result type, expected ")
+           << resultType << " provided " << compExOp.getType();
+  }
+
+  return success();
 }
 
 //===----------------------------------------------------------------------===//
--- a/lib/SPIRV/SPIRVTypes.cpp
+++ b/lib/SPIRV/SPIRVTypes.cpp
@@ -19,6 +19,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/IR/StandardTypes.h"
 #include "mlir/SPIRV/SPIRVTypes.h"
 #include "llvm/ADT/StringSwitch.h"
 
@@ -27,6 +28,36 @@ using namespace mlir::spirv;
 
 // Pull in all enum utility function definitions
 #include "mlir/SPIRV/SPIRVEnums.cpp.inc"
+
+//===----------------------------------------------------------------------===//
+// CompositeType
+//===----------------------------------------------------------------------===//
+
+Type CompositeType::getMemberType(uint64_t index) const {
+  switch (getKind()) {
+  case spirv::TypeKind::Struct:
+    return cast<StructType>().getMemberType(index);
+  case spirv::TypeKind::Array:
+    return cast<ArrayType>().getElementType();
+  case StandardTypes::Vector:
+    return cast<VectorType>().getElementType();
+  default:
+    llvm_unreachable("invalid composite type");
+  }
+}
+
+uint64_t CompositeType::getNumMembers() const {
+  switch (getKind()) {
+  case spirv::TypeKind::Struct:
+    return cast<StructType>().getNumMembers();
+  case spirv::TypeKind::Array:
+    return cast<ArrayType>().getElementCount();
+  case StandardTypes::Vector:
+    return cast<VectorType>().getNumElements();
+  default:
+    llvm_unreachable("invalid composite type");
+  }
+}
 
 //===----------------------------------------------------------------------===//
 // ArrayType
--- a/test/SPIRV/ops.mlir
+++ b/test/SPIRV/ops.mlir
@@ -1,6 +1,142 @@
 // RUN: mlir-opt -split-input-file -verify-diagnostics %s | FileCheck %s
 
 //===----------------------------------------------------------------------===//
+// spv.CompositeExtractOp
+//===----------------------------------------------------------------------===//
+
+func @composite_extract_f32_from_1D_array(%arg0: !spv.array<4xf32>) -> f32 {
+  // CHECK: %0 = spv.CompositeExtract %arg0[1 : i32] : !spv.array<4 x f32>
+  %0 = spv.CompositeExtract %arg0[1 : i32] : !spv.array<4xf32>
+  return %0: f32
+}
+
+// -----
+
+func @composite_extract_f32_from_2D_array(%arg0: !spv.array<4x!spv.array<4xf32>>) -> f32 {
+  // CHECK: %0 = spv.CompositeExtract %arg0[1 : i32, 2 : i32] : !spv.array<4 x !spv.array<4 x f32>>
+  %0 = spv.CompositeExtract %arg0[1 : i32, 2 : i32] : !spv.array<4x!spv.array<4xf32>>
+  return %0: f32
+}
+
+// -----
+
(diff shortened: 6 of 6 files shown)


lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
...
+                  OpAsmPrinter *printer) {
+  *printer << spirv::CompositeExtractOp::getOperationName() << ' '
+           << *compositeExtractOp.composite() << compositeExtractOp.indices()
+           << " : " << compositeExtractOp.composite()->getType();
+}
+
+static LogicalResult verify(spirv::CompositeExtractOp compositeExtractOp) {
+  auto resultType = compositeExtractOp.getType();
+  if (resultType == compositeExtractOp.composite()->getType()) {
+    return compositeExtractOp.emitOpError(
```
antiagainst: Could you still call the function to verify the result type agree with the extracted type? It's a bit unfortunate that we are duplicating the check logic but we need that in the parser to properly set the result type. And this verifier is used across all paths (parsing custom assembly form, parsing generic assembly form, and building ops with C++ API, etc.) so it would be nice to guarantee.

test/SPIRV/ops.mlir
```diff
@@ -496,3 +496,139 @@ func @cannot_be_generic_storage_class(%arg0: f32) -> () {
   %0 = spv.Variable : !spv.ptr<f32, Generic>
   return
 }
+
+// -----
```
antiagainst: Awesome tests! :D

test/SPIRV/ops.mlir
```diff
@@ -496,3 +496,139 @@ func @cannot_be_generic_storage_class(%arg0: f32) -> () {
...
+ // expected-error @+1 {{invalid type for indexing}}
+  %0 = spv.CompositeExtract %arg0[1 : i32] : f32
+  return
+}
+
+// -----
+
+func @composite_extract_invalid_extracted_type(%arg0: !spv.array<4x!spv.array<4xf32>>) -> () {
+  // expected-error @+1 {{extracted element type can not be the same with composite type}}
+  %0 = spv.CompositeExtract %arg0[] : !spv.array<4x!spv.array<4xf32>>
```
antiagainst: Ah, it's interesting. I think we need a different check for this. OpCompositeExtract should not allow empty indices. (Although it's not clearly written on the spec.)

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
...
+// Extracts the Result Type from Composite Type iterating all over the indices,
+// with check for out of bounds.
+static Type getElementTypeByIndices(OpAsmParser *parser, OperationState *state,
+                                    Type compositeType,
+                                    ArrayAttr indicesArrayAttr,
+                                    llvm::SMLoc attrLocation) {
+  Type resultType = compositeType;
+  for (Attribute indexAttr : indicesArrayAttr) {
+    if (!resultType) {
+      parser->emitError(attrLocation, "invalid type to extract from");
```
antiagainst: What about `..., "cannnot index into ") << resultType;`?

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
...
+      return nullptr;
+    }
+    // The type cast is save, because of the description in TableGen.
```
antiagainst: This is not true here... The TableGen'd checks are inside `verify()` but this one is used for parsing.

denis0x0D:
@antiagainst thanks a lot for the detailed review! I'll update it, regarding to your comments.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// spv.CompositeExtractOp
```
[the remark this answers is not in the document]
denis0x0D: Oh, I see, thanks, fixed.

include/mlir/SPIRV/SPIRVOps.td
```diff
@@ -324,4 +324,47 @@ def SPV_VariableOp : SPV_Op<"Variable"> {
...
+    ```
+
+  }];
+
+  let arguments = (ins
+    SPV_Composite:$composite,
+    I32ArrayAttr:$indices
+  );
+
+  let results = (outs SPV_Type);
```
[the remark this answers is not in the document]
denis0x0D: Fixed.

include/mlir/SPIRV/SPIRVOps.td
```diff
@@ -324,4 +324,47 @@ def SPV_VariableOp : SPV_Op<"Variable"> {
   let opcode = 59;
 }
 
+def SPV_CompositeExtractOp : SPV_Op<"CompositeExtract", [NoSideEffect]> {
+  let summary = "Extract a part of a composite object.";
+
+  let description = [{
+    Extract a part of a composite object.
```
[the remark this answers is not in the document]
denis0x0D: Fixed.

include/mlir/SPIRV/SPIRVBase.td
```diff
@@ -79,14 +80,15 @@ def SPV_Vector : VectorOf<[SPV_Bool, SPV_Integer, SPV_Float]>;
 def SPV_AnyPtr : Type<SPV_IsPtrType, "any SPIR-V pointer type">;
 def SPV_AnyArray : Type<SPV_IsArrayType, "any SPIR-V array type">;
 def SPV_AnyRTArray : Type<SPV_IsRTArrayType, "any SPIR-V runtime array type">;
+def SPV_AnyStruct : Type<SPV_IsStructType, "any SPIR-V struct type">;
 
 def SPV_Numerical : AnyTypeOf<[SPV_Integer, SPV_Float]>;
 def SPV_Scalar : AnyTypeOf<[SPV_Numerical, SPV_Bool]>;
 def SPV_Aggregrate : AnyTypeOf<[SPV_AnyArray]>;
```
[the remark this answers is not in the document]
denis0x0D: Thanks, added.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
...
+                                           OperationState *state) {
+  OpAsmParser::OperandType compositeInfo;
+  Attribute indicesAttr;
+  Type compositeType;
+  llvm::SMLoc attrLocation;
+
+  if (parser->parseOperand(compositeInfo) ||
+      parser->getCurrentLocation(&attrLocation) ||
+      parser->parseAttribute(indicesAttr, "indices", state->attributes) ||
+      parser->parseColon() || parser->parseType(compositeType) ||
```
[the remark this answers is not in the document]
denis0x0D: Thanks, fixed.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
...
+      parser->resolveOperand(compositeInfo, compositeType, state->operands)) {
+    return failure();
+  }
+
+  auto indicesArrayAttr = indicesAttr.dyn_cast<ArrayAttr>();
+  if (!indicesArrayAttr) {
+    parser->emitError(attrLocation, "expected an array attribute");
+    return failure();
```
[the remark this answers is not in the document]
denis0x0D: Fixed.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
...
+      parser->getCurrentLocation(&attrLocation) ||
+      parser->parseAttribute(indicesAttr, "indices", state->attributes) ||
+      parser->parseColon() || parser->parseType(compositeType) ||
+      parser->resolveOperand(compositeInfo, compositeType, state->operands)) {
+    return failure();
+  }
+
+  auto indicesArrayAttr = indicesAttr.dyn_cast<ArrayAttr>();
+  if (!indicesArrayAttr) {
+    parser->emitError(attrLocation, "expected an array attribute");
```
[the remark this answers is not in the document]
denis0x0D: Fixed.

test/SPIRV/ops.mlir
```diff
@@ -496,3 +496,139 @@ func @cannot_be_generic_storage_class(%arg0: f32) -> () {
   %0 = spv.Variable : !spv.ptr<f32, Generic>
   return
 }
+
+// -----
+
+//===----------------------------------------------------------------------===//
+// spv.CompositeExtractOp
```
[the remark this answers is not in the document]
denis0x0D: Fixed.

test/SPIRV/ops.mlir
```diff
@@ -496,3 +496,139 @@ func @cannot_be_generic_storage_class(%arg0: f32) -> () {
...
+//===----------------------------------------------------------------------===//
+
+func @composite_extract_f32_from_1D_array(%arg0: !spv.array<4xf32>) -> () {
```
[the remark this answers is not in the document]
denis0x0D: Fixed.

test/SPIRV/ops.mlir
```diff
@@ -496,3 +496,139 @@ func @cannot_be_generic_storage_class(%arg0: f32) -> () {
...
+  %2 = spv.Load "Function" %1 ["Volatile"] : !spv.array<4x!spv.array<4xf32>>
+  // expected-error @+1 {{expected non-function type}}
+  %3 = spv.CompositeExtract %2[%0] : !spv.array<4x!spv.array<4xf32>>
+  return
+}
+
+// -----
+
+func @composite_extract_invalid_index_type_2(%arg0 : !spv.array<4x!spv.array<4xf32>>) -> () {
+  // expected-error @+1 {{failed to satisfy constraint}}
```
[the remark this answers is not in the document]
denis0x0D: Fixed.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
...
+      parser->emitError(attrLocation, "invalid type to extract from");
+      return nullptr;
+    }
+    // The type cast is save, because of the description in TableGen.
+    int32_t index = indexAttr.dyn_cast<IntegerAttr>().getInt();
+    if (resultType.getKind() == spirv::TypeKind::Array) {
+      if (index < 0 ||
+          static_cast<uint32_t>(index) >=
+              resultType.cast<spirv::ArrayType>().getElementCount()) {
+        parser->emitError(attrLocation, "index out of bounds");
```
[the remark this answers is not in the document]
denis0x0D: Fixed, in verifier.

lib/SPIRV/SPIRVOps.cpp
denis0x0D: Added a verifier.

test/SPIRV/ops.mlir
denis0x0D: Thanks, fixed.

lib/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, added in some cases, but in this case seems like resultType == nullptr.

lib/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, got it, updated.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -740,6 +740,102 @@ static LogicalResult verify(spirv::VariableOp varOp) {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// spv.CompositeExtractOp
+//===----------------------------------------------------------------------===//
+
+// Extracts the Result Type from Composite Type iterating all over the indices,
+// with check for out of bounds.
+static Type getElementTypeByIndices(OpAsmParser *parser, OperationState *state,
```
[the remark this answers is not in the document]
denis0x0D: I've added CompositeType as a base class between ArrayType, StructType. The CompositeTypeStorage does not have a static "construct" method, because we don't need to directly construct it and therefore it does not have a operator == for the hash function.
It provides a method getElementTypByIndices which consumes array of int32_t, as result it returns a pair<Type, StringRef> on success it would be a type selected by the last provided index, on failure a nullpr with an error message. I didn't come with better idea, how to handle different error messages and don't insert  asserts.

denis0x0D:
@antiagainst I've updated the patch, regarding your comments, can you please take a look. 
Thanks.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -197,6 +198,144 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+
+  auto indicesArrayAttr = indicesAttr.dyn_cast<ArrayAttr>();
+  if (!indicesArrayAttr) {
+    return compExOp.emitOpError(
+        "expected an 32-bit integer array attribute for 'indices'");
+  }
+
+  if (!indicesArrayAttr.size()) {
+    return compExOp.emitOpError("expexted at least one index for ")
+           << resultType;
```
antiagainst: You mean `spv.CompositeExtract` here?

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -197,6 +198,144 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+    if (auto indexIntAttr = indexAttr.dyn_cast<IntegerAttr>()) {
+      indexes.push_back(indexIntAttr.getInt());
+    } else {
+      return parser->emitError(attrLocation,
+                               "expexted an 32-bit integer for index");
+    }
+  }
+
+  std::pair<Type, StringRef> resultPair;
+  switch (compositeType.getKind()) {
```
antiagainst: This should be simplified after letting CompositeType to include VectorType.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -197,6 +198,144 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+           << *compositeExtractOp.composite() << compositeExtractOp.indices()
+           << " : " << compositeExtractOp.composite()->getType();
+}
+
+static LogicalResult verify(spirv::CompositeExtractOp compExOp) {
+  auto resultType = compExOp.composite()->getType();
+  auto indicesAttr = compExOp.indices();
+
+  auto indicesArrayAttr = indicesAttr.dyn_cast<ArrayAttr>();
+  if (!indicesArrayAttr) {
```
antiagainst: This check is already enforced by `I32ArrayAttr`.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -197,6 +198,144 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+           << resultType;
+  }
+
+  int32_t index;
+  for (auto indexAttr : indicesArrayAttr) {
+    if (!resultType) {
+      return compExOp.emitError("invalid type to extract from");
+    }
+
+    if (auto indexIntAttr = indexAttr.dyn_cast<IntegerAttr>()) {
```
antiagainst: This check is already enforced by `I32ArrayAttr`

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -197,6 +198,144 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+      index = indexIntAttr.getInt();
+    } else {
+      return compExOp.emitOpError("expexted an 32-bit integer for index");
+    }
+
+    switch (resultType.getKind()) {
```
antiagainst: Why not using CompositeType's getMemberType() here?

include/mlir/SPIRV/SPIRVTypes.h
```diff
@@ -52,9 +53,28 @@ enum Kind {
 };
 }
 
+// Provides a common base class between SPIR-V ArrayType, SPIR-V StructType.
```
antiagainst: `... for VectorType, SPIR-V ArrayType, SPIR-V StructType`.

include/mlir/SPIRV/SPIRVTypes.h
```diff
@@ -52,9 +53,28 @@ enum Kind {
...
+class CompositeType : public Type {
+public:
+  using ImplType = detail::CompositeTypeStorage;
```
antiagainst: I think we can use `DefaultTypeStorage` here to avoid defining `CompositeTypeStorage` given that we are not storing anything inside this base class actually?

include/mlir/SPIRV/SPIRVTypes.h
```diff
@@ -52,9 +53,28 @@ enum Kind {
...
+  using Type::Type;
+
+  static bool kindof(unsigned kind) {
+    return (kind == TypeKind::Array || kind == TypeKind::Struct);
+  }
+
+  // Consumes an array of 32-bit indices.
+  // Returns a pair with the type selected by the last provided index on
+  // success, in other case returns a nullptr with an error message.
+  std::pair<Type, StringRef>
```
antiagainst: I see we have `getMemberType` in this class too. Then it seems to me we don't need this method anymore. This way we don't need to handle the error cases inside the type library (and force the same error message to all callers). At call sites we can do something like

```cpp
for (int64_t index : indices) {
  if (auto cType = type.dyn_cast<CompositeType>()) {
    Type elemType = cType.getMemberType(index);
    if (!elemType) return ...emitError("cannot index into type ") << type << " with index " << index;
    type = elemType;
  } else {
    return ...emitError("cannot index into non-composite type ") << type;
  }
}
```
?

lib/SPIRV/SPIRVTypes.cpp
```diff
@@ -28,11 +29,77 @@ using namespace mlir::spirv;
 // Pull in all enum utility function definitions
 #include "mlir/SPIRV/SPIRVEnums.cpp.inc"
 
+//===----------------------------------------------------------------------===//
+// CompositeType
+//===----------------------------------------------------------------------===//
+
+struct spirv::detail::CompositeTypeStorage : public TypeStorage {
```
antiagainst: I think we can remove this and use DefaultTypeStorage.

lib/SPIRV/SPIRVTypes.cpp
```diff
@@ -28,11 +29,77 @@ using namespace mlir::spirv;
...
+  CompositeTypeStorage(unsigned numMembers = 0) : TypeStorage(numMembers) {}
+};
+
+Type CompositeType::getMemberType(uint64_t index) const {
+  switch (getKind()) {
+  case spirv::TypeKind::Struct:
+    return cast<StructType>().getMemberType(index);
+  case spirv::TypeKind::Array:
+    return cast<ArrayType>().getElementType();
+  default:
```
antiagainst: Missing VectorType

lib/SPIRV/SPIRVTypes.cpp
```diff
@@ -28,11 +29,77 @@ using namespace mlir::spirv;
...
+    assert(0 && "invalid composite type");
```
antiagainst: Please use llvm_unreachable here.

lib/SPIRV/SPIRVTypes.cpp
```diff
@@ -28,11 +29,77 @@ using namespace mlir::spirv;
...
+  return {resultType, "sucess"};
+}
+
+uint64_t CompositeType::getNumMembers() const {
+  switch (getKind()) {
+  case spirv::TypeKind::Struct:
+    return cast<StructType>().getNumMembers();
+  case spirv::TypeKind::Array:
+    return cast<ArrayType>().getElementCount();
+  default:
```
antiagainst: Missing VectorType

lib/SPIRV/SPIRVTypes.cpp
```diff
@@ -28,11 +29,77 @@ using namespace mlir::spirv;
...
+    assert(0 && "invalid composite type");
```
antiagainst: Also please use llvm_unreachable here

test/SPIRV/ops.mlir
```diff
@@ -496,3 +632,5 @@ func @cannot_be_generic_storage_class(%arg0: f32) -> () {
   %0 = spv.Variable : !spv.ptr<f32, Generic>
   return
 }
+
```
antiagainst: Redundant empty lines :)

include/mlir/SPIRV/SPIRVTypes.h
```diff
@@ -52,9 +53,28 @@ enum Kind {
...
 }
 
+// Provides a common base class between SPIR-V ArrayType, SPIR-V StructType.
+class CompositeType : public Type {
+public:
+  using ImplType = detail::CompositeTypeStorage;
+  using Type::Type;
+
+  static bool kindof(unsigned kind) {
+    return (kind == TypeKind::Array || kind == TypeKind::Struct);
```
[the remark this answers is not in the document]
denis0x0D: Oh, I see, to be able to use kinda "cast" thing we have to provide "kindof" function. I was stuck for I a while trying to understand how to use Vector from Standard Types as kinda CompositeType, in case we can't directly modify it. Thanks a lot, now I'm understanding MLIR's type system more better. Sorry, for this kind of realization.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -197,6 +198,144 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+           << compositeType;
+  }
+
+  SmallVector<int32_t, 4> indexes;
+  for (auto indexAttr : indicesArrayAttr) {
+    if (auto indexIntAttr = indexAttr.dyn_cast<IntegerAttr>()) {
+      indexes.push_back(indexIntAttr.getInt());
+    } else {
+      return parser->emitError(attrLocation,
+                               "expexted an 32-bit integer for index");
```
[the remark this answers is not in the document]
denis0x0D: Fixed.

lib/SPIRV/SPIRVOps.cpp
```diff
@@ -197,6 +198,144 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+  auto indicesArrayAttr = indicesAttr.dyn_cast<ArrayAttr>();
+  if (!indicesArrayAttr) {
+    return parser->emitError(
+        attrLocation,
+        "expected an 32-bit integer array attribute for 'indices'");
+  }
+
+  if (!indicesArrayAttr.size()) {
+    return parser->emitError(attrLocation, "expexted at least one index for ")
+           << compositeType;
```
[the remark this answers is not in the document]
denis0x0D: Thanks, updated.

lib/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, fixed.

lib/SPIRV/SPIRVOps.cpp
denis0x0D: Updated.

lib/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, my fault.

lib/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, fixed.

lib/SPIRV/SPIRVOps.cpp
denis0x0D: Updated

include/mlir/SPIRV/SPIRVTypes.h
denis0x0D: Updated.

include/mlir/SPIRV/SPIRVTypes.h
denis0x0D: Fixed.

include/mlir/SPIRV/SPIRVTypes.h
denis0x0D: Yea, seems more cleaner. Updated.

lib/SPIRV/SPIRVTypes.cpp
denis0x0D: Removed

lib/SPIRV/SPIRVTypes.cpp
denis0x0D: Updated.

lib/SPIRV/SPIRVTypes.cpp
denis0x0D: Thanks, fixed.

lib/SPIRV/SPIRVTypes.cpp
denis0x0D: Thanks, fixed.

lib/SPIRV/SPIRVTypes.cpp
denis0x0D: Thanks, updated.

test/SPIRV/ops.mlir
denis0x0D: Thanks, fixed.

denis0x0D:
@antiagainst thanks for the review! I've updated the patch, can you please take a look again :) Thanks!

include/mlir/SPIRV/SPIRVTypes.h
```diff
@@ -52,9 +53,27 @@ enum Kind {
 };
 }
 
+// Provides a common base class between VectorType, SPIR-V ArrayType, SPIR-V
+// StructType.
+class CompositeType : public Type {
+public:
+  using ImplType = DefaultTypeStorage;
```
River707: You don't need to specify this, ImplType already defaults to DefaultTypeStorage.

denis0x0D: @River707 Oh, my fault, thanks! Trying to get more familiar with MLIR.

denis0x0D:
@antiagainst @River707 thanks for review! Updated regarding to the latest comment.

antiagainst:
This is now landed as b31675b. Thanks a lot for your contribution, @denis0x0D!

denis0x0D:
@antiagainst thanks a lot! It would be nice, if you have some other tasks, not assigned to someone, which I can help with! Thanks.

antiagainst:
@denis0x0D, that's awesome! We certainly have lots to do and I really appreciate your help here! :) Let me create some other issues and assign to you. :)

denis0x0D:
@antiagainst thanks a lot! BTW, I will be in travel in [15, 21) of July, so, if it possible, some issue which I can start from 21-th of July.

antiagainst:
@denis0x0D, sure! Thanks for letting me know! I wonder whether you can help to add `spv.AccessChain` too. Given that you've done `spv.CompositeExtract`, it should be quite similar. I've just landed https://github.com/tensorflow/mlir/commit/9a8202eea72efd3992e59216f5ffd545913ba943, so you can use it to inject the basic definition into `SPIRVOps.td` with `./define_inst.sh OpAccessChain` and update according to the spec. But certainly no pressure! Just let me know. Thanks again! :)

denis0x0D:
@antiagainst thanks a lot for your time! It would be nice to implement OpAccessChain op, if it can wait until the next week, I'll take it wih a pleasure.
I see that helpful script gen_spirv_dialect.py, thanks!

antiagainst:
No worries! Let's see whether we can get it done this week. If not, you are free to take it next week! Otherwise we can certainly use your help somewhere else. :)

denis0x0D:
@antiagainst is it ok if I'll implement OpAccessChain? If noone works on it. Thanks.

antiagainst:
@denis0x0D, yup, feel free to take it! Thanks a lot! :)

denis0x0D:
@antiagainst thanks! Can you please answer some quick question about implementation, I want to make sure, that I understand it in right way.
In case of AccessChainOp it creates a pointer into composite type from pointer to composite type, the result pointer type must be provided by walking the Base's type hierarchy down to last provided index; the indexes are in SSA forms, so we can't walk down to last provided index during compile time, or I'm missing something. In parser we need to resolve the SSA values provided by Base (ssa-id) and get the type of result. to add it into list So, the ebnf form must look like this:

```
 op ::= ssa-id `=` `spv.AccessChain` ssa-use
           `[` ssa-use (`,` ssa-use)* `]`
           `:` pointer-type `,` pointer-type

```
for example 

`%result_ptr = spv.AccessChain %composite_ptr[%index0] : !spv.ptr<!spv.struct<f32, spv.array<4xf32>>, Function>, !spv.ptr<!spv.array<4xf32>, Function>
`

The first pointer-type is a type of composite object, the second one is a result type, in this case we can easily resolve the base type and result type.
But I might miss something, please correct me if I'm wrong.

antiagainst:
> so we can't walk down to last provided index during compile time, or I'm missing something.

We can get the accessed component's type statically; it's just that we are not guaranteed to be in bounds dynamically. An vector/array/runtime-array type contains elements of the same type, so we will just get its element type. For structure type which contains heterogeneous elements, the spec requires the index to be from an `OpConstant` so we can statically deduce what type the accessed component is of.

For the trailing types in the custom assembly form, I think only output the composite's type should suffice?

denis0x0D:
@antiagainst oh, I see, we can get an Operation associated with SSA Value (getDefiningOp) and thereofore all attributes (actual value of ConstantOp) to deduce the type extracted from the Composite object, sorry for the stupid question. 
Thanks.

PR #49, opened 2019-07-19
Add TTI pass initialization to pass managers.

dcaballe:
Hi!

This PR is mostly to ask for feedback :). We are using `mlir::ExecutionEngine` to JIT-compile and execute nGraph ops on CPUs and noticed that LLVM optimizer wasn't getting the proper target information because TTI implementation was initialized with `NoTTIImpl`. This happens because `TargetTransformInfoWrapperPass` is not added to the pass managers.

I think there are a couple of ways to fix this:
1) MLIR repo: something like the code attached, which adds `TargetTransformInfoWrapperPass` initialized with the proper `TargetMachine` to the pass managers. I tested it by passing a `TargetMachine` with all the default host sub-target features and specific cpu name (skylake-avx512). However, I must be missing something since I’m getting skylake-avx512 assembly with this approach even though the target machine used by OrcJit doesn’t have any sub-target features/cpu name set: https://github.com/tensorflow/mlir/blob/master/lib/ExecutionEngine/ExecutionEngine.cpp#L150
2) nGraph repo: using `mlir::makeLLVMPassesTransformer` (https://github.com/tensorflow/mlir/blob/master/lib/ExecutionEngine/OptUtils.cpp#L103) in nGraph to add `TargetTransformInfoWrapperPass` to the pass managers.
3) Any other idea?

I see value in adding `TargetTransformInfoWrapperPass` pass in MLIR repo so that other users don’t hit the same issue or have to replicate the same code. I also see value in having a utility in MLIR repo that creates a `TargetMachine` with the default host cpu parameters (cpu name, sub-target features, etc.) and reusing it every time is needed (for example, in the OrcJIT code I pointed before and also here: https://github.com/tensorflow/mlir/blob/master/lib/ExecutionEngine/ExecutionEngine.cpp#L150)
 
Please, let me know what you think. I should be able to help with this.

Thanks!
Diego


include/mlir/ExecutionEngine/OptUtils.h
```diff
@@ -41,9 +42,12 @@ void initializeLLVMPasses();
...
 /// Create a module transformer function for MLIR ExecutionEngine that runs
 /// LLVM IR passes corresponding to the given speed and size optimization
-/// levels (e.g. -O2 or -Os).
+/// levels (e.g. -O2 or -Os). If not null, \p targetMachine is used to
+/// initialize passes that provide target-specific information to the LLVM
+/// optimizer.
 std::function<llvm::Error(llvm::Module *)>
-makeOptimizingTransformer(unsigned optLevel, unsigned sizeLevel);
+makeOptimizingTransformer(unsigned optLevel, unsigned sizeLevel,
+                          llvm::TargetMachine *targetMachine = nullptr);
```
joker-eph: I would not even have a default nullptr value here to avoid forgetting it at call-sites.

joker-eph:
Thanks! I had a similar patch locally that I didn't complete.

Ultimately my take was that most of the code here could be moved directly in LLVM as there isn't much specific to MLIR in the JIT abstraction we have in MLIR. In the meantime this PR seems like a good thing to do.

dcaballe:
> Ultimately my take was that most of the code here could be moved directly in LLVM as there isn't much specific to MLIR in the JIT abstraction we have in MLIR.

Yeah, that would be great.

> I also see value in having a utility in MLIR repo that creates a TargetMachine with the default host cpu parameters (cpu name, sub-target features, etc.)

Do you think adding a utility for this ^ in a separate commit is also a good idea?

This is a suggestion for the squashed commit message. Sorry, I didn't add it to the first commit:
Add TargetMachine parameter to makeOptimizingTransformer and populatePassManagers APIs and initialize `TargetTransformInfoWrapperPass` with the right target information and add it to the pass managers.

Thanks, Mehdi!
Diego

dcaballe:
> > I also see value in having a utility in MLIR repo that creates a TargetMachine with the default host cpu parameters (cpu name, sub-target features, etc.)
> 
> Do you think adding a utility for this ^ in a separate commit is also a good idea?

https://github.com/dcaballe/mlir/pull/1 is a stepping stone into ^. Please, let me know what you think.

Thanks!
Diego

joker-eph:
You have conflicts now apparently, can you rebase?

> dcaballe#1 is a stepping stone into ^. Please, let me know what you think.

Thanks, it seems a bit strange to me to manually unwrap a TM to build a TM-builder, I'm not sure why ORCJIT does not just take a `std::function<std::unique_ptr<TargetMachine>()>` callback instead of this class?
Most of the code you're adding is putting this whole utility even more in LLVM land, at this point I rather invest time into moving this into LLVM itself than adding so much "low-level" glue code in MLIR.

dcaballe:
> You have conflicts now apparently, can you rebase?

Done

> I'm not sure why ORCJIT does not just take a std::function<std::unique_ptr<TargetMachine>()> callback instead of this class?

Yeah, it would make more sense since TM seems to be needed before ORCJIT is created. There is sort of a chicken-egg problem as we need TM to properly initialize TTI pass in the `transformer` function that we pass to create ORCJIT. 

> at this point I rather invest time into moving this into LLVM itself than adding so much "low-level" glue code in MLIR.

Sure. Please, keep me posted and let me know if I can help.

Diego

dcaballe:
Hi Mehdi,

Do you think we could proceed with this PR in the meantime until things are moved to LLVM? That would be very helpful.

Thanks,
Diego

lib/ExecutionEngine/OptUtils.cpp
```diff
@@ -69,7 +71,8 @@ void mlir::initializeLLVMPasses() {
 // This behaves similarly to LLVM opt.
 static void populatePassManagers(llvm::legacy::PassManager &modulePM,
                                  llvm::legacy::FunctionPassManager &funcPM,
-                                 unsigned optLevel, unsigned sizeLevel) {
+                                 unsigned optLevel, unsigned sizeLevel,
+                                 llvm::TargetMachine *targetMachine = nullptr) {
```
joker-eph: There is a single call site, you should be able to skip the default value here

include/mlir/ExecutionEngine/OptUtils.h
```diff
@@ -41,9 +42,12 @@ void initializeLLVMPasses();
 
 /// Create a module transformer function for MLIR ExecutionEngine that runs
 /// LLVM IR passes corresponding to the given speed and size optimization
-/// levels (e.g. -O2 or -Os).
+/// levels (e.g. -O2 or -Os). If not null, \p targetMachine is used to
+/// initialize passes that provide target-specific information to the LLVM
+/// optimizer.
```
joker-eph: Can you mention that the the provided TM is expected to outlive the returned std::function?

joker-eph:
Sorry, I didn't mean to stale the current PR of course. If you don't mind updating it with the two nit I just commented, that seems OK to merge.

lib/ExecutionEngine/OptUtils.cpp
```diff
@@ -127,7 +127,8 @@ std::function<llvm::Error(llvm::Module *)> mlir::makeLLVMPassesTransformer(
         continue;
 
       if (insertOptPasses && optPassesInsertPos == i) {
-        populatePassManagers(modulePM, funcPM, mbOptLevel.getValue(), 0);
+        populatePassManagers(modulePM, funcPM, mbOptLevel.getValue(), 0,
+                             nullptr /*TTI*/);
```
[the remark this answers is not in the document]
dcaballe: Thanks, it makes sense. I didn't know that about clang-tidy. Good to know!

lib/Support/JitRunner.cpp
```diff
@@ -308,8 +308,8 @@ int mlir::JitRunnerMain(
     if (failed(mlirTransformer(m.get())))
       return EXIT_FAILURE;
 
-  auto transformer =
-      mlir::makeLLVMPassesTransformer(passes, optLevel, optPosition);
+  auto transformer = mlir::makeLLVMPassesTransformer(
+      passes, optLevel, /*targetMachine=*/nullptr, optPosition);
```
joker-eph: Here can (and should?) we do something like:

```
auto TM_or_error = JITTargetMachineBuilder::detectHost().createTargetMachine();
if (!TM) return EXIT_FAILURE;
auto transformer = mlir::makeLLVMPassesTransformer(
		       passes, optLevel, /*targetMachine=*/TM_or_error->get(), optPosition);
```

dcaballe: No strong opinion here but maybe we should add this in a separate commit since this is changing runner's behavior. Also, I'm not sure how much we are getting from `detectHost().createTargetMachine()` since this is not adding sub-target features. We would need something more in the direction of what I pointed out (https://github.com/dcaballe/mlir/pull/1) but I understood you preferred that code in LLVM. Let me know if we can reuse part of that code. I could prepare another commit with that. WDYT?

joker-eph: Isn't most of the TLI independent of sub-target features?
In any case adding sub-target features seems like the kind of thing that `detectHost().createTargetMachine()` should take care of, don't you think?

include/mlir/ExecutionEngine/OptUtils.h
```diff
@@ -41,17 +42,23 @@ void initializeLLVMPasses();
 
 /// Create a module transformer function for MLIR ExecutionEngine that runs
 /// LLVM IR passes corresponding to the given speed and size optimization
-/// levels (e.g. -O2 or -Os).
+/// levels (e.g. -O2 or -Os). If not null, `targetMachine` is used to
+/// initialize passes that provide target-specific information to the LLVM
+/// optimizer. `targetMachine` must outlive the returned std::function.
 std::function<llvm::Error(llvm::Module *)>
-makeOptimizingTransformer(unsigned optLevel, unsigned sizeLevel);
+makeOptimizingTransformer(unsigned optLevel, unsigned sizeLevel,
```
ftynse: Please update the call sites in examples/toy/Ch5/toyc.cpp and the likes

ftynse: see https://source.cloud.google.com/results/invocations/d9d8a397-5ac0-4333-82e2-a0706b00a1df/log

lib/Support/JitRunner.cpp
dcaballe: Vectorization is sub-target dependent, for example.

I agree with you but I feel there must be a reason why it's not there already :). Let me try to add it to LLVM and I'll create another PR to update jit runner if you don't beat me to it.

include/mlir/ExecutionEngine/OptUtils.h
dcaballe: Sorry, I had examples disabled. Done. Thanks!

ftynse: Thanks!

PR #50, opened 2019-07-22
Introduce explicit copying optimization by generalizing the DMA generation pass

bondhugula:
Explicit copying to contiguous buffers is a standard technique to avoid
conflict misses and TLB misses, and improve hardware prefetching
performance. When done in conjunction with cache tiling, it nearly
eliminates all cache conflict and TLB misses, and a single hardware
prefetch stream is needed per data tile.

- generalize/extend DMA generation pass (renamed data copying pass) to
  perform either point-wise explicit copies to fast memory buffers or
  DMAs (depending on a cmd line option). All logic is the same as
  erstwhile -dma-generate.

- -affine-dma-generate is now renamed -affine-data-copy; when -dma flag is
  provided, DMAs are generated, or else explicit copy loops are generated
  (point-wise) by default.

- point-wise copying could be used for CPUs (or GPUs); some indicative
  performance numbers with a "C" version of the MLIR when compiled with
  and without this optimization (about 2x improvement here).

  With a matmul on 4096^2 matrices on a single core of an Intel Core i7
  Skylake i7-8700K with clang 8.0.0:

  clang -O3:                       518s
  clang -O3 with MLIR tiling (128x128):      24.5s
  clang -O3 with MLIR tiling + data copying  12.4s
  (code equivalent to test/Transforms/data-copy.mlir func @matmul)

- fix some misleading comments.

- change default fast-mem space to 0 (more intuitive now with the
  default copy generation using point-wise copies instead of DMAs)

On a simple 3-d matmul loop nest, code generated with -affine-data-copy:

```
  affine.for %arg3 = 0 to 4096 step 128 {
    affine.for %arg4 = 0 to 4096 step 128 {
      %0 = affine.apply #map0(%arg3, %arg4)
      %1 = affine.apply #map1(%arg3, %arg4)
      %2 = alloc() : memref<128x128xf32, 2>
      // Copy-in Out matrix.
      affine.for %arg5 = 0 to 128 {
        %5 = affine.apply #map2(%arg3, %arg5)
        affine.for %arg6 = 0 to 128 {
          %6 = affine.apply #map2(%arg4, %arg6)
          %7 = load %arg2[%5, %6] : memref<4096x4096xf32>
          affine.store %7, %2[%arg5, %arg6] : memref<128x128xf32, 2>
        }
      }
      affine.for %arg5 = 0 to 4096 step 128 {
        %5 = affine.apply #map0(%arg3, %arg5)
        %6 = affine.apply #map1(%arg3, %arg5)
        %7 = alloc() : memref<128x128xf32, 2>
        // Copy-in LHS.
        affine.for %arg6 = 0 to 128 {
          %11 = affine.apply #map2(%arg3, %arg6)
          affine.for %arg7 = 0 to 128 {
            %12 = affine.apply #map2(%arg5, %arg7)
            %13 = load %arg0[%11, %12] : memref<4096x4096xf32>
            affine.store %13, %7[%arg6, %arg7] : memref<128x128xf32, 2>
          }
        }
        %8 = affine.apply #map0(%arg5, %arg4)
        %9 = affine.apply #map1(%arg5, %arg4)
        %10 = alloc() : memref<128x128xf32, 2>
        // Copy-in RHS.
        affine.for %arg6 = 0 to 128 {
          %11 = affine.apply #map2(%arg5, %arg6)
          affine.for %arg7 = 0 to 128 {
            %12 = affine.apply #map2(%arg4, %arg7)
            %13 = load %arg1[%11, %12] : memref<4096x4096xf32>
            affine.store %13, %10[%arg6, %arg7] : memref<128x128xf32, 2>
          }
        }
        // Compute.
        affine.for %arg6 = #map7(%arg3) to #map8(%arg3) {
          affine.for %arg7 = #map7(%arg4) to #map8(%arg4) {
            affine.for %arg8 = #map7(%arg5) to #map8(%arg5) {
              %11 = affine.load %7[-%arg3 + %arg6, -%arg5 + %arg8] : memref<128x128xf32, 2>
              %12 = affine.load %10[-%arg5 + %arg8, -%arg4 + %arg7] : memref<128x128xf32, 2>
              %13 = affine.load %2[-%arg3 + %arg6, -%arg4 + %arg7] : memref<128x128xf32, 2>
              %14 = mulf %11, %12 : f32
              %15 = addf %13, %14 : f32
              affine.store %15, %2[-%arg3 + %arg6, -%arg4 + %arg7] : memref<128x128xf32, 2>
            }
          }
        }
        dealloc %10 : memref<128x128xf32, 2>
        dealloc %7 : memref<128x128xf32, 2>
      }
      %3 = affine.apply #map0(%arg3, %arg4)
      %4 = affine.apply #map1(%arg3, %arg4)
      // Copy out result matrix.
      affine.for %arg5 = 0 to 128 {
        %5 = affine.apply #map2(%arg3, %arg5)
        affine.for %arg6 = 0 to 128 {
          %6 = affine.apply #map2(%arg4, %arg6)
          %7 = affine.load %2[%arg5, %arg6] : memref<128x128xf32, 2>
          store %7, %arg2[%5, %6] : memref<4096x4096xf32>
        }
      }
      dealloc %2 : memref<128x128xf32, 2>
    }
  }
```

With -affine-data-copy -dma:

```
  affine.for %arg3 = 0 to 4096 step 128 {
    %0 = affine.apply #map3(%arg3)
    %1 = alloc() : memref<128xf32, 2>
    %2 = alloc() : memref<1xi32>
    affine.dma_start %arg2[%arg3], %1[%c0], %2[%c0], %c128_0 : memref<4096xf32>, memref<128xf32, 2>, memref<1xi32>
    affine.dma_wait %2[%c0], %c128_0 : memref<1xi32>
    %3 = alloc() : memref<1xi32>
    affine.for %arg4 = 0 to 4096 step 128 {
      %5 = affine.apply #map0(%arg3, %arg4)
      %6 = affine.apply #map1(%arg3, %arg4)
      %7 = alloc() : memref<128x128xf32, 2>
      %8 = alloc() : memref<1xi32>
      affine.dma_start %arg0[%arg3, %arg4], %7[%c0, %c0], %8[%c0], %c16384, %c4096, %c128_2 : memref<4096x4096xf32>, memref<128x128xf32, 2>, memref<1xi32>
      affine.dma_wait %8[%c0], %c16384 : memref<1xi32>
      %9 = affine.apply #map3(%arg4)
      %10 = alloc() : memref<128xf32, 2>
      %11 = alloc() : memref<1xi32>
      affine.dma_start %arg1[%arg4], %10[%c0], %11[%c0], %c128_1 : memref<4096xf32>, memref<128xf32, 2>, memref<1xi32>
      affine.dma_wait %11[%c0], %c128_1 : memref<1xi32>
      affine.for %arg5 = #map3(%arg3) to #map5(%arg3) {
        affine.for %arg6 = #map3(%arg4) to #map5(%arg4) {
          %12 = affine.load %7[-%arg3 + %arg5, -%arg4 + %arg6] : memref<128x128xf32, 2>
          %13 = affine.load %10[-%arg4 + %arg6] : memref<128xf32, 2>
          %14 = affine.load %1[-%arg3 + %arg5] : memref<128xf32, 2>
          %15 = mulf %12, %13 : f32
          %16 = addf %14, %15 : f32
          affine.store %16, %1[-%arg3 + %arg5] : memref<128xf32, 2>
        }
      }
      dealloc %11 : memref<1xi32>
      dealloc %10 : memref<128xf32, 2>
      dealloc %8 : memref<1xi32>
      dealloc %7 : memref<128x128xf32, 2>
    }
    %4 = affine.apply #map3(%arg3)
    affine.dma_start %1[%c0], %arg2[%arg3], %3[%c0], %c128 : memref<128xf32, 2>, memref<4096xf32>, memref<1xi32>
    affine.dma_wait %3[%c0], %c128 : memref<1xi32>
    dealloc %3 : memref<1xi32>
    dealloc %2 : memref<1xi32>
    dealloc %1 : memref<128xf32, 2>
  }
```

### DIFF
--- a/lib/Transforms/AffineDataCopyGeneration.cpp
+++ b/lib/Transforms/AffineDataCopyGeneration.cpp
@@ -1,4 +1,4 @@
-//===- DmaGeneration.cpp - DMA generation pass ------------------------ -*-===//
+//===- AffineDataCopyGeneration.cpp - Explicit memref copying pass ------*-===//
 //
 // Copyright 2019 The MLIR Authors.
 //
@@ -17,7 +17,14 @@
 //
 // This file implements a pass to automatically promote accessed memref regions
 // to buffers in a faster memory space that is explicitly managed, with the
-// necessary data movement operations expressed as DMAs.
+// necessary data movement operations performed through either regular
+// point-wise load/store's or DMAs. Such explicit copying (also referred to as
+// array packing/unpacking in the literature), when done on arrays that exhibit
+// reuse, results in near elimination of conflict misses, TLB misses, reduced
+// use of hardware prefetch streams, and reduced false sharing. It is also
+// necessary for hardware that explicitly managed levels in the memory
+// hierarchy, and where DMAs may have to be used. This optimization is often
+// performed on already tiled code.
 //
 //===----------------------------------------------------------------------===//
 
@@ -34,7 +41,7 @@
 #include "llvm/Support/Debug.h"
 #include <algorithm>
 
-#define DEBUG_TYPE "affine-dma-generate"
+#define DEBUG_TYPE "affine-data-copy-generate"
 
 using namespace mlir;
 using llvm::SmallMapVector;
@@ -42,38 +49,45 @@ using llvm::SmallMapVector;
 static llvm::cl::OptionCategory clOptionsCategory(DEBUG_TYPE " options");
 
 static llvm::cl::opt<unsigned long long> clFastMemoryCapacity(
-    "dma-fast-mem-capacity",
+    "affine-data-copy-generate-fast-mem-capacity",
     llvm::cl::desc(
         "Set fast memory space capacity in KiB (default: unlimited)"),
     llvm::cl::cat(clOptionsCategory));
 
+static llvm::cl::opt<bool>
+    clDma("affine-data-copy-generate-dma",
+          llvm::cl::desc("Generate DMA instead of point-wise copy"),
+          llvm::cl::cat(clOptionsCategory));
+
 static llvm::cl::opt<unsigned> clFastMemorySpace(
-    "dma-fast-mem-space", llvm::cl::init(2),
+    "affine-data-copy-generate-fast-mem-space", llvm::cl::init(0),
     llvm::cl::desc(
-        "Fast memory space identifier for DMA generation (default: 1)"),
+        "Fast memory space identifier for copy generation (default: 1)"),
     llvm::cl::cat(clOptionsCategory));
 
 static llvm::cl::opt<bool> clSkipNonUnitStrideLoop(
-    "dma-skip-non-unit-stride-loops", llvm::cl::Hidden, llvm::cl::init(false),
+    "affine-data-copy-generate-skip-non-unit-stride-loops", llvm::cl::Hidden,
+    llvm::cl::init(false),
     llvm::cl::desc("Testing purposes: avoid non-unit stride loop choice depths "
-                   "for DMA placement"),
+                   "for copy placement"),
     llvm::cl::cat(clOptionsCategory));
 
 namespace {
 
 /// Replaces all loads and stores on memref's living in 'slowMemorySpace' by
-/// introducing DMA operations (strided DMA if necessary) to transfer data into
-/// `fastMemorySpace` and rewriting the original load's/store's to instead
-/// load/store from the allocated fast memory buffers. Additional options
-/// specify the identifier corresponding to the fast memory space and the amount
-/// of fast memory space available. The pass traverses through the nesting
-/// structure, recursing to inner levels if necessary to determine at what depth
-/// DMA transfers need to be placed so that the allocated buffers fit within the
-/// memory capacity provided.
-// TODO(bondhugula): We currently can't generate DMAs correctly when stores are
-// strided. Check for strided stores.
-struct DmaGeneration : public FunctionPass<DmaGeneration> {
-  explicit DmaGeneration(
+/// introducing copy operations to transfer data into `fastMemorySpace` and
+/// rewriting the original load's/store's to instead load/store from the
+/// allocated fast memory buffers. Additional options specify the identifier
+/// corresponding to the fast memory space and the amount of fast memory space
+/// available. The pass traverses through the nesting structure, recursing to
+/// inner levels if necessary to determine at what depth copies need to be
+/// placed so that the allocated buffers fit within the memory capacity
+/// provided.
+// TODO(bondhugula): We currently can't generate copies correctly when stores
+// are strided. Check for strided stores.
+struct AffineDataCopyGeneration
+    : public FunctionPass<AffineDataCopyGeneration> {
+  explicit AffineDataCopyGeneration(
       unsigned slowMemorySpace = 0,
       unsigned fastMemorySpace = clFastMemorySpace, unsigned tagMemorySpace = 0,
       int minDmaTransferSize = 1024,
@@ -82,7 +96,7 @@ struct DmaGeneration : public FunctionPass<DmaGeneration> {
         tagMemorySpace(tagMemorySpace), minDmaTransferSize(minDmaTransferSize),
         fastMemCapacityBytes(fastMemCapacityBytes) {}
 
-  explicit DmaGeneration(const DmaGeneration &other)
+  explicit AffineDataCopyGeneration(const AffineDataCopyGeneration &other)
       : slowMemorySpace(other.slowMemorySpace),
         fastMemorySpace(other.fastMemorySpace),
         tagMemorySpace(other.tagMemorySpace),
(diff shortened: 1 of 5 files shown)


lib/Transforms/DataCopy.cpp
```diff
@@ -34,46 +40,50 @@
...
 
 static llvm::cl::opt<unsigned long long> clFastMemoryCapacity(
-    "dma-fast-mem-capacity",
+    "copy-fast-mem-capacity",
     llvm::cl::desc(
         "Set fast memory space capacity in KiB (default: unlimited)"),
     llvm::cl::cat(clOptionsCategory));
 
+static llvm::cl::opt<bool>
+    clDma("dma", llvm::cl::desc("Generate DMA instead of point-wise copy"),
```
River707: All of these cl options should start with the pass arg, otherwise they don't make sense.
https://github.com/tensorflow/mlir/blob/master/g3doc/DeveloperGuide.md#pass-name-and-other-command-line-options

lib/Transforms/DataCopy.cpp
```diff
@@ -1,4 +1,4 @@
-//===- DmaGeneration.cpp - DMA generation pass ------------------------ -*-===//
+//===- DataCopy.cpp - Explicit data copying optimization pass --------- -*-===//
```
River707: Please add "AffineDma" to this file name so that it makes more sense contextually. 'DataCopy' is ambiguous and this also only deals with the Affine dialect.

lib/Transforms/DataCopy.cpp
```diff
@@ -93,24 +103,28 @@ struct DmaGeneration : public FunctionPass<DmaGeneration> {
   bool runOnBlock(Block *block);
   uint64_t runOnBlock(Block::iterator begin, Block::iterator end);
 
-  bool generateDma(const MemRefRegion &region, Block *block,
-                   Block::iterator begin, Block::iterator end,
-                   uint64_t *sizeInBytes, Block::iterator *nBegin,
-                   Block::iterator *nEnd);
+  bool generateCopy(const MemRefRegion &region, Block *block,
```
River707: Can this function be updated to return LogicalResult?

lib/Transforms/DataCopy.cpp
bondhugula: The pass is no longer about generating DMA alone, and generating DMAs is not the default. The copies are done via DMA only if an additional flag is provided; otherwise, it's just point-wise copies (via loads and stores) -- the way things move for example b/w GPU shared memory and its global memory.

Unfortunately, there isn't a good short name for this optimization. AffineExplicitCopy and AffineDataCopy are the best I can think of now.

River707: Ah I see, thanks for the explanation. AffineDataCopy seems fine then.

lib/Transforms/CMakeLists.txt
```diff
@@ -3,8 +3,8 @@ add_subdirectory(Utils)
 add_llvm_library(MLIRTransforms
   Canonicalizer.cpp
   CSE.cpp
+  DataCopy.cpp
```
andydavis1: DataCopy sounds a like a generic library utility function. How about DataCopyGeneration or DataTransferGeneration?

bondhugula: Yes, DataCopyGeneration or DataTransferGeneration are better names, albeit longer. Also, I missed mentioning that "array packing" is another name used in the literature for this optimization:
http://www.cs.utexas.edu/users/flame/pubs/TOMS-BLIS-Analytical.pdf

"MemrefPacking" (-affine-memref-pack) in this context is another possibility.

lib/Transforms/DataCopy.cpp
bondhugula: Done.

lib/Transforms/DataCopy.cpp
bondhugula: Done - changed to AffineDataCopyGeneration for now.

lib/Transforms/DataCopy.cpp
bondhugula: Done.

lib/Transforms/CMakeLists.txt
bondhugula: Changed to AffineDataCopyGeneration for now.

lib/Transforms/AffineDataCopyGeneration.cpp
```diff
@@ -392,50 +453,75 @@ bool DmaGeneration::generateDma(const MemRefRegion &region, Block *block,
...
-  fullyComposeAffineMapAndOperands(&tagAffineMap, &tagIndices);
-  if (!region.isWrite()) {
-    // DMA non-blocking read from original buffer to fast buffer.
-    b.create<AffineDmaStartOp>(loc, memref, memAffineMap, memIndices,
-                               fastMemRef, bufAffineMap, bufIndices, tagMemRef,
-                               tagAffineMap, tagIndices, numElementsSSA, stride,
-                               numEltPerStride);
+
+  if (!clDma) {
+    auto copyNest = generatePointWiseCopy(loc, memref, fastMemRef, memIndices,
```
andydavis1: Not in this PR, but this makes me wonder if in the future we could separate this logic out. For example, we could create a higher level AffineDataTransferOp(%src_memref, %dst_memref, %num_elements) which could be lowered at a later time to either DMAs or copy loop nests.

I'm wondering if this might help us sort out two concerns where are tangled together in this pass:
1) Computing the union of memref regions and loop nest depth at which to insert copies.
2) Transforming the IR to implement these copies.

Alternatively, we could split this PR into 3 things:
1) DataTransferAnalysis pass (which returns memref region unions and loop depths)
2) DmaGenerationPass
3) LoopNestCopyGenerationPass

bondhugula: Yes, before I started I implementing this, I was actually thinking about having such a transfer op that could be expanded out either to point-wise copies or DMAs - the op actually will carry all the information to do this expansion. But in terms of code reuse and separation, the current code is already pretty close to that (although all in the same file). Also, I didn't right now see the need for delaying that expansion, but if we do, adding such an op appears to be the way to go.

andydavis1:
Uday, we are seeing some broken builds from this PR, can you take a look:

 https://source.cloud.google.com/results/invocations/7b5b30bc-5586-49ce-8f79-5981b03686de/log

thanks

andydavis1:
verify the failure by building release build (cmake /tmpfs/src/git/llvm/llvm -DLLVM_BUILD_EXAMPLES=ON -DLLVM_ENABLE_CXX1Y=Y -DLLVM_TARGETS_TO_BUILD=host -DCMAKE_RULE_MESSAGES:BOOL=OFF -DCMAKE_VERBOSE_MAKEFILE:BOOL=OFF -DCMAKE_BUILD_TYPE=Release), the kokoro ones show gcc 5.3.

lib/Transforms/AffineDataCopyGeneration.cpp
```diff
@@ -220,31 +237,78 @@ emitRemarkForBlock(Block &block) {
...
+  SmallVector<Value *, 4> fastBufIndices, memIndices;
+  AffineForOp copyNestRoot;
+  for (unsigned d = 0, e = fastBufferShape.size(); d < e; ++d) {
+    auto forOp = b.create<AffineForOp>(loc, 0, fastBufferShape[d]);
+    if (d == 0)
+      copyNestRoot = forOp;
+    b = forOp.getBodyBuilder();
+    fastBufIndices.push_back(forOp.getInductionVar());
+    // Construct the subscript for the slow memref being copied.
+    ArrayRef<Value *> operands = {memIndicesStart[d], forOp.getInductionVar()};
```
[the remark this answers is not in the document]
bondhugula: Thanks very much!

> This essentially creates a reference to a temporary (initializer list on the right) and may lead to memory problems. Use an actual vector instead.

Thanks very much!

PR #52, opened 2019-07-24
[spirv] Add AccessChainOp operation.

denis0x0D:
This patch implements a AccessChainOp.
ebnf form:
```
access_chain_op ::= ssa-id `=` `spv.AccessChain` ssa-use
                        `[` ssa-use (',' ssa-use)* `]`
                        `:` pointer-type
```
Custom assembly exampes:

```
func @foo(%arg0 : index) -> () {
  %0 = spv.Variable : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
  %1 = spv.AccessChain %0[%arg0, %arg0] : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
  %2 = spv.Load "Function" %1 ["Volatile"] : f32
  return
}
```
```
func @access_chain_struct() -> () {
  %0 = spv.constant 1: index
  %1 = spv.Variable : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
  %2 = spv.AccessChain %1[%0, %0] : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
  return
}

```

I've some concerns about index type, regarding to the spec the index type must be an ConstantOp for the struct and an integer for other types, so, I've decided to take an IndexType from standard ops, but I'm not sure that's a right way to represent an index for spirv dialect.
@antiagainst can you please take a look, I'll be appriciate for any feedback.

### DIFF
--- a/include/mlir/Dialect/SPIRV/SPIRVBase.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVBase.td
@@ -96,6 +96,7 @@ def SPV_OC_OpFunctionEnd       : I32EnumAttrCase<"OpFunctionEnd", 56>;
 def SPV_OC_OpVariable          : I32EnumAttrCase<"OpVariable", 59>;
 def SPV_OC_OpLoad              : I32EnumAttrCase<"OpLoad", 61>;
 def SPV_OC_OpStore             : I32EnumAttrCase<"OpStore", 62>;
+def SPV_OC_OpAccessChain       : I32EnumAttrCase<"OpAccessChain", 65>;
 def SPV_OC_OpDecorate          : I32EnumAttrCase<"OpDecorate", 71>;
 def SPV_OC_OpCompositeExtract  : I32EnumAttrCase<"OpCompositeExtract", 81>;
 def SPV_OC_OpFMul              : I32EnumAttrCase<"OpFMul", 133>;
@@ -110,7 +111,8 @@ def SPV_OpcodeAttr :
       SPV_OC_OpConstantFalse, SPV_OC_OpConstant, SPV_OC_OpConstantComposite,
       SPV_OC_OpConstantNull, SPV_OC_OpFunction, SPV_OC_OpFunctionParameter,
       SPV_OC_OpFunctionEnd, SPV_OC_OpVariable, SPV_OC_OpLoad, SPV_OC_OpStore,
-      SPV_OC_OpDecorate, SPV_OC_OpCompositeExtract, SPV_OC_OpFMul, SPV_OC_OpReturn
+      SPV_OC_OpAccessChain, SPV_OC_OpDecorate, SPV_OC_OpCompositeExtract,
+      SPV_OC_OpFMul, SPV_OC_OpReturn
       ]> {
     let returnType = "::mlir::spirv::Opcode";
     let convertFromStorage = "static_cast<::mlir::spirv::Opcode>($_self.getInt())";
--- a/include/mlir/Dialect/SPIRV/SPIRVOps.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVOps.td
@@ -43,6 +43,64 @@ include "mlir/Dialect/SPIRV/SPIRVStructureOps.td"
 
 // -----
 
+def SPV_AccessChainOp : SPV_Op<"AccessChain", [NoSideEffect]> {
+  let summary = [{
+    Create a pointer into a composite object that can be used with OpLoad
+    and OpStore.
+  }];
+
+  let description = [{
+    Result Type must be an OpTypePointer. Its Type operand must be the type
+    reached by walking the Base’s type hierarchy down to the last provided
+    index in Indexes, and its Storage Class operand must be the same as the
+    Storage Class of Base.
+
+    Base must be a pointer, pointing to the base of a composite object.
+
+    Indexes walk the type hierarchy to the desired depth, potentially down
+    to scalar granularity. The first index in Indexes will select the top-
+    level member/element/component/element of the base composite. All
+    composite constituents use zero-based numbering, as described by their
+    OpType… instruction. The second index will apply similarly to that
+    result, and so on. Once any non-composite type is reached, there must be
+    no remaining (unused) indexes.
+
+     Each index in Indexes
+    - must be a scalar integer type,
+    - is treated as a signed count, and
+    - must be an OpConstant when indexing into a structure.
+
+    ### Custom assembly form
+    ``` {.ebnf}
+    access-chain-op ::= ssa-id `=` `spv.AccessChain` ssa-use
+                        `[` ssa-use (',' ssa-use)* `]`
+                        `:` pointer-type
+    ```
+
+    For example:
+
+    ```
+    %0 = "spv.constant"() { value = 1: i32} : () -> i32
+    %1 = spv.Variable : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
+    %2 = spv.AccessChain %1[%0] : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
+    %3 = spv.Load "Function" %2 ["Volatile"] : !spv.array<4xf32>
+    ```
+  }];
+
+  let arguments = (ins
+    SPV_AnyPtr:$base_ptr,
+    Variadic<SPV_Integer>:$indices
+  );
+
+  let results = (outs
+    SPV_AnyPtr:$component_ptr
+  );
+
+  let autogenSerialization = 0;
+}
+
+// -----
+
 def SPV_CompositeExtractOp : SPV_Op<"CompositeExtract", [NoSideEffect]> {
   let summary = "Extract a part of a composite object.";
 
--- a/lib/Dialect/SPIRV/SPIRVOps.cpp
+++ b/lib/Dialect/SPIRV/SPIRVOps.cpp
@@ -57,6 +57,21 @@ inline Dst bitwiseCast(Src source) noexcept {
   return dest;
 }
 
+static LogicalResult extractValueFromConstOp(Operation *op,
+                                             int32_t &indexValue) {
+  auto constOp = llvm::dyn_cast<spirv::ConstantOp>(op);
+  if (!constOp) {
+    return failure();
+  }
+  auto valueAttr = constOp.value();
+  auto integerValueAttr = valueAttr.dyn_cast<IntegerAttr>();
+  if (!integerValueAttr) {
+    return failure();
+  }
+  indexValue = integerValueAttr.getInt();
+  return success();
+}
+
 template <typename EnumClass>
 static ParseResult parseEnumAttribute(EnumClass &value, OpAsmParser *parser,
                                       OperationState *state) {
@@ -197,6 +212,129 @@ static LogicalResult verifyLoadStorePtrAndValTypes(LoadStoreOpTy op, Value *ptr,
 static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
   *printer << op->getName();
   printer->printOptionalAttrDict(op->getAttrs());
+}
+
+//===----------------------------------------------------------------------===//
+// spv.AccessChainOp
+//===----------------------------------------------------------------------===//
+
+static Type getElementPtrType(Type type, ArrayRef<Value *> indices,
+                              Location baseLoc) {
+  if (!indices.size()) {
+    emitError(baseLoc, "'spv.AccessChain' op expected at least "
+                       "one index ");
+    return nullptr;
+  }
+
+  auto ptrType = type.dyn_cast<spirv::PointerType>();
+  if (!ptrType) {
+    emitError(baseLoc, "'spv.AccessChain' op expected a pointer "
+                       "to composite type, but provided ")
+        << type;
+    return nullptr;
+  }
+
+  auto resultType = ptrType.getPointeeType();
+  auto resultStorageClass = ptrType.getStorageClass();
+  int32_t index = 0;
+
+  for (auto indexSSA : indices) {
+    auto cType = resultType.dyn_cast<spirv::CompositeType>();
+    if (!cType) {
+      emitError(baseLoc,
+                "'spv.AccessChain' op cannot extract from non-composite type ")
+          << resultType << " with index " << index;
+      return nullptr;
+    }
+    index = 0;
+    if (resultType.isa<spirv::StructType>()) {
+      Operation *op = indexSSA->getDefiningOp();
+      if (!op) {
+        emitError(baseLoc, "'spv.AccessChain' op index must be an "
+                           "integer spv.constant to access "
+                           "element of spv.struct");
+        return nullptr;
+      }
+
+      // TODO(denis0x0D): this should be relaxed to allow
+      // integer literals of other bitwidths.
+      if (failed(extractValueFromConstOp(op, index))) {
+        emitError(baseLoc,
+                  "'spv.AccessChain' index must be an integer spv.constant to "
+                  "access element of spv.struct, but provided ")
+            << op->getName();
+        return nullptr;
+      }
+      if (index < 0 || static_cast<uint64_t>(index) >= cType.getNumElements()) {
+        emitError(baseLoc, "'spv.AccessChain' op index ")
+            << index << " out of bounds for " << resultType;
+        return nullptr;
+      }
+    }
+    resultType = cType.getElementType(index);
+  }
+  return spirv::PointerType::get(resultType, resultStorageClass);
+}
+
+static ParseResult parseAccessChainOp(OpAsmParser *parser,
+                                      OperationState *state) {
+  OpAsmParser::OperandType ptrInfo;
+  SmallVector<OpAsmParser::OperandType, 4> indicesInfo;
+  Type type;
+  // TODO(denis0x0D): regarding to the spec an index must be any integer type,
+  // figure out how to use resolveOperand with a range of types and do not
+  // fail on first attempt.
+  Type indicesType = parser->getBuilder().getIntegerType(32);
+
+  if (parser->parseOperand(ptrInfo) ||
+      parser->parseOperandList(indicesInfo, OpAsmParser::Delimiter::Square) ||
+      parser->parseColonType(type) ||
+      parser->resolveOperand(ptrInfo, type, state->operands) ||
+      parser->resolveOperands(indicesInfo, indicesType, state->operands)) {
+    return failure();
+  }
+
+  Location baseLoc = state->operands.front()->getLoc();
+  auto resultType = getElementPtrType(
+      type, llvm::makeArrayRef(state->operands).drop_front(), baseLoc);
+  if (!resultType) {
+    return failure();
+  }
+
+  state->addTypes(resultType);
+  return success();
+}
+
+static void print(spirv::AccessChainOp op, OpAsmPrinter *printer) {
+  *printer << spirv::AccessChainOp::getOperationName() << ' ' << *op.base_ptr()
+           << '[';
+  printer->printOperands(op.indices());
+  *printer << "] : " << op.base_ptr()->getType();
+}
+
+static LogicalResult verify(spirv::AccessChainOp accessChainOp) {
+  SmallVector<Value *, 4> indices(accessChainOp.indices().begin(),
+                                  accessChainOp.indices().end());
+  auto resultType = getElementPtrType(accessChainOp.base_ptr()->getType(),
+                                      indices, accessChainOp.getLoc());
+  if (!resultType) {
+    return failure();
+  }
+
+  auto providedResultType =
+      accessChainOp.getType().dyn_cast<spirv::PointerType>();
+  if (!providedResultType) {
+    return accessChainOp.emitOpError(
+               "result type must be a pointer, but provided")
+           << providedResultType;
+  }
+
+  if (resultType != providedResultType) {
+    return accessChainOp.emitOpError("invalid result type: expected ")
+           << resultType << ", but provided " << providedResultType;
+  }
+
+  return success();
 }
 
 //===----------------------------------------------------------------------===//
--- a/test/Dialect/SPIRV/ops.mlir
+++ b/test/Dialect/SPIRV/ops.mlir
@@ -1,6 +1,125 @@
 // RUN: mlir-opt -split-input-file -verify-diagnostics %s | FileCheck %s
 
 //===----------------------------------------------------------------------===//
+// spv.AccessChain
+//===----------------------------------------------------------------------===//
+
+func @access_chain_struct() -> () {
+  %0 = spv.constant 1: i32
+  %1 = spv.Variable : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
+  // CHECK: {{.*}} = spv.AccessChain {{.*}}[{{.*}}, {{.*}}] : !spv.ptr<!spv.struct<f32, !spv.array<4 x f32>>, Function>
+  %2 = spv.AccessChain %1[%0, %0] : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
+  return
+}
+
+// -----
+
+func @access_chain_1D_array(%arg0 : i32) -> () {
+  %0 = spv.Variable : !spv.ptr<!spv.array<4xf32>, Function>
+  // CHECK: {{.*}} = spv.AccessChain {{.*}}[{{.*}}] : !spv.ptr<!spv.array<4 x f32>, Function>
+  %1 = spv.AccessChain %0[%arg0] : !spv.ptr<!spv.array<4xf32>, Function>
+  return
+}
+
+// -----
+
+func @access_chain_2D_array_1(%arg0 : i32) -> () {
+  %0 = spv.Variable : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
+  // CHECK: {{.*}} = spv.AccessChain {{.*}}[{{.*}}, {{.*}}] : !spv.ptr<!spv.array<4 x !spv.array<4 x f32>>, Function>
+  %1 = spv.AccessChain %0[%arg0, %arg0] : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
+  %2 = spv.Load "Function" %1 ["Volatile"] : f32
+  return
+}
+
+// -----
+
+func @access_chain_2D_array_2(%arg0 : i32) -> () {
+  %0 = spv.Variable : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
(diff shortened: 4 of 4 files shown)


antiagainst:
Good question. I think we want to use `SPV_Integer` here instead of index type for the following reasons:

* There is no hardware-dependent integer types in SPIR-V. All integer types are explicitly sized. We don't want to introduce something that has no correspondence in the SPIR-V spec unless there are compelling reasons. 
* `OpAccessChain` is clear regarding on the requirements on indices: 1) must be a scalar integer type, 2) is treated as a signed count, and 3) must be an OpConstant when indexing into a structure.

Right now I'm reusing standard `i<bitwidth>` types for signed integer types in SPIR-V. SPIR-V can also have unsigned types but it has been a confusing point thus far because SPIR-V want the instruction to dictate whether to interpret the integer as singed or unsigned (so we have `OpSDiv`, `OpUDiv`, etc.), not the type. So I'm a bit hesitating to introduce a special dialect type for unsigned integers. But I guess it's needed to convey the information for deserialization.

include/mlir/Dialect/SPIRV/SPIRVBase.td
```diff
@@ -143,14 +145,18 @@ def SPV_AnyPtr : Type<SPV_IsPtrType, "any SPIR-V pointer type">;
 def SPV_AnyArray : Type<SPV_IsArrayType, "any SPIR-V array type">;
 def SPV_AnyRTArray : Type<SPV_IsRTArrayType, "any SPIR-V runtime array type">;
 def SPV_AnyStruct : Type<SPV_IsStructType, "any SPIR-V struct type">;
+// TODO(denis0x0D): Find the better way to define an index type.
```
antiagainst: We should not introduce this type here as explained earlier.

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -43,6 +43,62 @@ include "mlir/Dialect/SPIRV/SPIRVStructureOps.td"
...
+    no remaining (unused) indexes.
+
+     Each index in Indexes
+    - must be a scalar integer type,
+    - is treated as a signed count, and
+    - must be an OpConstant when indexing into a structure.
+
+    ### Custom assembly form
+    ``` {.ebnf}
+    access_chain_op ::= ssa-id `=` `spv.AccessChain` ssa-use
```
antiagainst: Nit: `access-chain-op`

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -43,6 +43,62 @@ include "mlir/Dialect/SPIRV/SPIRVStructureOps.td"
...
+                        `[` ssa-use (',' ssa-use)* `]`
+                        `:` pointer-type
+    ```
+
+    For example:
+
+    ```
+    %0 = "spv.constant"() { value = 1: index} : () -> index
```
antiagainst: Can write as `spv.constant 1 : i32` here

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -43,6 +43,62 @@ include "mlir/Dialect/SPIRV/SPIRVStructureOps.td"
...
+    %1 = spv.Variable : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
+    %2 = spv.AccessChainOp %1[%0] : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
```
antiagainst: `spv.AccessChain`, no `Op` suffix

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -43,6 +43,62 @@ include "mlir/Dialect/SPIRV/SPIRVStructureOps.td"
...
+    %3 = spv.Load "Function" %2 ["Volatile"] : !spv.array<4xf32>
+    ```
+  }];
+
+  let arguments = (ins
+    SPV_AnyPtr:$base_ptr,
+    SPV_Optional<SPV_Index>:$indices
```
antiagainst: Please use `Variadic` instead of `SPV_Optional` here. We can have multiple indices. `SPV_Optional` is intended to support zero or one. It is aliased to `Variadic` right now, and that need to be fixed later (TODO for it).

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+
+  for (uint64_t i = 0; i < indicesInfo.size(); ++i) {
+    if (auto cType = resultType.dyn_cast<spirv::CompositeType>()) {
+      index = 0;
+      if (resultType.isa<spirv::StructType>()) {
+        Operation *op = state->operands[i + 1]->getDefiningOp();
+        if (!op) {
+          return parser->emitError(
+              indicesLocation, "index must be an integer constant to access "
+                               "element of StructType");
```
antiagainst: Nit: s/StructType/spv.struct/

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+  *printer << spirv::AccessChainOp::getOperationName() << ' ' << *op.base_ptr()
+           << '[';
+  printer->printOperands(op.indices());
+  *printer << "] : " << op.base_ptr()->getType();
+}
+
+static LogicalResult verify(spirv::AccessChainOp accessChainOp) {
+  auto resultType = accessChainOp.base_ptr()
+                        ->getType()
+                        .dyn_cast<spirv::PointerType>()
```
antiagainst: We need to check the `dyn_cast`'ed type is actually not null before getting its pointee type. In verifier we cannot assume too much. :)

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+      resultType = cType.getElementType(index);
+    } else {
+      return accessChainOp.emitOpError(
+                 "cannot extract from non-composite type ")
+             << resultType << " with index " << index;
+    }
+    ++indexCount;
+  }
+
+  if (!indexCount) {
```
antiagainst: Move this to happen before the check on each index?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+      accessChainOp.getType().dyn_cast<spirv::PointerType>();
+  if (!providedResultType) {
+    return accessChainOp.emitOpError(
+               "result type must be a pointer, but provided")
+           << providedResultType;
+  }
+
+  if (resultType != providedResultType.getPointeeType()) {
+    return accessChainOp.emitOpError(
+               "invalid result type: expected a pointer to ")
```
antiagainst: The type itself should print as a pointer type; so I think we don't need the `a pointer to` part in the message.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+  if (parser->parseOperand(ptrInfo) ||
+      parser->getCurrentLocation(&indicesLocation) ||
+      parser->parseOperandList(indicesInfo, OpAsmParser::Delimiter::Square) ||
+      parser->parseColonType(type) ||
+      parser->resolveOperand(ptrInfo, type, state->operands) ||
+      parser->resolveOperands(indicesInfo, indicesType, state->operands)) {
+    return failure();
+  }
+
+  if (!indicesInfo.size()) {
```
antiagainst: Can we extract this part (L245 - L287) and share them between the parser and the verifier? There is `mlir::emitError` that we can use; we just need to pass in the `Location` for error reporting: `return emitError(loc, "'spv.AccessChain' op ...")`.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
   printer->printOptionalAttrDict(op->getAttrs());
 }
 
+//===----------------------------------------------------------------------===//
+// spv.AccessChainOp
+//===----------------------------------------------------------------------===//
+
+// Function to extract an actual value from ConstantOp.
+static std::pair<StringRef, bool> extractValueFromConstOp(Operation *op,
```
antiagainst: I think we just need to return a bool to indicating whether the extraction is successful. This is a utility function so it's up to the callers to decide what suitable error message will be good for their cases. (From this utility function's perspective, it should not know about struct type.)

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+                                                          int32_t &indexValue) {
+  // spv.constant operation contains kValueAttrName.
+  auto valueAttr = op->getAttr(kValueAttrName);
```
antiagainst: We should actually `dyn_cast<spirv::ConstantOp>()` here to make sure this is actually from a constant op, not some other op with `value` attribute. And then we can use the `op.value()` accessor.

test/SPIRV/ops.mlir
```diff
@@ -1,5 +1,123 @@
 // RUN: mlir-opt -split-input-file -verify-diagnostics %s | FileCheck %s
 
+//===----------------------------------------------------------------------===//
+// spv.AccessChain
+//===----------------------------------------------------------------------===//
+
+func @access_chain_struct() -> () {
+  %0 = spv.constant 1: index
+  %1 = spv.Variable : !spv.ptr<!spv.struct<f32, !spv.array<4xf32>>, Function>
+  // CHECK: %2 = spv.AccessChain %1[%0, %0] : !spv.ptr<!spv.struct<f32, !spv.array<4 x f32>>, Function>
```
River707: Do not check the names of the SSA values directly, use regex instead:
(See https://github.com/tensorflow/mlir/blob/master/g3doc/TestingGuide.md)

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+        }
+        if (index < 0 ||
+            static_cast<uint64_t>(index) >= cType.getNumElements()) {
+          return parser->emitError(indicesLocation, "index ")
+                 << index << " out of bounds for " << type;
+        }
+      }
+      resultType = cType.getElementType(index);
+    } else {
+      return parser->emitError(indicesLocation,
```
River707: Restructure this to early exit instead:
auto cType = ...;
if (!cType)
 return parser->emitError

...

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +199,162 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+  }
+
+  resultStorageClass = ptrType.getStorageClass();
+  auto resultType = ptrType.getPointeeType();
+
+  for (uint64_t i = 0; i < indicesInfo.size(); ++i) {
+    if (auto cType = resultType.dyn_cast<spirv::CompositeType>()) {
+      index = 0;
+      if (resultType.isa<spirv::StructType>()) {
+        Operation *op = state->operands[i + 1]->getDefiningOp();
```
River707: This should likely be left to the verifier, there isn't a guarantee that this refers to what you expect when the IR is being parsed.

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -43,6 +43,62 @@ include "mlir/Dialect/SPIRV/SPIRVStructureOps.td"
...
+  );
+
+  let results = (outs
+    SPV_AnyPtr:$component_ptr
+  );
+}
```
MaheshRavishankar: Can you also set 
```
  let autogenSerialization=0
}
```

By default serialization methods are set to be autogenerated, but it would good to verify that it works for this Op. In a separate CL you can add serialization/deserialization test, and if something breaks, then the auto-generation can be fixed. Since this is not exercised here, better to disable the auto-generation

include/mlir/Dialect/SPIRV/SPIRVBase.td
denis0x0D: Fixed.

include/mlir/Dialect/SPIRV/SPIRVOps.td
denis0x0D: Thanks, fixed.

include/mlir/Dialect/SPIRV/SPIRVOps.td
denis0x0D: Fixed.

include/mlir/Dialect/SPIRV/SPIRVOps.td
denis0x0D: Thanks, fixed.

include/mlir/Dialect/SPIRV/SPIRVOps.td
denis0x0D: Thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: I see, thanks.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, moved.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Updated.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Updated, but have some concerns about an error message, in this case error message points to ssa value and not to the proccessing operation. I'm not sure how to convert llvmLoc to Location.
Thanks.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, updated.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Oh, I see, thanks. Fixed.

test/SPIRV/ops.mlir
denis0x0D: Thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, moved.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks for review, in case of AccessChain the index type for the !svp.struct must be an spv.constant, so, I hope it's a valid operation to get its value, in case of other composite type, we just getting the element type. It seems like the only way to get the resultType is walking into composite type by indices.  It was some small discussion in other tread, but I might miss something, please fix me if I'm wrong.
Thanks.

include/mlir/Dialect/SPIRV/SPIRVOps.td
denis0x0D: Thanks for pointing this. Fixed.

denis0x0D:
@antiagainst @River707 @MaheshRavishankar thanks for review!
I've updated regarding to the comments. I still have some concerns about index type, in this patch, I'm using i32, but the spec says any integer type, so I've left some comments in the code with TODO.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
   printer->printOptionalAttrDict(op->getAttrs());
 }
 
+//===----------------------------------------------------------------------===//
+// spv.AccessChainOp
+//===----------------------------------------------------------------------===//
+
+static Type extractPtrTypeFromPtrToCompositeType(ArrayRef<Value *> indices,
```
antiagainst: Nit: can we move `type` to be ahead of `indices`?

antiagainst: What about naming it as `getElementPtrType`?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+                                                 Type type, Location baseLoc) {
+  if (!indices.size()) {
+    return mlir::emitError(baseLoc, "'spv.AccessChain' op expected at least "
```
antiagainst: No need to qualify with `mlir::` given we have already `using namespace mlir;`.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+                                    "one index for spv.AccessChain"),
```
antiagainst: Redundant `for spv.AccessChain` part.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+           nullptr;
+  }
+
+  auto ptrType = type.dyn_cast<spirv::PointerType>();
+  if (!ptrType) {
+    return mlir::emitError(baseLoc, "'spv.AccessChain' op expected a pointer "
+                                    "to composite type, but provided ")
+               << type,
+           nullptr;
```
antiagainst: It would be better to use separate statements here given the error message takes several lines: `emitError(...); return nullptr;`. Similarly for other places in the function.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+                   baseLoc,
+                   "'spv.AccessChain' index must be an integer spv.constant to "
+                   "access element of spv.struct, but provided ")
+                   << op->getName(),
+               nullptr;
+      }
+
+      if (index < 0 || static_cast<uint64_t>(index) >= cType.getNumElements()) {
+        return mlir::emitError(baseLoc, "'spv.AccessChain' op index ")
+                   << index << " out of bounds for " << type,
```
antiagainst: Should we stream `resultType` instead of `type` here to be more specific?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+    if (resultType.isa<spirv::StructType>()) {
+      Operation *op = indexSSA->getDefiningOp();
+      if (!op) {
+        return mlir::emitError(baseLoc, "'spv.AccessChain' op index must be an "
+                                        "integer spv.constant to access "
+                                        "element of spv.struct"),
+               nullptr;
+      }
+
+      if (!extractValueFromConstOp(op, index)) {
```
antiagainst: Can you put a TODO here to say that this should be relaxed to allow integer literals of other bitwidths?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+  if (parser->parseOperand(ptrInfo) ||
+      parser->parseOperandList(indicesInfo, OpAsmParser::Delimiter::Square) ||
+      parser->parseColonType(type) ||
+      parser->resolveOperand(ptrInfo, type, state->operands) ||
+      parser->resolveOperands(indicesInfo, indicesType, state->operands)) {
+    return failure();
+  }
+
+  Location baseLoc = state->operands.front()->getLoc();
+  SmallVector<Value *, 4> indices(state->operands.begin() + 1,
```
antiagainst: I think we don't need to materialize a `SmallVector` here? Passing `llvm::makeArrayRef(state->operands).drop_front()` in should do.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+
+static void print(spirv::AccessChainOp op, OpAsmPrinter *printer) {
+  *printer << spirv::AccessChainOp::getOperationName() << ' ' << *op.base_ptr()
+           << '[';
+  printer->printOperands(op.indices());
+  *printer << "] : " << op.base_ptr()->getType();
+}
+
+static LogicalResult verify(spirv::AccessChainOp accessChainOp) {
+  SmallVector<Value *, 4> indices;
```
antiagainst: `SmallVector<Value *, 4> indices(accessChainOp.indices().begin(), accessChainOp.indices().end())`?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -199,6 +213,136 @@ static void printNoIOOp(Operation *op, OpAsmPrinter *printer) {
...
+  }
+  return spirv::PointerType::get(resultType, resultStorageClass);
+}
+
+static ParseResult parseAccessChainOp(OpAsmParser *parser,
+                                      OperationState *state) {
+  OpAsmParser::OperandType ptrInfo;
+  SmallVector<OpAsmParser::OperandType, 4> indicesInfo;
+  Type type;
+  // TODO(denis0x0D): regarding to the spec an index must be any integer type,
```
antiagainst: We may want to `parseOperand()` then `parseColonType()` for the operands. And iterate for all operands. This will give us a way to determine the type of each operand. But fine to leave a TODO for now.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Fixed.

denis0x0D: Fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Updated.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Oh, thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Fixed, thanks.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Thanks for mentioning this, if it's ok, I'll leave a TODO.

denis0x0D:
@antiagainst thanks a lot for review, I've updated it, can you please take a look.
Thanks.

lib/Dialect/SPIRV/SPIRVOps.cpp
antiagainst: SGTM!

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -57,6 +57,20 @@ inline Dst bitwiseCast(Src source) noexcept {
   return dest;
 }
 
+static bool extractValueFromConstOp(Operation *op, int32_t &indexValue) {
```
River707: This should return LogicalResult instead of bool

denis0x0D: Thanks, fixed.

denis0x0D:
@antiagainst @River707 thanks for review, updated to the latest comment.

denis0x0D:
@antiagainst сan I help with other tasks that no one works on?
Thanks.

antiagainst:
@denis0x0D: that is awesome; thanks for your help! Let me find some other tasks. :)

antiagainst:
Hey @denis0x0D,

We are working on basics to enable the conversion from other dialects to the SPIR-V dialect right now. Hopefully after a few days we will enable parallel development so that you can play with the conversion and possibly contribute there! :)

But at any time you are always welcome to add more op definitions given they are certainly needed anyway to have a full-fledged SPIR-V dialect! When doing that, just keep a few things in mind. Please ignore control flow related ops (section 3.32.17 in the spec) right now; we have separate plans for them. Please also make sure serialization and deserialization is covered with tests. Given that most of the stuff is auto-generated, there is no need for extensive tests if certain feature is already tested in a similar op (e.g., no need to test `OpFAdd` extensively given it's very similar to `OpFMul` which is already extensively tested). Other than that, feel free to pick whatever ops you'd like to implement or even flesh out some in a batch (like those binary arithmetic ops)! :)

denis0x0D:
@antiagainst thanks a lot! I'll definitely implement some other ops! Thanks for detailed description!

PR #58, opened 2019-07-31
Fix verification of zero-dim memref in affine.load/affine.store/std.load/std.store

dcaballe:
Verification complained when using zero-dimensional memrefs in affine.load, affine.store, std.load and std.store. Assuming that tests provided are expected use cases for affine and std loads and stores, this PR extends verification so that those memrefs can be used.


lib/AffineOps/AffineOps.cpp
```diff
@@ -1636,7 +1640,12 @@ LogicalResult AffineLoadOp::verify() {
   auto mapAttr = getAttrOfType<AffineMapAttr>(getMapAttrName());
   if (mapAttr) {
     AffineMap map = getAttrOfType<AffineMapAttr>(getMapAttrName()).getValue();
-    if (map.getNumResults() != getMemRefType().getRank())
+    auto rank = getMemRefType().getRank();
+    if (rank == 0) {
+      if (map.getNumResults() - 1 != rank)
```
ftynse: if (map.getNumResults() != 1) looks simpler and rank is guaranteed to be 0 due to the condition above

test/IR/load-store.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: mlir-opt %s -split-input-file | FileCheck %s
+
```
ftynse: These tests can do into test/IR/core-ops.mlir to live together with other load/store tests

dcaballe:
Thanks for the quick feedback, Alex! I'll address your comments after clarifying this point: 

> Do you have a specific use case where you want to use one subscript for a zero-dimensional memref? If not, I would think that using zero subscripts is more consistent, and also makes the verification code simpler (and can be reconsidered in future)

In fact, the use case we have is for one subscript. We are modeling single-scalar tensors as zero-dimensional tensors and lower them to zero-dimensional memrefs in affine. We want to create affine loads/stores for them but we need to provide a map for that, IIUC. We provide `{} -> 0` map, as you can see here: 
https://github.com/tensorflow/mlir/pull/58/files#diff-3744a45b757312019c0aa9667cd244eeR1600, so affine loads/stores end up with one subscript. Maybe we could provide a `{} -> {}` map? Could we add an interface to `Builder` that does that?

ftynse:
> Maybe we could provide a {} -> {} map? Could we add an interface to Builder that does that?

I think this would be great to have.  There may be places in the code that assume affine maps have non-zero outputs though.

dcaballe:
> > Maybe we could provide a {} -> {} map? Could we add an interface to Builder that does that?
> 
> I think this would be great to have. There may be places in the code that assume affine maps have non-zero outputs though.

Yeah, I did hit some asserts when I tried. Let me see how difficult would be to make that work and we can make a decision. Thanks!

ftynse:
LMK if there are too many changes involved. I'm fine with using constant zero subscript for zero-dimensional memrefs as a temporary solution. There are other places in the code that do that.

dcaballe:
It was easier than expected. Last commit should address all the comments.

include/mlir/IR/AffineMap.h
```diff
@@ -141,6 +144,9 @@ class AffineMap {
 
 private:
   ImplType *map;
+
+  static AffineMap get_impl(unsigned dimCount, unsigned symbolCount,
```
ftynse: please use camel case for function names (MLIR follows LLVM style)

include/mlir/IR/Builders.h
```diff
@@ -148,6 +148,8 @@ class Builder {
                          ArrayRef<AffineExpr> results);
 
   // Special cases of affine maps and integer sets
+  /// Returns a zero result affine map with no dimensions or symbols: () -> ().
+  AffineMap getNullAffineMap(MLIRContext *context);
```
ftynse: This sounds like it would return an AffineMap wrapping a null pointer (which is what AffineMap() gives you).  Please consider something like getEmptyAffineMap or getZeroOutputAffineMap instead.

lib/AffineOps/AffineOps.cpp
```diff
@@ -1642,9 +1642,10 @@ LogicalResult AffineLoadOp::verify() {
     AffineMap map = getAttrOfType<AffineMapAttr>(getMapAttrName()).getValue();
     auto rank = getMemRefType().getRank();
     if (rank == 0) {
```
ftynse: This special case should become unnecessary with 0D maps

lib/AffineOps/AffineOps.cpp
```diff
@@ -1740,9 +1741,9 @@ LogicalResult AffineStoreOp::verify() {
     AffineMap map = mapAttr.getValue();
     auto rank = getMemRefType().getRank();
     if (rank == 0) {
```
ftynse: same

lib/StandardOps/Ops.cpp
```diff
@@ -1681,8 +1681,7 @@ static LogicalResult verify(LoadOp op) {
   auto rank = op.getMemRefType().getRank();
   auto numOperands = op.getNumOperands();
   if (rank == 0) {
```
ftynse: same as above

lib/StandardOps/Ops.cpp
```diff
@@ -1999,8 +1998,7 @@ static LogicalResult verify(StoreOp op) {
...
```
ftynse: same as above

dcaballe:
I think I addressed all the comments. Please, let me know if something else is missing. Thanks!

lib/IR/Builders.cpp
```diff
@@ -297,6 +297,10 @@ IntegerSet Builder::getIntegerSet(unsigned dimCount, unsigned symbolCount,
   return IntegerSet::get(dimCount, symbolCount, constraints, isEq);
 }
 
+AffineMap Builder::getEmptyAffineMap(MLIRContext *context) {
+  return AffineMap::get(context);
```
ftynse: Builder stores a context, there is no need to pass another context in this call.

dcaballe:
Good catch. Thanks!

River707:
Make sure to squash your commits.

dcaballe:
For those doing the merge, there is an option to merge the changes doing a squash. That's the approach that I've seen in other projects. Are you asking me to do a explicit squash?

joker-eph:
Our infra will squash anyway.
The only advantage of the single commit is that we don’t have to reword the message ourselves on integration (otherwise it includes all the commits in the squashed commit message).

dcaballe:
The problem with me doing the squash and a push force is that, IIRC, this messed up the review comments, but I don't remember the details. Yeah, just let me know. I can do a squash including my commits and the merge commits.

River707:
(Sorry, I was OOO for a while a missed the auto-squash update. Please disregard my comment then.)

PR #61, opened 2019-08-01
[spirv] Add binary logical operations.

denis0x0D:
This patch adds binary logical operations regarding to the spec section 3.32.15:
OpIEqual, OpINotEqual, OpUGreaterThan, OpSGreaterThan,
OpUGreaterThanEqual, OpSGreaterThanEqual, OpULessThan, OpSLessThan,
OpULessThanEqual, OpSLessThanEqual, this ops are useful for integer comparison.

I've splitted bianary ops to logical ops and arithmetic ops.
Also added custom parser, since default one does not fit the requirements (result type is different to operands type), and printer, verifier.

Tests added to ops.mlir and to  serialization test machinery.
Some tests related to arithmetic ops in ops.mlir reordered to the lexicographic order, since I was mistaken in previous commit.

Thanks.

### DIFF
--- a/include/mlir/Dialect/SPIRV/SPIRVBase.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVBase.td
@@ -113,6 +113,16 @@ def SPV_OC_OpSRem              : I32EnumAttrCase<"OpSRem", 138>;
 def SPV_OC_OpSMod              : I32EnumAttrCase<"OpSMod", 139>;
 def SPV_OC_OpFRem              : I32EnumAttrCase<"OpFRem", 140>;
 def SPV_OC_OpFMod              : I32EnumAttrCase<"OpFMod", 141>;
+def SPV_OC_OpIEqual            : I32EnumAttrCase<"OpIEqual", 170>;
+def SPV_OC_OpINotEqual         : I32EnumAttrCase<"OpINotEqual", 171>;
+def SPV_OC_OpUGreaterThan      : I32EnumAttrCase<"OpUGreaterThan", 172>;
+def SPV_OC_OpSGreaterThan      : I32EnumAttrCase<"OpSGreaterThan", 173>;
+def SPV_OC_OpUGreaterThanEqual : I32EnumAttrCase<"OpUGreaterThanEqual", 174>;
+def SPV_OC_OpSGreaterThanEqual : I32EnumAttrCase<"OpSGreaterThanEqual", 175>;
+def SPV_OC_OpULessThan         : I32EnumAttrCase<"OpULessThan", 176>;
+def SPV_OC_OpSLessThan         : I32EnumAttrCase<"OpSLessThan", 177>;
+def SPV_OC_OpULessThanEqual    : I32EnumAttrCase<"OpULessThanEqual", 178>;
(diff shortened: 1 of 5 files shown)


include/mlir/Dialect/SPIRV/SPIRVBase.td
```diff
@@ -683,22 +697,39 @@ class SPV_Op<string mnemonic, list<OpTrait> traits = []> :
...
+      SPV_BinaryOp<mnemonic, type, type,
+                   !listconcat(traits,
+                               [NoSideEffect, SameOperandsAndResultType])> {
+}
+
+class SPV_LogicalOp<string mnemonic, Type operandsType,
+                    list<OpTrait> traits = []> :
+      // Result type is SPV_Bool.
+      SPV_BinaryOp<mnemonic, SPV_Bool, operandsType,
+                   !listconcat(traits, [NoSideEffect, SameTypeOperands])> {
```
antiagainst: We may also want `SameOperandsAndResultShape` here.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -154,6 +177,37 @@ printMemoryAccessAttribute(LoadStoreOpTy loadStoreOp, OpAsmPrinter *printer,
   elidedAttrs.push_back(spirv::attributeName<spirv::StorageClass>());
 }
 
+static LogicalResult verifyBinaryLogicalOp(Operation *logicalOp) {
```
antiagainst: We don't need this manual checks anymore with `SameOperandsAndResultShape`.

antiagainst:
Hey @denis0x0D, one question regarding the `define_inst.sh` script. I think it does not handle the case of subclassing and omitting arguments/results well. Have you experienced that? Did you fixed it or something? Would be nice to upstream if you already got a fix for it. :)

MaheshRavishankar:
No more comments from me. Thanks!

include/mlir/Dialect/SPIRV/SPIRVBase.td
denis0x0D: Oh, thanks! Fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Fixed, thanks!

denis0x0D:
@antiagainst @MaheshRavishankar thanks a lot for review!
@antiagainst it does not work with subclasses, I've already generated a susbset of operation, before the commit with arithmetic operations. I did not have a fix for this moment, but I can take a look, since it saves a lot of time.

antiagainst:
Oh, I see. It also might be easier if we create a new `.td` file for these binary ops and move them there so the main `SPIRVOps.td` file can use the existing script.

antiagainst:
But it does mean we may not have the automatic updates of documentation on those binary ops. But that would be fine for now given that the spec is fairly stable and changes with low frequency.

denis0x0D:
@antiagainst if it's ok, I can move them to a new *.td, with the next patch for other bin operations.

antiagainst:
SGTM! Please have a PR for just moving stuff so it's easier to review. :)

denis0x0D:
@antiagainst got it, thanks!

PR #64, opened 2019-08-02
Add the LLVM IR unreachable instruction to the LLVMIR dialect.

schweitzpgi:
http://llvm.org/docs/LangRef.html#unreachable-instruction

This instruction will be to support (otherwise regular) calls to the Flang
runtime that are known to never return to the caller.

### DIFF
--- a/include/mlir/LLVMIR/LLVMOps.td
+++ b/include/mlir/LLVMIR/LLVMOps.td
@@ -301,14 +301,9 @@ def LLVM_ReturnOp : LLVM_TerminatorOp<"return", []> {
   let printer = [{ printReturnOp(p, *this); }];
 }
 def LLVM_UnreachableOp : LLVM_TerminatorOp<"unreachable", []> {
-  string llvmBuilder = [{
-    builder.CreateUnreachable();
-  }];
-  let verifier = [{
-    return success();
-  }];
-  let parser = [{ return parseUnreachableOp(parser, result); }];
-  let printer = [{ printUnreachableOp(p, *this); }];
+  string llvmBuilder = [{ builder.CreateUnreachable(); }];
+  let parser = [{ return success(); }];
+  let printer = [{ *p << getOperationName(); }];
 }
 
 // Pseudo-operations (do not appear in LLVM IR but necessary for the dialect to
--- a/include/mlir/LLVMIR/LLVMOps.td
+++ b/include/mlir/LLVMIR/LLVMOps.td
@@ -300,6 +300,16 @@ def LLVM_ReturnOp : LLVM_TerminatorOp<"return", []> {
   let parser = [{ return parseReturnOp(parser, result); }];
   let printer = [{ printReturnOp(p, *this); }];
 }
+def LLVM_UnreachableOp : LLVM_TerminatorOp<"unreachable", []> {
+  string llvmBuilder = [{
+    builder.CreateUnreachable();
+  }];
+  let verifier = [{
+    return success();
+  }];
+  let parser = [{ return parseUnreachableOp(parser, result); }];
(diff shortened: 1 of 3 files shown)


schweitzpgi:
Nvidia has signed the CLA.

@googlebot I signed it!

include/mlir/LLVMIR/LLVMOps.td
```diff
@@ -300,6 +300,16 @@ def LLVM_ReturnOp : LLVM_TerminatorOp<"return", []> {
   let parser = [{ return parseReturnOp(parser, result); }];
   let printer = [{ printReturnOp(p, *this); }];
 }
+def LLVM_UnreachableOp : LLVM_TerminatorOp<"unreachable", []> {
+  string llvmBuilder = [{
+    builder.CreateUnreachable();
+  }];
+  let verifier = [{
+    return success();
```
River707: I don't think it's necessary to define the verifier if you don't need to.

lib/LLVMIR/IR/LLVMDialect.cpp
```diff
@@ -635,6 +635,10 @@ static void printReturnOp(OpAsmPrinter *p, ReturnOp &op) {
   *p << ' ' << *op.getOperand(0) << " : " << op.getOperand(0)->getType();
 }
 
+static void printUnreachableOp(OpAsmPrinter *p, UnreachableOp &op) {
+  *p << op.getOperationName() << " : () -> ()";
```
River707: Don't print a trailing empty functional type, it isn't necessary.

River707: On that note, can we just inline this function into the op definition? It is small enough.

lib/LLVMIR/IR/LLVMDialect.cpp
```diff
@@ -653,6 +657,15 @@ static ParseResult parseReturnOp(OpAsmParser *parser, OperationState *result) {
   return success();
 }
 
+// <operation> ::= `llvm.unreachable` `:` type-list-no-parens
+static ParseResult parseUnreachableOp(OpAsmParser *parser,
+                                      OperationState *result) {
+  Type type;
```
River707: This parser shouldn't need to do anything, just inline it into the op definition.

schweitzpgi:
Thanks for the review, River. I've made your requested changes.

River707:
Can you add a test to "test/Target/llvmir.mlir" to test the conversion to LLVM IR?

schweitzpgi:
> Can you add a test to "test/Target/llvmir.mlir" to test the conversion to LLVM IR?

Sure.

test/Target/llvmir.mlir
```diff
@@ -825,3 +825,9 @@ func @stringconstant() -> !llvm<"i8*"> {
   llvm.return %1 : !llvm<"i8*">
 }
 
+// CHECK-LABEL: define void @noreach() {
+// CHECK-NEXT:    unreachable
```
River707: We don't need to check the trailing '}' here.

joker-eph:
CLA overridden: contribution from Nvidia.

test/Target/llvmir.mlir
schweitzpgi: I see your point, however I left it in there to be consistent with some similar checks earlier in the file.

River707: It's alright to diverge here, we can clean the rest in a followup.

schweitzpgi: It looks like these last few tests cherry pick the LLVM IR output. If that's the new preferred pattern, perhaps I should remove the CHECK-LABEL line as well.

PR #70, opened 2019-08-08
Command toyc should be toyc-ch2 in this chapter

chenkanhw:
The output was 'toyc-ch2' after compilation instead of 'toyc'.


joker-eph:
Thanks!

PR #71, opened 2019-08-08
Enable TTI for host TargetMachine in JitRunner

dcaballe:
This commit improves JitRunner so that it creates a target machine
for the current CPU host which is used to properly initialize LLVM's
TargetTransformInfo for such a target. This will enable optimizations
such as vectorization in LLVM when using JitRunner. Please, note that,
as part of this work, JITTargetMachineBuilder::detectHost() has been
extended to include the host CPU name and sub-target features as part of
the host CPU detection (https://reviews.llvm.org/D65760).


dcaballe:
I don't have access to the report on the checks that fail. Please, let me know if it's something I should take care of.

PR #74, opened 2019-08-13
Fix build of affine load/store with empty map

dcaballe:
https://github.com/tensorflow/mlir/pull/58 fixed and exercised verification of load/store ops using empty affine maps. Unfortunately, it didn't exercise the creation of them. This PR addresses this aspect by:
1. Removing the assumption of AffineMap having at least one result. This means that we no longer can do `map.getResult(0)->getContext()`.
2.  Storing a pointer to MLIRContext as member of AffineMap to perverse previous functionality.

Initially, I tried to propagate MLIRContext through the interfaces instead of storing it in AffineMap but it impacted too many interfaces, including `AffineMapAttr` and templatized functions expecting common API across attributes (https://github.com/tensorflow/mlir/blob/master/lib/Transforms/SimplifyAffineStructures.cpp#L62). Please, let me know if you think there is a better approach.

I'd also appreciate some suggestions on how to test affine.load/affine.store builders. Is there a dialect I can use for that?

Thanks!
Diego


bondhugula:
> #58 fixed and exercised verification of load/store ops using empty affine maps. Unfortunately, it didn't exercise the creation of them. This PR addresses this aspect by:
> 
> 1. Removing the assumption of AffineMap having at least one result. This means that we no longer can do `map.getResult(0)->getContext()`.
> 2. Storing a pointer to MLIRContext as member of AffineMap to perverse previous functionality.
> 
> Initially, I tried to propagate MLIRContext through the interfaces instead of storing it in AffineMap but it impacted too many interfaces, including `AffineMapAttr` and templatized functions expecting common API across attributes (https://github.com/tensorflow/mlir/blob/master/lib/Transforms/SimplifyAffineStructures.cpp#L62). Please, let me know if you think there is a better approach.
> 
> I'd also appreciate some suggestions on how to test affine.load/affine.store builders. Is there a dialect I can use for that?
> 

As far as I know, unfortunately, I don't think there is a direct way to test build methods of ops. An indirect way would be to write a small test case for -affine-data-copy-generate - it can generate copies to a scratchpad, and in the case of 0-d memref's, the buffers will also be trivially 0-d, i.e., the build method would be called load's/store's on 0-d memrefs. 

```
func @foo(%A : memref<vector<8xf32>>) {
  %v = load %A[] : memref<vector<8xf32>>
   return
}
```
$ mlir-opt -affine-data-copy-generate -affine-data-copy-generate-dma=false test.mlir

> Thanks!
> Diego

bondhugula:
> > #58 fixed and exercised verification of load/store ops using empty affine maps. Unfortunately, it didn't exercise the creation of them. This PR addresses this aspect by:
> > 
> > 1. Removing the assumption of AffineMap having at least one result. This means that we no longer can do `map.getResult(0)->getContext()`.
> > 2. Storing a pointer to MLIRContext as member of AffineMap to perverse previous functionality.
> > 

It think it looks fine to store a pointer to MLIRContext in AffineMap. If we really want to save space in the future, I think one could use a PointerUnion<ArrayRef<AffineExpr>, MLIRContext *> since we either have an ArrayRef of affine expr's or an MLIRContext?

dcaballe:
Thanks Uday!

> ```
> func @foo(%A : memref<vector<8xf32>>) {
>   %v = load %A[] : memref<vector<8xf32>>
>    return
> }
> ```
> 
> $ mlir-opt -affine-data-copy-generate -affine-data-copy-generate-dma=false test.mlir

Great suggestion. However, I'm worried about depending on other optimizations to test basic infra stuff. I wondered if creating a C++ unit test that calls the builder API to create a 0-dim load/store would work. What do you think?

> > > 2. Storing a pointer to MLIRContext as member of AffineMap to perverse previous functionality.
> 
> I think it looks fine to store a pointer to MLIRContext in AffineMap. If we really want to save space in the future, I think one could use a PointerUnion<ArrayRef, MLIRContext *> since we either have an ArrayRef of affine expr's or an MLIRContext?

+1

bondhugula:
> Thanks Uday!
> 
> > ```
> > func @foo(%A : memref<vector<8xf32>>) {
> >   %v = load %A[] : memref<vector<8xf32>>
> >    return
> > }
> > ```
> > 
> > 
> > $ mlir-opt -affine-data-copy-generate -affine-data-copy-generate-dma=false test.mlir
> 
> Great suggestion. However, I'm worried about depending on other optimizations to test basic infra stuff. I wondered if creating a C++ unit test that calls the builder API to create a 0-dim load/store would work. What do you think?

Having such a builder API test would be great - AFAIK there isn't one for any of the standard/affine ops. Something like that does exist for the EDSC builder though:
https://github.com/tensorflow/mlir/blob/master/test/EDSC/builder-api-test.cpp
Something with a similar structure could perhaps be created to exercise these cases.

dcaballe:
> Having such a builder API test would be great - AFAIK there isn't one for any of the standard/affine ops. Something like that does exist for the EDSC builder though:
> https://github.com/tensorflow/mlir/blob/master/test/EDSC/builder-api-test.cpp
> Something with a similar structure could perhaps be created to exercise these cases.

Actually, I think using EDSC would be appropriate here since we have to check that EDSC supports empty masks and it's invoking the builders correctly. I added a test accordingly and added same fix (sorry, I missed it before) for affine.store. WDYT?

include/mlir/IR/AffineMap.h
```diff
@@ -47,9 +47,10 @@ class AffineMap {
 public:
   using ImplType = detail::AffineMapStorage;
 
-  AffineMap() : map(nullptr) {}
-  explicit AffineMap(ImplType *map) : map(map) {}
-  AffineMap(const AffineMap &other) : map(other.map) {}
+  AffineMap() : map(nullptr), context(nullptr) {}
+  explicit AffineMap(ImplType *map, MLIRContext *context)
+      : map(map), context(context) {}
```
River707: Why not store the context in the AffineMapStorage? Otherwise you are duplicating per instance, and AffineMap is now two words instead of one.

dcaballe: Good point! Yeah, it makes more sense. Done. Thanks!

dcaballe:
Please, let me know if there are any other comments.

Thanks!
Diego

ftynse:
Most of the builders are only covered by tests indirectly (e.g. through rewrite patterns or passes that create the relevant operations).  Until this creates recurrent problems, this is a trade-off we are willing to take to reduce the complexity of the test infrastructure.

Issue #76, opened 2019-08-13
[spirv] Extend spv.array with layoutinfo and update serialization


denis0x0D:
The spv.array type must be extended with layoutinfo to support decoration:
`OpDecorate %array ArrayStride 4`
Note: this issue is a blocker for https://github.com/tensorflow/mlir/issues/60

denis0x0D:
@antiagainst please let me know if it's ok, if I take this one.
Thanks!

antiagainst:
SGTM, thanks Denis!

denis0x0D:
this is closed by https://github.com/tensorflow/mlir/commit/43cb2e491b095b331659e298f3bb536b29583e56

Issue #77, opened 2019-08-13
[spirv] Support serialization for spv.struct.


denis0x0D:
Serialization for spv.struct must be added alongside with decoration with BufferBlock.

_BufferBlock apply to a structure type to establish it is an SSBO-like shader-interface block._

`
OpDecorate %structType BufferBlock`

Note: it is a blocker for https://github.com/tensorflow/mlir/issues/60

denis0x0D:
@antiagainst @MaheshRavishankar I've started to work on this.
Thanks!

antiagainst:
Cool, thanks @denis0x0D!

MaheshRavishankar:
Fantastic @denis0x0D ! Please let me know if you need any help.

denis0x0D:
@antiagainst @MaheshRavishankar thanks a lot!
@MaheshRavishankar serialization for struct type and members decoration looks straight forward, since struct type already has layoutinfo. But I'm not sure about decoration with `Buffer` or `BufferBlock` , the spec requires to decorate struct type with BufferBlock :

> Apply to a structure type to establish it is an
> SSBO-like shader-interface block

and Block:

> Apply to a structure type to establish it is a
> non-SSBO-like shader-interface block

it seems like we need an additional field in the StructType to specify SSBO or non-SSBO, and therefore update the ebnf of this type, but I might miss something, please fix me if I'm wrong.
Thanks.

antiagainst:
Good question! 

It's a confusing point that by applying `BufferBlock` decoration, you can actually write to `Uniform` storage class: `Uniform` storage class together with `BufferBlock` actually means `StorageBuffer` storage class. [SPV_KHR_storage_buffer_storage_class](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_storage_buffer_storage_class.html) extension is introduced to address this and it becomes core since SPIR-V 1.4. `BufferBlock` decoration is depreciated and will be removed from SPIR-V 1.4. So I'd suggest we always use `Block` decoration and choose `Uniform` and `StorageBuffer` storage class to differentiate to be future proof. (Yes it requires the `SPV_KHR_storage_buffer_storage_class` extension for using `StorageBuffer` storage class pre-1.4 but it is fairly commonly implemented.)

During serialization of the struct type, there is another thing to pay attention to. Per the spec:
[...]

Issue #68, opened 2019-08-05
Add spv.FuncOp to SPIR-V dialect


MaheshRavishankar:
Currently SPIR-V dialect is using the FuncOp from StandardOps dialect. FuncOp do not allow implicit capture, whereas the SPIR-V specification requires that functions in SPIR-V dialect have access to global variables defined outside off the function. SPIR-V dialect needs to have its own FuncOp to allow such functionality.

MaheshRavishankar:
Looking into adding this right now.

River707:
Is the SPIRV dialect only intended to live on the edge, like the LLVM dialect, or are there optimizations that are expected to run on this form?

joker-eph:
Can you rely on the symbol table for accesses to global variable defined outside the function? (just like `std.call` is also referencing global FuncOp outside)

MaheshRavishankar:
@River707 : Yes SPIR-V is intended to be like LLVM dialect. So I dont see a conversion from SPIR-V to any other dialect. Indeed the SPIR-V dialect has been designed to be able to serialize it into the SPIR-V binary format easily (most of it is auto-generated from ODS).

@joker-eph  : I don't know if that would work. That would have to be named symbols. A function in spir-v module should be able to refer to any value generated from a spv.Constant or an spv.Variable as follows

```
%0 = spv.Variable : !spv.ptr<f32, StorageBuffer>
func @foo() {
   %1 = spv.Load "StorageBuffer" %0 : f32
}
```

A lot of the SPIR-V spec relies on this. The same mechanism is used to access global invocation IDS, workgroups size, etc. It seems the easiest solution is to allow refering to global variables.

antiagainst:
The SPIR-V dialect is [designed](https://github.com/tensorflow/mlir/blob/master/g3doc/Dialects/SPIR-V.md#design-principles) to be a "proxy" to the SPIR-V binary format. It should stay as the same semantic level and try to be a mechanical 1:1 mapping to SPIR-V binary counterparts if possible. It should be easy to (de)serialize with the binary format.

[...]

PR #80, opened 2019-08-14
[spirv] Extend spv.array with Layoutinfo

denis0x0D:
This patch is addressing issue 
https://github.com/tensorflow/mlir/issues/76

1. The ebnf is updated to support layoutinfo:
For example, spv.array with ArrayStride information:
`!spv.array< 4 x !spv.array<4 x f32 [4]> [128]>`
 The check for layoutinfo in the parser is relaxed, the checks in the (de)serializer generate erros in case ArrayStride is not present for OpTypeArray regarding to the spec.
2. Tests related to serialization machinery updated.
3. Changed specialization of template function from `typename StructType::LayoutInfo_` to `typename uint64_t` to be able to reuse it for ArrayType::LayoutInfo.

Thanks!

### DIFF
--- a/include/mlir/Dialect/SPIRV/SPIRVTypes.h
+++ b/include/mlir/Dialect/SPIRV/SPIRVTypes.h
@@ -73,14 +73,23 @@ class ArrayType : public Type::TypeBase<ArrayType, CompositeType,
                                         detail::ArrayTypeStorage> {
 public:
   using Base::Base;
+  // Zero layout specifies that is no layout
+  using LayoutInfo = uint64_t;
 
   static bool kindof(unsigned kind) { return kind == TypeKind::Array; }
 
   static ArrayType get(Type elementType, unsigned elementCount);
 
+  static ArrayType get(Type elementType, unsigned elementCount,
+                       LayoutInfo layoutInfo);
+
   unsigned getNumElements() const;
 
   Type getElementType() const;
+
+  bool hasLayout() const;
+
+  uint64_t getArrayStride() const;
 };
 
 // SPIR-V image type
--- a/lib/Dialect/SPIRV/SPIRVDialect.cpp
+++ b/lib/Dialect/SPIRV/SPIRVDialect.cpp
@@ -53,6 +53,18 @@ SPIRVDialect::SPIRVDialect(MLIRContext *context)
 // Type Parsing
 //===----------------------------------------------------------------------===//
 
+// Forward declarations.
+template <typename ValTy>
+static Optional<ValTy> parseAndVerify(SPIRVDialect const &dialect, Location loc,
+                                      StringRef spec);
+template <>
+Optional<Type> parseAndVerify<Type>(SPIRVDialect const &dialect, Location loc,
+                                    StringRef spec);
+
+template <>
+Optional<uint64_t> parseAndVerify(SPIRVDialect const &dialect, Location loc,
+                                  StringRef spec);
+
 // Parses "<number> x" from the beginning of `spec`.
 static bool parseNumberX(StringRef &spec, int64_t &number) {
   spec = spec.ltrim();
@@ -150,7 +162,8 @@ static Type parseAndVerifyType(SPIRVDialect const &dialect, StringRef spec,
 //                | vector-type
 //                | spirv-type
 //
-// array-type ::= `!spv.array<` integer-literal `x` element-type `>`
+// array-type ::= `!spv.array<` integer-literal `x` element-type
+//                (`[` integer-literal `]`)? `>`
 static Type parseArrayType(SPIRVDialect const &dialect, StringRef spec,
                            Location loc) {
   if (!spec.consume_front("array<") || !spec.consume_back(">")) {
@@ -171,11 +184,37 @@ static Type parseArrayType(SPIRVDialect const &dialect, StringRef spec,
     return Type();
   }
 
+  ArrayType::LayoutInfo layoutInfo = 0;
+  size_t lastLSquare;
+
+  // Handle case when element type is not a trivial type
+  auto lastRDelimiter = spec.rfind('>');
+  if (lastRDelimiter != StringRef::npos) {
+    lastLSquare = spec.find('[', lastRDelimiter);
+  } else {
+    lastLSquare = spec.rfind('[');
+  }
+
+  if (lastLSquare != StringRef::npos) {
+    auto layoutSpec = spec.substr(lastLSquare);
+    auto layout =
+        parseAndVerify<ArrayType::LayoutInfo>(dialect, loc, layoutSpec);
+    if (!layout) {
+      return Type();
+    }
+
+    if (!(layoutInfo = layout.getValue())) {
+      emitError(loc, "ArrayStride must be greater than zero");
+      return Type();
+    }
+    spec = spec.substr(0, lastLSquare);
+  }
+
   Type elementType = parseAndVerifyType(dialect, spec, loc);
   if (!elementType)
     return Type();
 
-  return ArrayType::get(elementType, count);
+  return ArrayType::get(elementType, count, layoutInfo);
 }
 
 // TODO(ravishankarm) : Reorder methods to be utilities first and parse*Type
@@ -267,18 +306,17 @@ Optional<Type> parseAndVerify<Type>(SPIRVDialect const &dialect, Location loc,
 }
 
 template <>
-Optional<spirv::StructType::LayoutInfo>
-parseAndVerify(SPIRVDialect const &dialect, Location loc, StringRef spec) {
+Optional<uint64_t> parseAndVerify(SPIRVDialect const &dialect, Location loc,
+                                  StringRef spec) {
   uint64_t offsetVal = std::numeric_limits<uint64_t>::max();
   if (!spec.consume_front("[")) {
     emitError(loc, "expected '[' while parsing layout specification in '")
         << spec << "'";
     return llvm::None;
   }
+  spec = spec.trim();
   if (spec.consumeInteger(10, offsetVal)) {
-    emitError(
-        loc,
-        "expected unsigned integer to specify offset of member in struct: '")
+    emitError(loc, "expected unsigned integer to specify layout information: '")
         << spec << "'";
     return llvm::None;
   }
@@ -292,7 +330,7 @@ parseAndVerify(SPIRVDialect const &dialect, Location loc, StringRef spec) {
         << spec << "'";
     return llvm::None;
   }
-  return spirv::StructType::LayoutInfo{offsetVal};
+  return offsetVal;
 }
 
 // Functor object to parse a comma separated list of specs. The function
@@ -530,8 +568,11 @@ Type SPIRVDialect::parseType(StringRef spec, Location loc) const {
 //===----------------------------------------------------------------------===//
 
 static void print(ArrayType type, llvm::raw_ostream &os) {
-  os << "array<" << type.getNumElements() << " x " << type.getElementType()
-     << ">";
+  os << "array<" << type.getNumElements() << " x " << type.getElementType();
+  if (type.hasLayout()) {
+    os << " [" << type.getArrayStride() << "]";
+  }
+  os << ">";
 }
 
 static void print(RuntimeArrayType type, llvm::raw_ostream &os) {
--- a/lib/Dialect/SPIRV/Serialization/Deserializer.cpp
+++ b/lib/Dialect/SPIRV/Serialization/Deserializer.cpp
@@ -207,6 +207,9 @@ private:
   // Result <id> to decorations mapping.
   DenseMap<uint32_t, NamedAttributeList> decorations;
 
+  // Result <id> to type decorations.
+  DenseMap<uint32_t, uint32_t> typeDecorations;
+
   // List of instructions that are processed in a defered fashion (after an
   // initial processing of the entire binary). Some operations like
   // OpEntryPoint, and OpExecutionMode use forward references to function
@@ -329,6 +332,13 @@ LogicalResult Deserializer::processDecoration(ArrayRef<uint32_t> words) {
     decorations[words[0]].set(opBuilder.getIdentifier(attrName),
                               opBuilder.getStringAttr(stringifyBuiltIn(
                                   static_cast<spirv::BuiltIn>(words[2]))));
+    break;
+  case spirv::Decoration::ArrayStride:
+    if (words.size() != 3) {
+      return emitError(unknownLoc, "OpDecorate with ")
+             << decorationName << " needs a single integer literal";
+    }
+    typeDecorations[words[0]] = static_cast<uint32_t>(words[2]);
     break;
   default:
     return emitError(unknownLoc, "unhandled Decoration : '") << decorationName;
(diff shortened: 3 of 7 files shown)


lib/Dialect/SPIRV/SPIRVDialect.cpp
```diff
@@ -150,7 +162,8 @@ static Type parseAndVerifyType(SPIRVDialect const &dialect, StringRef spec,
 //                | vector-type
 //                | spirv-type
 //
-// array-type ::= `!spv.array<` integer-literal `x` element-type `>`
+// array-type ::= `!spv.array<` integer-literal `x` element-type
+// (`[` integer-literal `]`)? `>`
```
MaheshRavishankar: Nit : Please indent the expression to align with `!spv.array`

lib/Dialect/SPIRV/SPIRVDialect.cpp
```diff
@@ -53,6 +53,18 @@ SPIRVDialect::SPIRVDialect(MLIRContext *context)
 // Type Parsing
 //===----------------------------------------------------------------------===//
 
+// Forward declarations.
+template <typename ValTy>
+Optional<ValTy> parseAndVerify(SPIRVDialect const &dialect, Location loc,
+                               StringRef spec);
+template <>
+Optional<Type> parseAndVerify<Type>(SPIRVDialect const &dialect, Location loc,
```
MaheshRavishankar: I don't think we need to forward declare the template specialization.

lib/Dialect/SPIRV/SPIRVDialect.cpp
```diff
@@ -267,18 +307,16 @@ Optional<Type> parseAndVerify<Type>(SPIRVDialect const &dialect, Location loc,
...
-parseAndVerify(SPIRVDialect const &dialect, Location loc, StringRef spec) {
+Optional<uint64_t> parseAndVerify(SPIRVDialect const &dialect, Location loc,
+                                  StringRef spec) {
   uint64_t offsetVal = std::numeric_limits<uint64_t>::max();
   if (!spec.consume_front("[")) {
     emitError(loc, "expected '[' while parsing layout specification in '")
         << spec << "'";
     return llvm::None;
   }
   if (spec.consumeInteger(10, offsetVal)) {
```
MaheshRavishankar: This is not related to your change, but can you add spec.trim() before spec.consumeInteger?

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -207,6 +207,9 @@ class Deserializer {
   // Result <id> to decorations mapping.
   DenseMap<uint32_t, NamedAttributeList> decorations;
 
+  // Result <id> to type decorations.
+  DenseMap<uint32_t, uint64_t> typeDecorations;
```
MaheshRavishankar: I think it might be better to make this DenseMap<uint32_t, uint32_t> cause it will be interpreted to different types later.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -321,6 +324,9 @@ LogicalResult Deserializer::processDecoration(ArrayRef<uint32_t> words) {
         opBuilder.getIdentifier(attrName),
         opBuilder.getI32IntegerAttr(static_cast<int32_t>(words[2])));
     break;
+  case spirv::Decoration::ArrayStride:
+    typeDecorations[words[0]] = static_cast<uint64_t>(words[2]);
```
MaheshRavishankar: Could you add a check to see that words.size() == 3 (like it is done for the other decoration types.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -581,7 +587,15 @@ LogicalResult Deserializer::processArrayType(ArrayRef<uint32_t> operands) {
            << defOp->getName();
   }
 
-  typeMap[operands[0]] = spirv::ArrayType::get(elementTy, count);
+  spirv::ArrayType::LayoutInfo layoutInfo = 0;
```
MaheshRavishankar: You could just do `spirv::ArrayType::LayoutInfo layoutInfo = typeDecorations.lookup(operands[0])`

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -436,7 +461,15 @@ LogicalResult Serializer::processType(Location loc, Type type,
                                      operands))) ||
       succeeded(prepareBasicType(loc, type, typeEnum, operands))) {
     typeIDMap[type] = typeID;
-    return encodeInstructionInto(typesGlobalValues, typeEnum, operands);
+
+    if (failed(encodeInstructionInto(typesGlobalValues, typeEnum, operands))) {
+      return failure();
+    }
+
+    if (auto dType = type.dyn_cast<spirv::ArrayType>()) {
```
MaheshRavishankar: Why is this here? Shouldn't this be within `prepareBasicType` where it is checking for handling ArrayType?

lib/Dialect/SPIRV/SPIRVDialect.cpp
denis0x0D: I'll update it.

lib/Dialect/SPIRV/SPIRVDialect.cpp
denis0x0D: Thanks, I just didn't want to move the functions body, since I've used it in array parser, early then definitions. So, it it's ok to move them.

lib/Dialect/SPIRV/SPIRVDialect.cpp
denis0x0D: Thanks.

lib/Dialect/SPIRV/SPIRVDialect.cpp
MaheshRavishankar: I didn't mean move them. You can forward declare the template function. I dont think you need to forward declare the specialization. So keep 
```
template <typename ValTy>
Optional<ValTy> parseAndVerify(SPIRVDialect const &dialect, Location loc,
                               StringRef spec);
```

but remove the 

```
template <>
Optional<Type> parseAndVerify<Type>(SPIRVDialect const &dialect, Location loc, StringRef spec)
```
and 

```
template <>
Optional<uint64_t> parseAndVerify<uint64_t>(SPIRVDialect const &dialect, Location loc, StringRef spec);
```

I think that should work. If I am mistaken, I am sorry about that then.

denis0x0D: Oh, sorry didn't get it right at first. Seems like in this case we will get an error like specilization of "function", after instantiation.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: As far as I understood, please fix me if I'm wrong, to encode instruction with OpDecorate we need to call `encodeInstructionInto() `with last argument as array of args, when the first element of arg is unique id, so to process it in the prepareBasic function, we have to change the API prepareBasicType, so, if it's ok, I'll change it.
Thanks.

denis0x0D: Oh, I see the function prepareBasicType is private, so, if it's ok to I'll add one more args as resultId.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks for mention this!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, I'll update it.

denis0x0D:
@MaheshRavishankar  thanks a lot for review! I've updated the patch, can you please take a look?

lib/Dialect/SPIRV/SPIRVDialect.cpp
MaheshRavishankar: Oh Ok. Then sorry about that. This looks fine then.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
MaheshRavishankar: This looks good. Thanks!

denis0x0D:
@MaheshRavishankar thanks a lot for review!

lib/Dialect/SPIRV/SPIRVDialect.cpp
```diff
@@ -171,11 +184,38 @@ static Type parseArrayType(SPIRVDialect const &dialect, StringRef spec,
...
+
+  // Handle case when element type is not a trivial type
+  auto lastRDelimiter = spec.rfind('>');
+  if (lastRDelimiter != StringRef::npos) {
+    lastLSquare = spec.find('[', lastRDelimiter);
+  } else {
+    lastLSquare = spec.rfind('[');
+  }
+
+  // The check is relaxed in the parser
```
antiagainst: What does this comment mean? Can we change it to something more suitable?

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -581,7 +591,12 @@ LogicalResult Deserializer::processArrayType(ArrayRef<uint32_t> operands) {
            << defOp->getName();
   }
 
-  typeMap[operands[0]] = spirv::ArrayType::get(elementTy, count);
+  spirv::ArrayType::LayoutInfo layoutInfo = typeDecorations.lookup(operands[0]);
+  if (!layoutInfo) {
```
antiagainst: This is not true. Only types used by interface variables need to be explicitly laid out. Types used by function local variables are not. So we should drop this check.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -355,6 +359,27 @@ LogicalResult Serializer::processDecoration(Location loc, uint32_t resultID,
...
+    Location loc, spirv::ArrayType type, uint32_t resultID) {
+  if (type.hasLayout()) {
+    // OpDecorate %arrayTypeSSA ArrayStride strideLiteral
+    SmallVector<uint32_t, 3> args;
+    args.push_back(resultID);
+    args.push_back(static_cast<uint32_t>(spirv::Decoration::ArrayStride));
+    args.push_back(type.getArrayStride());
+    return encodeInstructionInto(decorations, spirv::Opcode::OpDecorate, args);
+  }
+  // Regarding to the spec each OpTypeArray must have ArrayStride
```
antiagainst: This is not true.

include/mlir/Dialect/SPIRV/SPIRVTypes.h
```diff
@@ -73,14 +73,22 @@ class ArrayType : public Type::TypeBase<ArrayType, CompositeType,
                                         detail::ArrayTypeStorage> {
 public:
   using Base::Base;
+  using LayoutInfo = uint64_t;
```
antiagainst: Could you put a comment here to mention that if this is 0 it means no layout info?

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Oh, sorry, my fault,
I was relying on this statement in the spec:

> **Each array type must have an ArrayStride decoration**, unless it is an array that contains a structure decorated with Block or BufferBlock, in which case it must not have an ArrayStride decoration.

and completely missed the header 

> Composite objects in the **StorageBuffer**, **Uniform**, and **PushConstant** Storage Classes must be explicitly laid out. The following apply to all the aggregate and matrix types describing such an object, recursively through their nested types:

Thanks a lot for fixing me!

lib/Dialect/SPIRV/SPIRVDialect.cpp
denis0x0D: Thanks, I've delete it.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Thanks, fixed!

include/mlir/Dialect/SPIRV/SPIRVTypes.h
denis0x0D: Thanks, updated.

denis0x0D:
@antiagainst thanks a lot for review! I've fixed my incorrect understanding of specifications for ArrayStride, can you please take a look?
Thanks.

denis0x0D:
tests updated.

denis0x0D:
@antiagainst @MaheshRavishankar if no one is interesting in the task https://github.com/tensorflow/mlir/issues/77 , can I take it?
Thanks!

antiagainst:
Sure! Please go ahead!

MaheshRavishankar:
Thanks Denis for taking up the extra tasks. We have been busy with trying to resolve the sticky issue with https://github.com/tensorflow/mlir/issues/68.

denis0x0D:
@antiagainst @MaheshRavishankar thanks, I'm happy to help with it. 
I've seen this issue https://github.com/tensorflow/mlir/issues/68 ,what the solution you've decided to take?

MaheshRavishankar:
@denis0x0D I think we are going to have to change the way module level variables and constants are represented. They will have a symbol name associated with such variables/constants and wont have an SSA value. I am in the process of changing variables to be this way. Will post an update when it goes into github. There are a lot of advantages of not doing implicit captures in functions. But it doesnt match directly with SPIR-V spec, so some things have to change in the current dialect.

antiagainst:
+1 to what @MaheshRavishankar said. We'll update #68 later to capture the offline discussion.

This pull request now has a conflict. Could you rebase against master, @denis0x0D? Thanks!

denis0x0D:
@MaheshRavishankar thanks, sounds interesting!
@antiagainst rebased on master.
Thanks.

denis0x0D:
BTW, it seems like CI build fails, because of hardcoded variable https://github.com/tensorflow/mlir/blob/master/CMakeLists.txt#L10
llvm was moved to C++14 recently and regarding to the https://gcc.gnu.org/projects/cxx-status.html#cxx14 gcc-5 must support c++14 standard.

lib/Dialect/SPIRV/SPIRVDialect.cpp
```diff
@@ -53,6 +53,18 @@ SPIRVDialect::SPIRVDialect(MLIRContext *context)
 // Type Parsing
 //===----------------------------------------------------------------------===//
 
+// Forward declarations.
+template <typename ValTy>
```
antiagainst: Could you forward-declare these as `static` to be consistent with later? I'm seeing compiler errors of "static declaration of 'parseAndVerify' follows non-static declaration".

denis0x0D: @antiagainst oh, sorry about this, I've updated it, this is indeed a bug regarding to the standard, but I cannot reproduce an error on my local machine with gcc 8, even with reduced test case for template functions.

antiagainst: No worries! And thanks for the ping regarding LLVM C++14 update! :)

PR #85, opened 2019-08-17
Add conversion for boolean attribute

schweitzpgi:


lib/Target/LLVMIR/ModuleTranslation.cpp
```diff
@@ -84,6 +84,8 @@ llvm::Constant *ModuleTranslation::getLLVMConstant(llvm::Type *llvmType,
     return llvm::ConstantFP::get(llvmType, floatAttr.getValue());
   if (auto funcAttr = attr.dyn_cast<SymbolRefAttr>())
     return functionMapping.lookup(funcAttr.getValue());
+  if (auto boolAttr = attr.dyn_cast<BoolAttr>())
```
River707: Can you add a test case to exercise this? (to test/Target/llvmir.mlir)

schweitzpgi: It appears that the problem we had may have been resolved another way.  We're still checking.

antiagainst:
Hey @schweitzpgi, are you still working on this PR? You will need to sign the CLA before we can accept the contribution. :)

schweitzpgi:
I think this was fixed a different way. It no longer seems to cause the same issue. Closing.

PR #87, opened 2019-08-17
Refactor / improve replaceAllMemRefUsesWith

bondhugula:
Refactor replaceAllMemRefUsesWith to split it into two methods: the new
method does the replacement on a single op, and is used by the existing
one.

- make the methods return LogicalResult instead of bool

- Earlier, when replacement failed (due to non-deferencing uses of the
  memref), the set of ops that had already been processed would have
  been replaced leaving the IR in an inconsistent state. Now, a
  pass is made over all ops to first check for non-deferencing
  uses, and then replacement is performed. No test cases were affected
  because all clients of this method were first checking for
  non-deferencing uses before calling this method (for other reasons).
  This isn't true for a use case in another upcoming PR (scalar
  replacement); clients can now bail out with consistent IR on failure
  of replaceAllMemRefUsesWith.

- multiple deferencing uses of the same memref in a single op is
  possible (we have no such use cases/scenarios), and this has always
  remained unsupported. Add an assertion for this.

### DIFF
--- a/include/mlir/Transforms/Utils.h
+++ b/include/mlir/Transforms/Utils.h
@@ -37,26 +37,26 @@ class AffineForOp;
 class Location;
 class OpBuilder;
 
-/// Replaces all "deferencing" uses of oldMemRef with newMemRef while optionally
-/// remapping the old memref's indices using the supplied affine map,
-/// 'indexRemap'. The new memref could be of a different shape or rank.
-/// 'extraIndices' provides additional access indices to be added to the start.
+/// Replaces all "dereferencing" uses of `oldMemRef` with `newMemRef` while
+/// optionally remapping the old memref's indices using the supplied affine map,
+/// `indexRemap`. The new memref could be of a different shape or rank.
+/// `extraIndices` provides additional access indices to be added to the start.
 ///
-/// 'indexRemap' remaps indices of the old memref access to a new set of indices
+/// `indexRemap` remaps indices of the old memref access to a new set of indices
 /// that are used to index the memref. Additional input operands to indexRemap
 /// can be optionally provided, and they are added at the start of its input
-/// list. 'indexRemap' is expected to have only dimensional inputs, and the
+/// list. `indexRemap` is expected to have only dimensional inputs, and the
 /// number of its inputs equal to extraOperands.size() plus rank of the memref.
 /// 'extraOperands' is an optional argument that corresponds to additional
 /// operands (inputs) for indexRemap at the beginning of its input list.
 ///
-/// 'domInstFilter', if non-null, restricts the replacement to only those
+/// `domInstFilter`, if non-null, restricts the replacement to only those
 /// operations that are dominated by the former; similarly, `postDomInstFilter`
 /// restricts replacement to only those operations that are postdominated by it.
 ///
 /// Returns true on success and false if the replacement is not possible,
-/// whenever a memref is used as an operand in a non-deferencing context, except
-/// for dealloc's on the memref which are left untouched. See comments at
+/// whenever a memref is used as an operand in a non-dereferencing context,
+/// except for dealloc's on the memref which are left untouched. See comments at
 /// function definition for an example.
 //
 //  Ex: to replace load %A[%i, %j] with load %Abuf[%t mod 2, %ii - %i, %j]:
@@ -66,12 +66,20 @@ class OpBuilder;
 //  extra operands, note that 'indexRemap' would just be applied to existing
 //  indices (%i, %j).
 //  TODO(bondhugula): allow extraIndices to be added at any position.
-bool replaceAllMemRefUsesWith(Value *oldMemRef, Value *newMemRef,
-                              ArrayRef<Value *> extraIndices = {},
-                              AffineMap indexRemap = AffineMap(),
-                              ArrayRef<Value *> extraOperands = {},
-                              Operation *domInstFilter = nullptr,
-                              Operation *postDomInstFilter = nullptr);
+LogicalResult replaceAllMemRefUsesWith(Value *oldMemRef, Value *newMemRef,
+                                       ArrayRef<Value *> extraIndices = {},
+                                       AffineMap indexRemap = AffineMap(),
+                                       ArrayRef<Value *> extraOperands = {},
+                                       Operation *domInstFilter = nullptr,
+                                       Operation *postDomInstFilter = nullptr);
+
+/// Performs the same replacement as the other version above but only for the
+/// dereferencing uses of `oldMemRef` in `op`.
+LogicalResult replaceAllMemRefUsesWith(Value *oldMemRef, Value *newMemRef,
+                                       Operation *op,
+                                       ArrayRef<Value *> extraIndices = {},
+                                       AffineMap indexRemap = AffineMap(),
+                                       ArrayRef<Value *> extraOperands = {});
 
 /// Creates and inserts into 'builder' a new AffineApplyOp, with the number of
 /// its results equal to the number of operands, as a composition
--- a/lib/Transforms/PipelineDataTransfer.cpp
+++ b/lib/Transforms/PipelineDataTransfer.cpp
@@ -115,13 +115,14 @@ static bool doubleBuffer(Value *oldMemRef, AffineForOp forOp) {
   auto ivModTwoOp = bInner.create<AffineApplyOp>(forOp.getLoc(), modTwoMap,
                                                  forOp.getInductionVar());
 
-  // replaceAllMemRefUsesWith will always succeed unless the forOp body has
-  // non-deferencing uses of the memref (dealloc's are fine though).
-  if (!replaceAllMemRefUsesWith(oldMemRef, newMemRef,
-                                /*extraIndices=*/{ivModTwoOp},
-                                /*indexRemap=*/AffineMap(),
-                                /*extraOperands=*/{},
-                                /*domInstFilter=*/&*forOp.getBody()->begin())) {
+  // replaceAllMemRefUsesWith will succeed unless the forOp body has
+  // non-dereferencing uses of the memref (dealloc's are fine though).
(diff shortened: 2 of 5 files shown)


include/mlir/Transforms/Utils.h
```diff
@@ -66,12 +66,20 @@ class OpBuilder;
...
-                              Operation *postDomInstFilter = nullptr);
+LogicalResult replaceAllMemRefUsesWith(Value *oldMemRef, Value *newMemRef,
+                                       ArrayRef<Value *> extraIndices = {},
+                                       AffineMap indexRemap = AffineMap(),
+                                       ArrayRef<Value *> extraOperands = {},
+                                       Operation *domInstFilter = nullptr,
+                                       Operation *postDomInstFilter = nullptr);
+
+/// Performs the same replacement as the other version above but only for the
+/// deferencing uses of `oldMemRef` in `op`.
```
joker-eph: I am not sure what  "deferencing uses of `oldMemRef` in `op`." means?

lib/Transforms/Utils/Utils.cpp
```diff
@@ -57,12 +57,178 @@ static NamedAttribute getAffineMapAttrForMemRef(Operation *op, Value *memref) {
...
+    assert(indexRemap.getNumResults() + extraIndices.size() == newMemRefRank);
+  } else {
+    assert(oldMemRefRank + extraIndices.size() == newMemRefRank);
+  }
+
+  // Assert same elemental type.
+  assert(oldMemRef->getType().cast<MemRefType>().getElementType() ==
+         newMemRef->getType().cast<MemRefType>().getElementType());
+
+  // Check if the memref was used in a non-deferencing context.
```
joker-eph: `deferencing` -> `dereferencing` (in many places in this PR)

Also the comment is literally paraphrasing what the next statement is, which is quite an anti-pattern.

include/mlir/Transforms/Utils.h
bondhugula: A dereferencing use is one where the memref is dereferenced, i.e., an individual element of it is accessed. For eg:

%v =  load %A[%i, %j] : memref<100 x 100 x f32> 

is a deferencing use of %A. affine.load, affine.store, load, store, dma_start, dma_finish are examples of ops that have deferencing uses of the memref. OTOH, uses in ops like:

%v = call "foo" (%A) : memref<100 x 100 x f32>
dealloc %A : memref<100 x 100 x f32>

are all non-deferencing uses of the memref. The doc comment is merely meant as a indirect reference to the version of the method above, which has all the info.

bondhugula: Sorry, just realized there was a typo! Thanks.

bondhugula:
> We should be able to expose such fix independently through tests: can one of the test passes be improved to expose this?

None of the current users of this method are actually affected by this fix, but there is another PR that I am about to submit on scalar replacement which relies on this (i.e., replaceAllMemRefUsesWith can fail and the failure is handled), and have a test case for that. However that PR is too big in itself and separate from this. 

The reason the existing users always make sure that this methods succeeds is that replaceAllMemRefUsesWith is typically called by a pass/utility after a bunch of IR actions have already been performed (new op creation, etc.). So, if it were to fail, one would have to undo/clear up all of those actions. As a result, those clients always first check from the outside a-priori whether it'd succeed (by checking for non-dereferencing uses) and then call it.

lib/Transforms/Utils/Utils.cpp
bondhugula: Fixed.

include/mlir/Transforms/Utils.h
bondhugula: Fixed.

bondhugula:
Just found a way to test this - PipelineDataTransfer does check for and handle the failure. Added a test case - with this change, the IR remains correct after the replacement failure. 

Thanks!

antiagainst:
Hey @joker-eph or @River707, could you take another look over this PR?

test/Transforms/pipeline-data-transfer.mlir
```diff
@@ -52,7 +52,7 @@ func @loop_nest_dma() {
 // CHECK-NEXT:  %{{.*}} = affine.load %{{.*}}[%{{.*}} mod 2, %{{.*}}] : memref<2x32xf32, 1>
 // CHECK-NEXT:  %{{.*}} = "compute"(%{{.*}}) : (f32) -> f32
 // CHECK-NEXT:  affine.store %{{.*}}, %{{.*}}[%{{.*}} mod 2, %{{.*}}] : memref<2x32xf32, 1>
-// CHECK-NEXT:  affine.for %{{.*}} = 0 to 128 {
+// CHECK-NEXT:  affine.for %{{.*}} = 0 to 32 {
```
joker-eph: Why these changes 128->32?

bondhugula: This one is minor and orthogonal. The DMA buffer size / transfer size in the op's above is actually 32, and so the loop should be of trip count 32 to be meaningful. It doesn't affect this test case - but just fixed it while on this file.

lib/Transforms/PipelineDataTransfer.cpp
```diff
@@ -117,11 +117,12 @@ static bool doubleBuffer(Value *oldMemRef, AffineForOp forOp) {
 
   // replaceAllMemRefUsesWith will always succeed unless the forOp body has
   // non-deferencing uses of the memref (dealloc's are fine though).
```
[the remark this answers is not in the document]
bondhugula: Fixed, thanks.

bondhugula:
Anything more on this?

lib/Transforms/Utils/Utils.cpp
```diff
@@ -89,170 +254,44 @@ bool mlir::replaceAllMemRefUsesWith(Value *oldMemRef, Value *newMemRef,
...
+    if (postDomInstFilter && !postDomInfo->postDominates(postDomInstFilter, op))
       continue;
 
-    // Skip dealloc's - no replacement is necessary, and a replacement doesn't
-    // hurt dealloc's.
-    if (isa<DeallocOp>(opInst))
+    // Skip dealloc's - no replacement is necessary, and a memref replacement
+    // at other uses doesn't hurt these dealloc's.
+    if (isa<DeallocOp>(op))
       continue;
```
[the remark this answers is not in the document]
bondhugula: Looks like this PR has already been committed. I'll address all of these coverage issues in a follow-up.

lib/Transforms/Utils/Utils.cpp
```diff
@@ -57,12 +57,177 @@ static NamedAttribute getAffineMapAttrForMemRef(Operation *op, Value *memref) {
...
+  }
+
+  // Assert same elemental type.
+  assert(oldMemRef->getType().cast<MemRefType>().getElementType() ==
+         newMemRef->getType().cast<MemRefType>().getElementType());
+
+  if (!isMemRefDereferencingOp(*op))
+    // Failure: memref used in a non-dereferencing context (potentially
+    // escapes); no replacement in these cases.
+    return failure();
```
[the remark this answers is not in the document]
bondhugula: This one does have test coverage - from lib/Transforms/PipelineDataTransfer.cpp and tested via:
https://github.com/tensorflow/mlir/blob/c340184d48a9650d6606d193cce4fdff127e821e/test/Transforms/pipeline-data-transfer.mlir#L220

lib/Transforms/Utils/Utils.cpp
```diff
@@ -57,12 +57,177 @@ static NamedAttribute getAffineMapAttrForMemRef(Operation *op, Value *memref) {
...
+
+  SmallVector<unsigned, 2> usePositions;
+  for (const auto &opEntry : llvm::enumerate(op->getOperands())) {
+    if (opEntry.value() == oldMemRef)
+      usePositions.push_back(opEntry.index());
+  }
+
+  // If memref doesn't appear, nothing to do.
+  if (usePositions.empty())
+    return success();
```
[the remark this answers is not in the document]
bondhugula: The current user of this method (only one) doesn't need/cover this case, but there is an upcoming scalar replacement utility for memref's that uses this method directly. Putting both in the same PR would have made it too big.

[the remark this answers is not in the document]
bondhugula: The current user of this method (only one) doesn't need/cover this case, but there is a scalar replacement utility for memref's that I plan to submit soon that uses this method directly. Putting both in the same PR would have made it too big.

lib/Transforms/Utils/Utils.cpp
```diff
@@ -89,170 +254,44 @@ bool mlir::replaceAllMemRefUsesWith(Value *oldMemRef, Value *newMemRef,
...
+    // that has the use, and that op could be postDomFilter or domFilter itself!
+    opsToReplace.insert(op);
   }
 
-  for (auto *opInst : opsToErase)
-    opInst->erase();
+  for (auto *op : opsToReplace) {
+    if (failed(replaceAllMemRefUsesWith(oldMemRef, newMemRef, op, extraIndices,
+                                        indexRemap, extraOperands)))
+      assert(false && "memref replacement guaranteed to succeed here");
```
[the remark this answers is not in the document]
bondhugula: Yes, will change this in a follow-up PR.

test/Transforms/pipeline-data-transfer.mlir
bondhugula: Sure. (I was under the impression that cleanup/NFC in the same file was fine.)

lib/Transforms/Utils/Utils.cpp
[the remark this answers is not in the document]
bondhugula: You are right. I confused this one with the same check in the other version of this method. This will be covered only when another PR that actually uses this method directly (scalar replacement) is submitted.

PR #90, opened 2019-08-19
Add vim scripts for indent/syntax

bondhugula:
- some of it has been adapted from LLVM's vim utils


bondhugula:
vim_mlir

bondhugula:
> Nice, could you add a screenshot too of the changes this results into for the syntax highlighting?
> 

You mean the comment above? It is a screenshot! :-)

It looks like some of the highlighting (elemental types in aggregate ones) isn't working. I'll do another round of update.

bondhugula:
> > Nice, could you add a screenshot too of the changes this results into for the syntax highlighting?
> 
> You mean the comment above? It is a screenshot! :-)
> 
> It looks like some of the highlighting (elemental types in aggregate ones) isn't working. I'll do another round of update.

Done. Screenshot inline below. 

> I'm unsure about the indent file, that seems to be encoding a convention (for example, re { }). Does 
> the indenting rules match what the printer would generate?

I tried auto-indenting on printer generated stuff (==) on a bunch of MLIR (mostly 'for' nests), and it all matches (nothing changed).

vim_mlir2

bondhugula:
This appears to be stuck at "internal safe review approval" (this PR makes no changes to source code). I have more updates to the syntax file - should I wait for this to be committed or force update this?

PR #94, opened 2019-08-19
[spirv] Support (de)serialization of spv.struct

denis0x0D:
This patch is addresing issue https://github.com/tensorflow/mlir/issues/77
and supports (de)serialization of spv.struct with offset decorations.

Thanks!

### DIFF
--- a/lib/Dialect/SPIRV/Serialization/Deserializer.cpp
+++ b/lib/Dialect/SPIRV/Serialization/Deserializer.cpp
@@ -28,6 +28,7 @@
 #include "mlir/IR/Location.h"
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Support/StringExtras.h"
+#include "llvm/ADT/Sequence.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/bit.h"
 
@@ -84,6 +85,9 @@ private:
   /// Method to process an OpDecorate instruction.
   LogicalResult processDecoration(ArrayRef<uint32_t> words);
 
+  // Method to process an OpMemberDecorate instruction.
+  LogicalResult processMemberDecoration(ArrayRef<uint32_t> words);
+
   /// Processes the SPIR-V function at the current `offset` into `binary`.
   /// The operands to the OpFunction instruction is passed in as ``operands`.
   /// This method processes each instruction inside the function and dispatches
@@ -121,6 +125,8 @@ private:
   LogicalResult processArrayType(ArrayRef<uint32_t> operands);
 
   LogicalResult processFunctionType(ArrayRef<uint32_t> operands);
+
+  LogicalResult processStructType(ArrayRef<uint32_t> operands);
 
   //===--------------------------------------------------------------------===//
   // Constant
@@ -231,6 +237,9 @@ private:
 
   // Result <id> to type decorations.
   DenseMap<uint32_t, uint32_t> typeDecorations;
+
+  // Result <id> to member decorations.
+  DenseMap<uint32_t, DenseMap<uint32_t, uint32_t>> memberDecorationMap;
 
   // List of instructions that are processed in a defered fashion (after an
   // initial processing of the entire binary). Some operations like
@@ -364,6 +373,23 @@ LogicalResult Deserializer::processDecoration(ArrayRef<uint32_t> words) {
     break;
   default:
     return emitError(unknownLoc, "unhandled Decoration : '") << decorationName;
+  }
+  return success();
+}
+
+LogicalResult Deserializer::processMemberDecoration(ArrayRef<uint32_t> words) {
+  // The binary layout of OpMemberDecorate is different comparing to OpDecorate
+  if (words.size() != 4) {
+    return emitError(unknownLoc, "OpMemberDecorate must have 4 operands");
+  }
+
+  switch (static_cast<spirv::Decoration>(words[2])) {
+  case spirv::Decoration::Offset:
+    memberDecorationMap[words[0]][words[1]] = words[3];
+    break;
+  default:
+    return emitError(unknownLoc, "unhandled OpMemberDecoration case: ")
+           << words[2];
   }
   return success();
 }
@@ -653,6 +679,8 @@ LogicalResult Deserializer::processType(spirv::Opcode opcode,
     return processArrayType(operands);
   case spirv::Opcode::OpTypeFunction:
     return processFunctionType(operands);
+  case spirv::Opcode::OpTypeStruct:
+    return processStructType(operands);
   default:
     return emitError(unknownLoc, "unhandled type instruction");
   }
@@ -719,6 +747,46 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
     returnTypes = llvm::makeArrayRef(returnType);
   }
   typeMap[operands[0]] = FunctionType::get(argTypes, returnTypes, context);
+  return success();
+}
+
+LogicalResult Deserializer::processStructType(ArrayRef<uint32_t> operands) {
+  // TODO(ravishankarm) : Regarding to the spec spv.struct must support zero
+  // amount of members.
+  if (operands.size() < 2) {
+    return emitError(unknownLoc, "OpTypeStruct must have at least 2 operand");
+  }
+
+  SmallVector<Type, 0> memberTypes;
+  for (auto op : llvm::drop_begin(operands, 1)) {
+    Type memberType = getType(op);
+    if (!memberType) {
+      return emitError(unknownLoc, "OpTypeStruct references undefined <id> ")
+             << op;
+    }
+    memberTypes.push_back(memberType);
+  }
+
+  SmallVector<spirv::StructType::LayoutInfo, 0> layoutInfo;
+  // Check for layoutinfo
+  auto memberDecorationIt = memberDecorationMap.find(operands[0]);
+  if (memberDecorationIt != memberDecorationMap.end()) {
+    // Each member must have an offset
+    const auto &offsetDecorationMap = memberDecorationIt->second;
+    auto offsetDecorationMapEnd = offsetDecorationMap.end();
+    for (auto memberIndex : llvm::seq<uint32_t>(0, memberTypes.size())) {
+      // Check that specific member has an offset
+      auto offsetIt = offsetDecorationMap.find(memberIndex);
+      if (offsetIt == offsetDecorationMapEnd) {
+        return emitError(unknownLoc, "OpTypeStruct with <id> ")
+               << operands[0] << " must have an offset for " << memberIndex
+               << "-th member";
+      }
+      layoutInfo.push_back(
+          static_cast<spirv::StructType::LayoutInfo>(offsetIt->second));
+    }
+  }
+  typeMap[operands[0]] = spirv::StructType::get(memberTypes, layoutInfo);
   return success();
 }
 
@@ -993,6 +1061,7 @@ LogicalResult Deserializer::processInstruction(spirv::Opcode opcode,
   case spirv::Opcode::OpTypeVector:
   case spirv::Opcode::OpTypeArray:
   case spirv::Opcode::OpTypeFunction:
+  case spirv::Opcode::OpTypeStruct:
   case spirv::Opcode::OpTypePointer:
     return processType(opcode, operands);
   case spirv::Opcode::OpConstant:
@@ -1015,6 +1084,8 @@ LogicalResult Deserializer::processInstruction(spirv::Opcode opcode,
     return processConstantNull(operands);
   case spirv::Opcode::OpDecorate:
     return processDecoration(operands);
+  case spirv::Opcode::OpMemberDecorate:
+    return processMemberDecoration(operands);
   case spirv::Opcode::OpFunction:
     return processFunction(operands);
   default:
--- a/lib/Dialect/SPIRV/Serialization/Serializer.cpp
+++ b/lib/Dialect/SPIRV/Serialization/Serializer.cpp
@@ -28,6 +28,7 @@
 #include "mlir/IR/Builders.h"
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Support/StringExtras.h"
+#include "llvm/ADT/Sequence.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/bit.h"
 #include "llvm/Support/raw_ostream.h"
@@ -147,6 +148,11 @@ private:
                                       uint32_t resultId) {
     return emitError(loc, "unhandled decoraion for type:") << type;
   }
+
+  /// Process member decoration
+  LogicalResult processMemberDecoration(uint32_t structID, uint32_t memberNum,
+                                        spirv::Decoration decorationType,
+                                        uint32_t value);
 
   //===--------------------------------------------------------------------===//
   // Types
@@ -411,6 +417,16 @@ LogicalResult Serializer::processTypeDecoration<spirv::ArrayType>(
   }
   return success();
 }
+
+LogicalResult
+Serializer::processMemberDecoration(uint32_t structID, uint32_t memberIndex,
+                                    spirv::Decoration decorationType,
+                                    uint32_t value) {
+  SmallVector<uint32_t, 4> args(
+      {structID, memberIndex, static_cast<uint32_t>(decorationType), value});
+  return encodeInstructionInto(decorations, spirv::Opcode::OpMemberDecorate,
+                               args);
+}
 } // namespace
 
 LogicalResult Serializer::processFuncOp(FuncOp op) {
@@ -615,6 +631,31 @@ Serializer::prepareBasicType(Location loc, Type type, uint32_t resultID,
     typeEnum = spirv::Opcode::OpTypePointer;
     operands.push_back(static_cast<uint32_t>(ptrType.getStorageClass()));
     operands.push_back(pointeeTypeID);
+    return success();
+  }
+
+  if (auto structType = type.dyn_cast<spirv::StructType>()) {
+    bool hasLayout = structType.hasLayout();
+    for (auto elementIndex :
+         llvm::seq<uint32_t>(0, structType.getNumElements())) {
+      uint32_t elementTypeID = 0;
+      if (failed(processType(loc, structType.getElementType(elementIndex),
+                             elementTypeID))) {
+        return failure();
+      }
+      operands.push_back(elementTypeID);
+      if (hasLayout) {
+        // Decorate each struct member with an offset
+        if (failed(processMemberDecoration(
+                resultID, elementIndex, spirv::Decoration::Offset,
+                static_cast<uint32_t>(structType.getOffset(elementIndex))))) {
+          return emitError(loc, "cannot decorate ")
+                 << elementIndex << "-th member of : " << structType
+                 << "with its offset";
+        }
+      }
+    }
+    typeEnum = spirv::Opcode::OpTypeStruct;
     return success();
   }
 
--- a/test/Dialect/SPIRV/Serialization/struct.mlir
+++ b/test/Dialect/SPIRV/Serialization/struct.mlir
@@ -1,0 +1,24 @@
+// RUN: mlir-translate -serialize-spirv %s | mlir-translate -deserialize-spirv | FileCheck %s
+
+func @spirvmodule() -> () {
+  spv.module "Logical" "VulkanKHR" {
+    // CHECK: !spv.ptr<!spv.struct<!spv.array<128 x f32 [4]> [0]>, Input>
+    spv.globalVariable @var0 bind(0, 1) : !spv.ptr<!spv.struct<!spv.array<128 x f32 [4]> [0]>, Input>
+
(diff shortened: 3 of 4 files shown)


lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -368,6 +376,23 @@ LogicalResult Deserializer::processDecoration(ArrayRef<uint32_t> words) {
...
+  if (words.size() != 4) {
+    return emitError(unknownLoc, "OpMemberDecorate must have 4 operands");
+  }
+
+  switch (static_cast<spirv::Decoration>(words[2])) {
+  case spirv::Decoration::Offset:
+    memberDecorationMap[words[0]][words[1]] = words[3];
+    break;
+  default:
+    return emitError(unknownLoc, "unhandled OpMemberDecoration with op code: '")
```
antiagainst: This is not opcode. What about "unhandled OpMemberDecoration case: "?

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +749,46 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
   return success();
 }
 
+LogicalResult Deserializer::processStructType(ArrayRef<uint32_t> operands) {
+  if (operands.size() < 2) {
+    return emitError(unknownLoc, "OpTypeStruct must have at least one member");
```
antiagainst: Actually the spec explicitly says "Declare a new structure type: an aggregate of zero or more potentially heterogeneous members." So we can have zero members.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +749,46 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
...
+    const auto &offsetDecorationMap = memberDecorationIterator->second;
+    for (uint32_t memberNum = 0; memberNum < memberTypes.size(); ++memberNum) {
+      // Check that specific member has an offset
+      auto offsetIterator = offsetDecorationMap.find(memberNum);
+      if (offsetIterator != offsetDecorationMap.end()) {
+        layoutInfo.push_back(
+            static_cast<spirv::StructType::LayoutInfo>(offsetIterator->second));
+      } else {
+        return emitError(unknownLoc, "OpTypeStruct with <id> ")
+               << operands[0] << " must have an offset for member with number "
```
antiagainst: "... for " << memberNum << "-th member" ?

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -411,6 +416,19 @@ LogicalResult Serializer::processTypeDecoration<spirv::ArrayType>(
   }
   return success();
 }
+
+LogicalResult
+Serializer::processMemberDecoration(uint32_t structID, uint32_t memberNum,
+                                    spirv::Decoration decorationType,
+                                    uint32_t value) {
+  SmallVector<uint32_t, 4> args;
```
antiagainst: I think we can just do `... args({structID, ...})` to initialize directly?

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -411,6 +416,19 @@ LogicalResult Serializer::processTypeDecoration<spirv::ArrayType>(
   }
   return success();
 }
+
+LogicalResult
+Serializer::processMemberDecoration(uint32_t structID, uint32_t memberNum,
```
antiagainst: Can we name it as `memberIndex`? `memberNum` always makes me think we are passing in the total count here.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -618,6 +636,29 @@ Serializer::prepareBasicType(Location loc, Type type, uint32_t resultID,
     return success();
   }
 
+  if (auto structType = type.dyn_cast<spirv::StructType>()) {
+    for (uint32_t elementNum = 0; elementNum < structType.getNumElements();
```
antiagainst: Same here. elementIndex?

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -618,6 +636,29 @@ Serializer::prepareBasicType(Location loc, Type type, uint32_t resultID,
...
+                             elementTypeID))) {
+        return failure();
+      }
+      operands.push_back(elementTypeID);
+      if (structType.hasLayout()) {
+        // Decorate each struct member with an offset
+        if (failed(processMemberDecoration(
+                resultID, elementNum, spirv::Decoration::Offset,
+                static_cast<uint32_t>(structType.getOffset(elementNum))))) {
+          return emitError(loc, "cannot decorate member: ")
```
antiagainst: "cannot decorate " << elementNum << "-th member of " << structType << " with its offset" ?

test/Dialect/SPIRV/Serialization/struct.mlir
```diff
@@ -0,0 +1,23 @@
...
+    // CHECK: !spv.ptr<!spv.struct<f32 [0], !spv.struct<f32 [0], !spv.array<16 x f32 [4]> [4]> [4]>, Input>
+    spv.globalVariable !spv.ptr<!spv.struct<f32 [0], !spv.struct<f32 [0], !spv.array<16 x f32 [4]> [4]> [4]>, Input> @var1 bind (0, 2)
+
+    // CHECK: !spv.ptr<!spv.struct<f32 [0], i32 [4], f64 [8], i64 [16], f32 [24], i32 [30], f32 [34], i32 [38]>, StorageBuffer>
+    spv.globalVariable !spv.ptr<!spv.struct<f32 [0], i32 [4], f64 [8], i64 [16], f32 [24], i32 [30], f32 [34], i32 [38]>, StorageBuffer> @var2
+
+    // CHECK: !spv.ptr<!spv.struct<!spv.array<128 x !spv.struct<!spv.array<128 x f32 [4]> [0]> [4]> [0]>, StorageBuffer> @var3
+    spv.globalVariable !spv.ptr<!spv.struct<!spv.array<128 x !spv.struct<!spv.array<128 x f32 [4]> [0]> [4]> [0]>, StorageBuffer> @var3
+
+    // CHECK: %{{.*}}: !spv.ptr<!spv.struct<!spv.array<128 x f32 [4]> [0]>, Input>, %{{.*}} !spv.ptr<!spv.struct<!spv.array<128 x f32 [4]> [0]>, Output>
```
antiagainst: Nit: can use `// CHECK-SAME` to break into two lines.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +749,46 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
   return success();
 }
 
+LogicalResult Deserializer::processStructType(ArrayRef<uint32_t> operands) {
+  if (operands.size() < 2) {
+    return emitError(unknownLoc, "OpTypeStruct must have at least one member");
+  }
+
+  SmallVector<Type, 0> memberTypes;
+  for (uint32_t opNum = 1; opNum < operands.size(); ++opNum) {
```
antiagainst: I think we can do `for (auto member : operands)` here?

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, I'll fix it.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks for mention this, but it seems like the current spv.struct must have at least one field, sorry if I'm missing something.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, seems like I can fix it like this:
`for (auto member :make_range(operands.begin() + 1, operands.end())`

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks!

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Thanks, I'll update it!

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Thanks, I'll fix it!

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Thanks!

test/Dialect/SPIRV/Serialization/struct.mlir
denis0x0D: Thanks for mention this!

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Thanks!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
River707: llvm::drop_begin(operands, 1)

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +749,46 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
...
+    memberTypes.push_back(memberType);
+  }
+
+  // Check for layoutinfo
+  auto memberDecorationIterator = memberDecorationMap.find(operands[0]);
+  if (memberDecorationIterator != memberDecorationMap.end()) {
+    // Each member must have an offset
+    SmallVector<spirv::StructType::LayoutInfo, 0> layoutInfo;
+    const auto &offsetDecorationMap = memberDecorationIterator->second;
+    for (uint32_t memberNum = 0; memberNum < memberTypes.size(); ++memberNum) {
```
River707: Please follow LLVM style to avoid recomputing the end of a loop,
https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +749,46 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
...
+      // Check that specific member has an offset
+      auto offsetIterator = offsetDecorationMap.find(memberNum);
+      if (offsetIterator != offsetDecorationMap.end()) {
+        layoutInfo.push_back(
+            static_cast<spirv::StructType::LayoutInfo>(offsetIterator->second));
```
River707: Prefer early exit, swap the the compare so that you can return early:

if (...) {
  return ...
}

lib/Dialect/SPIRV/Serialization/Serializer.cpp
River707: Same as above, don't recompute the end of the loop.

test/Dialect/SPIRV/Serialization/struct.mlir
River707: I don't think {{.*}} is useful when starting a CHECK, you can just drop it.

River707:
BTW, 'sequence' refers to llvm::seq:
https://github.com/llvm-mirror/llvm/blob/master/include/llvm/ADT/Sequence.h#L75

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: My fault, thanks for mention this!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Oh, thanks!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, I'll update it!

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Thanks.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
MaheshRavishankar: Maybe that is the case. Have to look into it.

denis0x0D:
@MaheshRavishankar @antiagainst @River707 thanks a lot for review!
I've updated almost on every comment, but still have some concerns about:
1. processStructType in Deserializer, seems like current spv.struct can not handle construction with zero amount of fields.
2. Iteration all over DenseMap<uint32_t, uint32_t> which maps memberIndex to its offset, in this specific case the order is critical for creating layoutinfo.
Please fix me if I'm wrong.

Can you please take a look again?
Thanks!

denis0x0D:
rebased on master.

MaheshRavishankar:
1) If current spv.struct doesnt support construction with zero amount of fields. Thats a bug. Please file one and add a `TODO(ravishankarm)` for me to take a look at that later.
2) My bad. Maybe you can use DenseMap<uint32_t, std::map<uint32_t, uint32_t>> to get it ordered. Anyway, I am not sure how the ordering changes the comment that I think relates to this. WIll continue on that comment.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
MaheshRavishankar: If you change `memberDecorationMap` to `DenseMap<uint32_t, std::map<uint32_t, uint32_t>>`, you could use the same code as

```
SmallVector<spirv::StructType::LayoutInfo, 0> layoutInfo;
for (auto &memberDecoration : memberDecorationMap) {
   for (auto &memberIndexIterator : llvm::enumerate(memberDecoration)) {
      if (memberIndexIterator.index() != memberIndexIterator.value().second) {
            return emitError(...);
      }
      layoutInfo.push_back(static_cast<spirv::StructType::LayoutInfo>(memberIndexIterator.value().second));
   }
}
```
Does that not work?

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +750,47 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
...
+    if (!memberType) {
+      return emitError(unknownLoc, "OpTypeStruct references undefined <id> ")
+             << op;
+    }
+    memberTypes.push_back(memberType);
+  }
+
+  SmallVector<spirv::StructType::LayoutInfo, 0> layoutInfo;
+  // Check for layoutinfo
+  auto memberDecorationIterator = memberDecorationMap.find(operands[0]);
```
antiagainst: We can just use `It` as a suffix for these iterators to save a few characters. It's well-known abbr.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +750,47 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
   return success();
 }
 
+LogicalResult Deserializer::processStructType(ArrayRef<uint32_t> operands) {
+  if (operands.size() < 2) {
```
antiagainst: Please add a TODO here as said by Mahesh.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
antiagainst: I think we prefer `llvm::DenseMap` instead of `std::map`. IIUC, `std::map` always does allocations so it's inferior to `llvm::DenseMap` here. Denis' current impl looks fine to me.

MaheshRavishankar: Yes, it is a Nit. I am fine with it as well.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, fixed.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, updated!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +750,49 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
...
+
+  SmallVector<spirv::StructType::LayoutInfo, 0> layoutInfo;
+  // Check for layoutinfo
+  auto memberDecorationIt = memberDecorationMap.find(operands[0]);
+  if (memberDecorationIt != memberDecorationMap.end()) {
+    // Each member must have an offset
+    const auto &offsetDecorationMap = memberDecorationIt->second;
+    auto offsetDecorationMapEnd = offsetDecorationMap.end();
+    for (auto memberIndex :
+         llvm::seq(static_cast<uint32_t>(0),
```
River707: Why not just llvm::seq<uint32_t>(...)

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -722,6 +750,49 @@ LogicalResult Deserializer::processFunctionType(ArrayRef<uint32_t> operands) {
...
+                   static_cast<uint32_t>(memberTypes.size()))) {
+      // Check that specific member has an offset
+      auto offsetIt = offsetDecorationMap.find(memberIndex);
+      if (offsetIt == offsetDecorationMapEnd) {
+        return emitError(unknownLoc, "OpTypeStruct with <id> ")
+               << operands[0] << " must have an offset for " << memberIndex
+               << "-th member";
+      } else {
```
River707: Drop the else here.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -618,6 +634,32 @@ Serializer::prepareBasicType(Location loc, Type type, uint32_t resultID,
     return success();
   }
 
+  if (auto structType = type.dyn_cast<spirv::StructType>()) {
+    bool hasLayout = structType.hasLayout();
+    for (auto elementIndex :
+         llvm::seq(static_cast<uint32_t>(0),
```
River707: Same here.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, fixed!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, updated!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: @MaheshRavishankar I've missed that comment somehow, sorry about this, let me check it! Thanks!

denis0x0D: @MaheshRavishankar llvm::enumarate does not work for DenseMap, as far as I understand it:
```
‘class llvm::detail::enumerator_iter<llvm::detail::DenseMapPair<unsigned int, llvm::DenseMap<unsigned int, unsigned int, llvm::DenseMapInfo<unsigned int>, llvm::detail::DenseMapPair<unsigned int, unsigned int> > >&>’:
/home/denis/llvm-project/llvm/projects/mlir/lib/Dialect/SPIRV/Serialization/Deserializer.cpp:772:70:   required from here
/home/denis/llvm-project/llvm/include/llvm/ADT/STLExtras.h:52:40: error: no matching function for call to ‘begin(llvm::detail::DenseMapPair<unsigned int, llvm::DenseMap<unsigned int, unsigned int, llvm::DenseMapInfo<unsigned int>, llvm::detail::DenseMapPair<unsigned int, unsigned int> > >&)’
 using IterOfRange = decltype(std::begin(std::declval<RangeT &>()));
```
BTW, the order here is needed, because we processing OpTypeStruct %field0, ... %fieldN in order.
Thanks.

denis0x0D:
@MaheshRavishankar @antiagainst @River707 updated regarding to the latest comments.

antiagainst:
@denis0x0D, yup, thanks! This will land soon internally.

denis0x0D:
@antiagainst thanks!

PR #95, opened 2019-08-20
Add mechanism to dump JIT-compiled objects to files

dcaballe:
This commit introduces the bits to be able to dump JIT-compile
objects to external files by passing an object cache to OrcJit.
The new functionality is tested in mlir-cpu-runner under the flag
`dump-object-file`.

I followed instructions provided here: https://groups.google.com/a/tensorflow.org/forum/?utm_medium=email&utm_source=footer#!msg/mlir/QrmgnlYq1x8/L9s_w6jbDwAJ


joker-eph:
As we discussed in the past, most of the JIT functionality in MLIR really does not have much specific to MLIR and could be refactored into LLVM itself.

The JIT code has been responsible for most build breakage recently (LLVM getting out-of-sync with MLIR), so seeing how it is already a maintenance burden I am reluctant to take more of it rather than investing into moving code upstream first as much as possible.

include/mlir/ExecutionEngine/ExecutionEngine.h
```diff
@@ -32,6 +33,8 @@
 namespace llvm {
 template <typename T> class Expected;
 class Module;
+
+class ObjectCache;
```
ftynse: This doesn't look necessary, the relevant header is included anyway.

include/mlir/ExecutionEngine/ExecutionEngine.h
```diff
@@ -42,6 +45,28 @@ namespace impl {
...
+/// We only use it to dump JIT-compiled code to object files (see
+/// `dumpObjectFile`).
+class MLIRObjectCache : public llvm::ObjectCache {
+public:
+  void notifyObjectCompiled(const llvm::Module *M,
+                            llvm::MemoryBufferRef ObjBuffer) override;
+
+  std::unique_ptr<llvm::MemoryBuffer>
+  getObject(const llvm::Module *M) override {
+    // Not implemented.
```
ftynse: Is this function ever called in normal operation?  If not, I'd rather put llvm_unreachable("unimplemented");

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -371,3 +376,26 @@ llvm::Error ExecutionEngine::invoke(StringRef name,
...
+  // Set up the output file.
+  std::string errorMessage;
+  auto file = openOutputFile(outputFilename, &errorMessage);
+  if (!file) {
+    llvm::errs() << errorMessage << "\n";
+    return;
+  }
+
+  // Dump the object generated for a single module to the output file. 
+  assert(cachedObjects.size() == 1 && "Expected only one object entry.");
```
ftynse: The execution engine API does not preclude from reusing the same object cache for different engines/runs without cleaning it.  Since the cache has a module id-based map, just use the module id to retrieve the correct buffer.

lib/Support/JitRunner.cpp
```diff
@@ -166,6 +171,35 @@ static LogicalResult convertAffineStandardToLLVMIR(ModuleOp module) {
...
+  if (!expectedEngine)
+    return expectedEngine.takeError();
+
+  auto engine = std::move(*expectedEngine);
+  auto expectedFPtr = engine->lookup(entryPoint);
+  if (!expectedFPtr)
+    return expectedFPtr.takeError();
+
+  if (dumpObjectFile)
+    objectCache->dumpToObjectFile(inputFilename + ".o");
```
ftynse: It would be great to have the name configurable

ftynse:
> As we discussed in the past, most of the JIT functionality in MLIR really does not have much specific to MLIR and should be refactored into LLVM itself.

In the meantime, this significantly simplifies life for those developers who work on the compilation and low-level debugging.

joker-eph:
> > As we discussed in the past, most of the JIT functionality in MLIR really does not have much specific to MLIR and should be refactored into LLVM itself.
> 
> In the meantime, this significantly simplifies life for those developers who work on the compilation and low-level debugging.

The "developers who work on the compilation and low-level debugging" should be motivated to do the refactoring then :)

It's been multiple months that this is pending, it seems that unless there is incentive no-one will do it. 
Also the more code / feature we add here the more difficult it is to upstream as you'd have to split it up in multiple patches to integrate it in LLVM.

ftynse:
> The "developers who work on the compilation and low-level debugging" should be motivated to do the refactoring then :)

It's a matter of time, not motivation.  And this code saves time, which could potentially be used for a refactoring with long review delays and uncertain outcome.

joker-eph:
I'm OK with this, as long as someone commits to start working on moving all this code out of MLIR (moving all of this to LLVM is pending since May)

ftynse:
Given your experience with both mlir and llvm, and that you seem to have a clear idea of how to propose this to llvm, maybe you would want to take a lead.

joker-eph:
I could if no one else want to, but then my way of handling it is to not take any new feature into this code until it is upstreamed.

dcaballe:
> I could if no one else want to, but then my way of handling it is to not take any new feature into this code until it is upstreamed.

Hopefully we can find a middle ground :). We should be able work on the refactoring while adding some new features. Blocking progress for months(?) until all the code is refactored would be a big deal, IMO.

As I mentioned, other than `MLIRObjectCache` code, changes in this PR are mostly about passing a pointer across the existing interfaces. That shouldn't be a problem for refactoring or maintenance. I'll follow up and try to move `MLIRObjectCache` to LLVM once we have the use case for it in MLIR. I might be able to help refactor some other bits base on the cycles I have available. In fact, I already committed some code to OrcJit that was needed for MLIR: https://reviews.llvm.org/D65760

Hopefully, this makes sense to you.
Thanks,
Diego

joker-eph:
> > I could if no one else want to, but then my way of handling it is to not take any new feature into this code until it is upstreamed.
> 
> Hopefully we can find a middle ground :). We should be able work on the refactoring while adding some new features.

I don't think this is any different from my previous proposal? Whoever needs the feature *now* should also be able to commit time to work on the refactoring/upstreaming.

> Blocking progress for months(?) until all the code is refactored would be a big deal, IMO.

I think it should be closer to days than months, at least until we allow to feature creep ;)

> ... changes in this PR are mostly about ...

Every single PR in isolation is always justified and harmless. The pitfalls of "features creep" in general is in the general trajectory :)

nicolasvasilache:
5b018e7afeadc4ed49bd457e65f56b0b09a0e6fa makes some changes that should go in a mutually agreeable direction.

@dcaballe, it should be easy for you to rebase your changes on top of this commit.

For now I just made the SimpleObjectCache owned by the ExecutionEngine, we can revisit later.

One thing I am not completely clear is what to do with this output (haven't tried your code):
  1. just call objdump on the .o ?
  2. call llvm-dis | llc -O3
If it is the former then all is good alread , if it is the latter, I would very much love to have a way to dump the actual assembly.

Hope this helps move the ball!

bondhugula:
> I could if no one else want to, but then my way of handling it is to not take any new feature into this code until it is upstreamed.

From the discussion above, I see the tradeoff between maintenance and the need to make experimental progress. But the small relative impact this line of changes has on build maintenance does not outweigh its great productivity / progress benefits to early developers. So I'm with @ftynse and @dcaballe here. :-)

dcaballe:
> [5b018e7](https://github.com/tensorflow/mlir/commit/5b018e7afeadc4ed49bd457e65f56b0b09a0e6fa) makes some changes that should go in a mutually agreeable direction.

That's great. Thanks! I'll rebase the PR into this new scheme and try to see if there is something else we can refactor.

> One thing I am not completely clear is what to do with this output (haven't tried your code):
> 
> 1. just call objdump on the .o ?
> 2. call llvm-dis | llc -O3
>    If it is the former then all is good alread 

Yes, objdump works just fine.

lib/Support/JitRunner.cpp
dcaballe: Added a flag for that

lib/ExecutionEngine/ExecutionEngine.cpp
dcaballe: The problem is that the cache requires an `llvm::Module` and we call this function from clients that don't have access to it (e.g., JitRunner). Only `mlir::ModuleOp`. If there is no other way to retrieve the `llvm::Module` generated for a given `mlir::ModuleOp`, I could introduce a map `mlir::ModuleOp` -> `llvm::Module` in `SimpleObjectCache`. However, I'm not sure if this would be a problem in an scenario where we generate multiple `llvm::Module`s for a single `mlir::ModuleOp`.

include/mlir/ExecutionEngine/ExecutionEngine.h
```diff
@@ -48,30 +49,11 @@ class SimpleObjectCache : public llvm::ObjectCache {
...
-    // Not implemented.
-    return nullptr;
-  }
-
   /// Dump cached object to output file `filename`.
   void dumpToObjectFile(llvm::StringRef filename);
 
 private:
-  llvm::StringMap<std::unique_ptr<llvm::MemoryBuffer>> cachedObjects;
+  llvm::StringMap<std::unique_ptr<llvm::MemoryBuffer>> CachedObjects;
```
nicolasvasilache: we use lowerCamelCase in MLIR, if/when this goes to LLVM this can change then

nicolasvasilache: we use lowerCamelCase in MLIR, if/when this goes to LLVM this can change then

nicolasvasilache:
Shooting from the hip here, you could try to use the same sort of prefix as
for the lib files in the mlir-cpu-runner test with Linalg. Maybe we can
write where we build ?

joker-eph:
Jacques fixed it using `%t` / `%T` which indicates to lit that you intend to use a temporary file.

dcaballe:
Great. Thanks. I had no idea about that.

PR #98, opened 2019-08-22
fix loop unroll and jam - operand mapping - imperfect nest case

bondhugula:
- fix operand mapping while cloning sub-blocks to jam - was incorrect
  for imperfect nests where def/use was across sub-blocks
- strengthen/generalize the first test case to cover the previously
  missed scenario
- clean up the other cases while on this.

Previously, unroll-jamming the following nest
```
    affine.for %arg0 = 0 to 2048 {
      %0 = alloc() : memref<512x10xf32>
      affine.for %arg1 = 0 to 10 {
        %1 = affine.load %0[%arg0, %arg1] : memref<512x10xf32>
      }
      dealloc %0 : memref<512x10xf32>
    }
```

would yield

```
      %0 = alloc() : memref<512x10xf32>
      %1 = affine.apply #map0(%arg0)
      %2 = alloc() : memref<512x10xf32>
      affine.for %arg1 = 0 to 10 {
        %4 = affine.load %0[%arg0, %arg1] : memref<512x10xf32>
        %5 = affine.apply #map0(%arg0)
        %6 = affine.load %0[%5, %arg1] : memref<512x10xf32>
      }
      dealloc %0 : memref<512x10xf32>
      %3 = affine.apply #map0(%arg0)
      dealloc %0 : memref<512x10xf32>

```

instead of

```

module {
    affine.for %arg0 = 0 to 2048 step 2 {
      %0 = alloc() : memref<512x10xf32>
      %1 = affine.apply #map0(%arg0)
      %2 = alloc() : memref<512x10xf32>
      affine.for %arg1 = 0 to 10 {
        %4 = affine.load %0[%arg0, %arg1] : memref<512x10xf32>
        %5 = affine.apply #map0(%arg0)
        %6 = affine.load %2[%5, %arg1] : memref<512x10xf32>
      }
      dealloc %0 : memref<512x10xf32>
      %3 = affine.apply #map0(%arg0)
      dealloc %2 : memref<512x10xf32>
    }
```

### DIFF
--- a/test/Transforms/unroll-jam.mlir
+++ b/test/Transforms/unroll-jam.mlir
@@ -7,48 +7,52 @@
 
 // CHECK-LABEL: func @unroll_jam_imperfect_nest() {
 func @unroll_jam_imperfect_nest() {
-  // CHECK: %c100 = constant 100 : index
-  // CHECK-NEXT: affine.for %arg0 = 0 to 100 step 2 {
   affine.for %i = 0 to 101 {
-    // CHECK: "addi32"(%arg0, %arg0) : (index, index) -> i32
-    // CHECK-NEXT: %3 = affine.apply [[MAP_PLUS_1]](%arg0)
-    // CHECK-NEXT: "addi32"(%3, %3) : (index, index) -> i32
     %x = "addi32"(%i, %i) : (index, index) -> i32
     affine.for %j = 0 to 17 {
-      // CHECK:      %8 = "addi32"(%arg0, %arg0) : (index, index) -> i32
-      // CHECK-NEXT: "addi32"(%8, %8) : (i32, i32) -> i32
-      // CHECK-NEXT: %10 = affine.apply [[MAP_PLUS_1]](%arg0)
-      // CHECK-NEXT: %11 = "addi32"(%10, %10) : (index, index) -> i32
-      // CHECK-NEXT: "addi32"(%11, %11) : (i32, i32) -> i32
       %y = "addi32"(%i, %i) : (index, index) -> i32
       %z = "addi32"(%y, %y) : (i32, i32) -> i32
     }
-    // CHECK: "addi32"(%arg0, %arg0) : (index, index) -> i32
-    // CHECK-NEXT: %6 = affine.apply [[MAP_PLUS_1]](%arg0)
-    // CHECK-NEXT: "addi32"(%6, %6) : (index, index) -> i32
-    %w = "addi32"(%i, %i) : (index, index) -> i32
-  } // CHECK }
-  // cleanup loop (single iteration)
-  // CHECK: "addi32"(%c100, %c100) : (index, index) -> i32
-  // CHECK-NEXT: affine.for %arg0 = 0 to 17 {
-  // CHECK-NEXT:   %2 = "addi32"(%c100, %c100) : (index, index) -> i32
-  // CHECK-NEXT:   "addi32"(%2, %2) : (i32, i32) -> i32
-  // CHECK-NEXT: }
-  // CHECK-NEXT: "addi32"(%c100, %c100) : (index, index) -> i32
+    %w = "foo"(%i, %x) : (index, i32) -> i32
+  }
   return
 }
+// CHECK:      affine.for [[IV0:%arg[0-9]+]] = 0 to 100 step 2 {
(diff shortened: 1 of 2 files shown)


test/Transforms/unroll-jam.mlir
```diff
@@ -7,43 +7,46 @@
...
-  // CHECK-NEXT: affine.for %arg0 = 0 to 17 {
-  // CHECK-NEXT:   %2 = "addi32"(%c100, %c100) : (index, index) -> i32
-  // CHECK-NEXT:   "addi32"(%2, %2) : (i32, i32) -> i32
-  // CHECK-NEXT: }
-  // CHECK-NEXT: "addi32"(%c100, %c100) : (index, index) -> i32
+    %w = "foo"(%i, %x) : (index, i32) -> i32
+  }
   return
 }
+// CHECK:      affine.for %arg0 = 0 to 100 step 2 {
```
River707: Can you fixup this test to avoid hard coding SSA value names?

bondhugula: Done. I've replaced similar hard code IVs and block arguments at other places, but there isn't apparently a way to define variables in CHECK-LABEL's -- as a result, block arguments had to be replaced with generic patterns and the test cases aren't as strong now. And for IV definitions, I couldn't avoid hardcoding the prefix "arg", i.e., 

for [[IV0:%arg[0-9]+]]

since

"for [[IV0:%{{.*}}]]"   doesn't appear to match.

test/Transforms/unroll-jam.mlir
```diff
@@ -7,48 +7,51 @@
...
+// CHECK:      %{{.*}} = "addi32"(%c100, %c100) : (index, index) -> i32
+// CHECK-NEXT: affine.for [[IV0]] = 0 to 17 {
+// CHECK-NEXT:   [[RESC:%[0-9]+]] = "addi32"(%c100, %c100) : (index, index) -> i32
+// CHECK-NEXT:   "addi32"([[RESC]], [[RESC]]) : (i32, i32) -> i32
+// CHECK-NEXT: }
+// CHECK-NEXT: %{{.*}} = "foo"(%c100, %{{.*}}) : (index, i32) -> i32
+// CHECK-NEXT: return
 
-// CHECK-LABEL: func @loop_nest_unknown_count_1(%arg0: index) {
+// CHECK-LABEL: func @loop_nest_unknown_count_1(%{{.*}}: index) {
```
joker-eph: The trick to capture in this case is:

```
// CHECK-LABEL: func @loop_nest_unknown_count_1
// CHECK-SAME: [[ARG0:%{{.*}}]]: index

File: test/Transforms/unroll-jam.mlir [hunk 3]
```diff
@@ -7,48 +7,51 @@
...
+// CHECK-NEXT:     [[INC1:%[0-9]+]] = affine.apply [[MAP_PLUS_1]]([[IV0]])
+// CHECK-NEXT:     [[RES4:%[0-9]+]] = "addi32"([[INC1]], [[INC1]]) : (index, index) -> i32
+// CHECK-NEXT:     "addi32"([[RES4]], [[RES4]]) : (i32, i32) -> i32
+// CHECK-NEXT:   }
+// CHECK:        "foo"([[IV0]], [[RES1]]) : (index, i32) -> i32
+// CHECK-NEXT:   {{.*}} = affine.apply [[MAP_PLUS_1]]([[IV0]])
+// CHECK-NEXT:   "foo"({{.*}}, [[RES2]]) : (index, i32) -> i32
+// CHECK:      }
+// Cleanup loop (single iteration).
+// CHECK:      %{{.*}} = "addi32"(%c100, %c100) : (index, index) -> i32
```
joker-eph: If you're not capturing the LHS, you can just leave it out?
(similarly the type may not be needed)

bondhugula: Dropped the type. I retained the LHS for readability - it's easier to identify/match these when comparing with the actual output.

test/Transforms/unroll-jam.mlir
bondhugula: Thanks very much! I tried something similar with CHECK (instead of CHECK-SAME), and that obviously didn't work!  
Fixed now.

PR #101, opened 2019-08-25
Update Ch.2 of the Toy tutorial.

River707:
Update Ch.2 of the Toy tutorial.

The code and documentation for this chapter of the tutorial have been updated after changes to the MLIR code base as well as to follow the new flow of the tutorial. The toy 'array' type has been replaced by usages of the MLIR tensor type. The code has also been cleaned up and modernized.

### DIFF
--- a/g3doc/Tutorials/Toy/Ch-2.md
+++ b/g3doc/Tutorials/Toy/Ch-2.md
@@ -23,30 +23,21 @@ addresses this issue by being designed for extensibility. As such, there are
 little to no pre-defined instructions (*operations* in MLIR terminology) or
 types.
 
-## MLIR Dialects, Regions, Blocks, and Operations
+## MLIR Dialects and Operations
 
 [Language reference](../../LangRef.md#dialects)
 
 In MLIR, the core unit of abstraction and computation is an `Operation`, similar
-in many ways to LLVM instructions. Operations may also hold a number
-[`Region`s](../../LangRef.md#regions), of which contain a list of
-[`Blocks`](../../LangRef.md#blocks) forming a CFG(Control Flow Graph). `Blocks`
-are similar to LLVM `BasicBlocks` in that they contain a sequential list of
-`Operations`. Operations can be used to represent all of the core IR structures
-in LLVM: instructions, globals(like functions), modules, etc; however MLIR does
-not have a closed set of operations. Instead, the MLIR operation set is fully
-extensible and operations can have application-specific semantics.
+in many ways to LLVM instructions. Operations can be used to represent all of
+the core IR structures in LLVM: instructions, globals(like functions), modules,
+etc; however MLIR does not have a closed set of operations. Instead, the MLIR
+operation set is fully extensible and operations can have application-specific
+semantics.
 
 MLIR supports this extensibility with the concept of
 [Dialects](../../LangRef.md#dialects). Among other things, Dialects provide a
 grouping mechanism for operations under a unique `namespace`. Dialects will be a
 discussed a bit more in the [next chapter](Ch-3.md).
-
-The top level structure in MLIR (like in LLVM) is a
-[Module](../../LangRef.md#module) (equivalent to a translation unit in C/C++). A
-module is an operation that contains a single region composed of a single block.
-This block may contain any nature of operations, including
-[functions](../../LangRef.md#functions).
 
 Here is the MLIR assembly for the Toy 'transpose' operations:
 
@@ -228,3 +219,4 @@ it should take an operand, and it shouldn't return any values.
 
 In the [next chapter](Ch-3.md) we will register our dialect and operations with
 MLIR, plug into the verifier, and add nicer APIs to manipulate our operations.
+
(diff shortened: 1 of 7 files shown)


g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -63,88 +78,89 @@ in MLIR the location is a core requirement which translates in APIs manipulating
...
 Here MLIR will enforce some structural constraints (SSA, block termination,
-return operand type coherent with function return type, etc.) but otherwise the
-types and the operation are completely opaque.
+etc.) but otherwise the types and the `custom.operation` are completely opaque.
 
-We will take advantage of this facility to emit MLIR for Toy by traversing the
-AST. Our types will be prefixed with "!toy" and our operation name with "toy.".
-MLIR refers to this prefix as a *dialect*, we will introduce this with more
-details in the [next chapter](Ch-3.md).
+We will take advantage of this facility for the initial emission of MLIR for Toy
```
joker-eph: I thought something you wanted was to include a more prominent warning about the fact that this is only for demo purpose, and pointing that the next chapter will introduce the more common way of creating IR? (or something like that).

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -8,52 +8,67 @@ to compile Toy.
...
+addresses this issue by being designed for extensibility. As such, there are
+little to no pre-defined instructions (*operations* in MLIR terminology) or
+types.
+
+## MLIR Dialects, Regions, Blocks, and Operations
+
+[Language reference](../../LangRef.md#dialects)
+
+In MLIR, the core unit of abstraction and computation is an `Operation`, similar
+in many ways to LLVM instructions. Operations may also hold a number
```
sana-damani: Typo here: "a number of regions"

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -8,52 +8,67 @@ to compile Toy.
...
+[`Region`s](../../LangRef.md#regions), of which contain a list of
+[`Blocks`](../../LangRef.md#blocks) forming a CFG(Control Flow Graph). `Blocks`
+are similar to LLVM `BasicBlocks` in that they contain a sequential list of
+`Operations`. Operations can be used to represent all of the core IR structures
+in LLVM: instructions, globals(like functions), modules, etc; however MLIR does
+not have a closed set of operations. Instead, the MLIR operation set is fully
```
sana-damani: Should we mention standard ops and types available in mlir (such as tensor) here?

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -8,52 +8,67 @@ to compile Toy.
...
+extensible and operations can have application-specific semantics.
+
+MLIR supports this extensibility with the concept of
+[Dialects](../../LangRef.md#dialects). Among other things, Dialects provide a
+grouping mechanism for operations under a unique `namespace`. Dialects will be a
+discussed a bit more in the [next chapter](Ch-3.md).
+
+The top level structure in MLIR (like in LLVM) is a
+[Module](../../LangRef.md#module) (equivalent to a translation unit in C/C++). A
+module is an operation that contains a single region composed of a single block.
```
sana-damani: If module is the top level structure and only allows for a single basic block, how do we get control flow?

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -8,52 +8,67 @@ to compile Toy.
...
+This block may contain any nature of operations, including
+[functions](../../LangRef.md#functions).
```
sana-damani: This section doesn't explain why we have regions/blocks etc leaving that to LangRef, but also does not appear to reference these terms later in the IR. Do we really need this section here? I think that discussing what a dialect is in detail before generating the toy IR may be more useful.

g3doc/Tutorials/Toy/Ch-2.md
sana-damani: I also like the idea of introducing IR registration first but then mentioning that the IR will roundtrip without registration but won't do any checking, without this being a part of the actual tutorial source.

g3doc/Tutorials/Toy/Ch-2.md
River707: I don't understand what you mean here? The module operation does not allow control flow within its body.

g3doc/Tutorials/Toy/Ch-2.md
River707: Standard ops here isn't important, but mentioning that we will reuse the standard tensor type is.

g3doc/Tutorials/Toy/Ch-2.md
River707: There needs to be some introduction to the structure, but not all of it is necessary up front. Trimmed it a bit.

g3doc/Tutorials/Toy/Ch-2.md
River707: I'm not sure of the exact flow of Ch.2 and Ch.3 at this point and they are likely to be merged. At this point I'm just updating/modernizing the chapter. I'm leaving the wording for this warning to a followup.

PR #102, opened 2019-08-25
[spirv] Add Block decoration for spv.struct.

denis0x0D:
This patch is addressing the second part of the issue https://github.com/tensorflow/mlir/issues/77

1. Adds Block decoration for `spv.struct` that are used as resource variable.
Decorates `spv.struct` for` spv.globalVariable` with type `!spv.ptr<!spv.struct<...>, storageClass>`, where storageClass is `Uniform `or `StorageBuffer`.
2. Adds unittest to check the Block decoration.

@antiagainst @MaheshRavishankar sorry for delay, can you please take a look?
Thanks!

### DIFF
--- a/lib/Dialect/SPIRV/Serialization/Serializer.cpp
+++ b/lib/Dialect/SPIRV/Serialization/Serializer.cpp
@@ -174,6 +174,10 @@ private:
 
   bool isVoidType(Type type) const { return type.isa<NoneType>(); }
 
+  /// Returns true if the given type is a pointer type to a struct in Uniform or
+  /// StorageBuffer storage class.
+  bool isInterfaceStructPtrType(Type type) const;
+
   /// Main dispatch method for serializing a type. The result <id> of the
   /// serialized type will be returned as `typeID`.
   LogicalResult processType(Location loc, Type type, uint32_t &typeID);
@@ -558,6 +562,22 @@ Serializer::processGlobalVariableOp(spirv::GlobalVariableOp varOp) {
   if (failed(processType(varOp.getLoc(), varOp.type(), resultTypeID))) {
     return failure();
   }
+
+  if (isInterfaceStructPtrType(varOp.type())) {
+    auto structType = varOp.type()
+                          .cast<spirv::PointerType>()
+                          .getPointeeType()
+                          .cast<spirv::StructType>();
+    SmallVector<uint32_t, 2> args{
+        findTypeID(structType),
+        static_cast<uint32_t>(spirv::Decoration::Block)};
+    if (failed(encodeInstructionInto(decorations, spirv::Opcode::OpDecorate,
+                                     args))) {
+      return varOp.emitError("cannot decorate ")
+             << structType << " with Block decoration";
+    }
+  }
+
   elidedAttrs.push_back("type");
   SmallVector<uint32_t, 4> operands;
   operands.push_back(resultTypeID);
@@ -608,6 +628,17 @@ Serializer::processGlobalVariableOp(spirv::GlobalVariableOp varOp) {
 //===----------------------------------------------------------------------===//
 // Type
 //===----------------------------------------------------------------------===//
+
+bool Serializer::isInterfaceStructPtrType(Type type) const {
+  if (auto ptrType = type.dyn_cast<spirv::PointerType>()) {
+    auto storageClass = ptrType.getStorageClass();
+    if (storageClass == spirv::StorageClass::Uniform ||
+        storageClass == spirv::StorageClass::StorageBuffer) {
+      return ptrType.getPointeeType().isa<spirv::StructType>();
+    }
+  }
+  return false;
+}
 
 LogicalResult Serializer::processType(Location loc, Type type,
                                       uint32_t &typeID) {
--- a/unittests/Dialect/SPIRV/SerializationTest.cpp
+++ b/unittests/Dialect/SPIRV/SerializationTest.cpp
@@ -1,0 +1,124 @@
+//===- SerializationTest.cpp - SPIR-V Seserialization Tests -------------===//
+//
+// Copyright 2019 The MLIR Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -558,6 +569,21 @@ Serializer::processGlobalVariableOp(spirv::GlobalVariableOp varOp) {
...
+    auto structType = varOp.type()
+                          .cast<spirv::PointerType>()
+                          .getPointeeType()
+                          .cast<spirv::StructType>();
+    SmallVector<uint32_t, 2> args{
+        findTypeID(structType),
+        static_cast<uint32_t>(spirv::Decoration::Block)};
+    if (failed(encodeInstructionInto(decorations, spirv::Opcode::OpDecorate,
+                                     args))) {
+      return emitError(varOp.getLoc(), "cannot decorate ") << structType;
```
River707: Do not use the global emitError if you have an op instance, use the 'op.emitError' methods instead:
varOp.emitError

unittests/Dialect/SPIRV/SerializationTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+  addGlobalVar(opBuilder, "var1", structNested16StructType,
+               spirv::StorageClass::StorageBuffer);
+  addGlobalVar(opBuilder, "var2", Array16IntType, spirv::StorageClass::Uniform);
+  addGlobalVar(opBuilder, "var3", Int32Type,
+               spirv::StorageClass::StorageBuffer);
+  addGlobalVar(opBuilder, "var4", structArray32IntType,
+               spirv::StorageClass::Input);
+
+  // Serialize module.
+  ASSERT_FALSE(failed(spirv::serialize(module, binary)));
```
River707: I'm not a huge fan of the way this test is structured; what is this actually testing? It seems like in these serializer tests you have to re-create parts of the deserializer, which requires a bunch of additional logic, verification and potential bugs that seem unrelated to the feature being tested. Is there a way to restructure this, hopefully without using c++ unit tests that still achieves the same goal?

denis0x0D: @River707 thanks for review, the purpose of this test to is check the some parts of the binary form which does not affect IR, I agree that looks bad; the discussion of the unittest of  binary form was here https://github.com/tensorflow/mlir/issues/77 , but we didn't discuss the realization, it might be better to collect Block decorated types in "deserilizer", but I'm not sure about it.
Here is a simple example of Block decoration:
`OpDecorate %_struct_6 Block`
```

; SPIR-V
; Version: 1.3
; Generator: Khronos SPIR-V Tools Assembler; 0
; Bound: 20
; Schema: 0
               OpMemoryModel Logical GLSL450
               OpName %var0 "var0"
               OpMemberDecorate %_struct_6 0 Offset 0
               OpDecorate %_struct_6 Block
        %int = OpTypeInt 32 1
     %int_16 = OpConstant %int 16
%_arr_int_int_16 = OpTypeArray %int %int_16
  %_struct_6 = OpTypeStruct %_arr_int_int_16
       %var0 = OpVariable %_ptr_Uniform__struct_6 Uniform
```


`%_struct_6` is a `<id> `and we have to deduce its type step by step:

` %_struct_6 = OpTypeStruct %_arr_int_int_16` 
`%_arr_int_int_16 = OpTypeArray %int %int_16` 
` %int_16 = OpConstant %int 16`
`%int = OpTypeInt 32 1`

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Thanks, fixed.

unittests/Dialect/SPIRV/SerializationTest.cpp
denis0x0D: @antiagainst thanks for detailed explanation, fixed regarding to you comment.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -174,6 +174,9 @@ class Serializer {
 
   bool isVoidType(Type type) const { return type.isa<NoneType>(); }
 
+  /// Method that checks if a type is suitable for a resource variable.
```
bondhugula: Nit:

/// Returns true if the type is suitable for a resource variable.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -174,6 +174,17 @@ class Serializer {
 
   bool isVoidType(Type type) const { return type.isa<NoneType>(); }
 
+  bool isStorageBufferType(Type type) const {
```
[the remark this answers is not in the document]
denis0x0D: Thanks, fixed.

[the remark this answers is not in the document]
denis0x0D: Thanks, fixed.

unittests/Dialect/SPIRV/SerializationTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/Sequence.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringRef.h"
+#include "gmock/gmock.h"
+
+using namespace mlir;
+
+// SPIR-V dialect registration.
+static mlir::DialectRegistration<mlir::spirv::SPIRVDialect> spirvDialect;
```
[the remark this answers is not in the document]
denis0x0D: Probably I'm missing some linker flags for CMake, but I'm getting an assert, that spirv dialect was not registered, while creating a spirv type, without explicit registration.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Thanks! Fixed.

unittests/Dialect/SPIRV/SerializationTest.cpp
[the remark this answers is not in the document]
denis0x0D: Ok, got it, I'l try to figure out why it does not work.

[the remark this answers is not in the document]
denis0x0D: Thanks a lot for mention this!

denis0x0D:
@antiagainst thanks a lot for review! I've updated the patch regarding to your comments, can you please take a look?

denis0x0D:
@antiagainst updated module creation part, seems like it's an invalid to create operation by Builder created from context, not a Region. I didn't see the assertion, because made latest updation on my laptop with llvm/mlir builded "MinRelSize", sorry about it.
Thanks.

PR #106, opened 2019-08-28
pipeline-data-transfer: remove dead tag alloc's and improve test cove…

bondhugula:
…rage for replaceMemRefUsesWith

- address remaining comments from PR #87 for better test coverage for
  pipeline-data-transfer (add dealloc's for test cases)
- remove dead tag allocs the same way they are removed for the replaced buffers

### DIFF
--- a/lib/Transforms/PipelineDataTransfer.cpp
+++ b/lib/Transforms/PipelineDataTransfer.cpp
@@ -294,7 +294,7 @@ void PipelineDataTransfer::runOnAffineForOp(AffineForOp forOp) {
       } else if (oldMemRef->hasOneUse()) {
         if (auto dealloc = dyn_cast<DeallocOp>(*oldMemRef->user_begin())) {
           dealloc.erase();
-          oldMemRef->getDefiningOp()->erase();
+          allocInst->erase();
         }
       }
     }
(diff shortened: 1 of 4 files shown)


bondhugula:
@River707 @joker-eph  - this I hope addresses the remaining comments for PR #87

lib/Transforms/PipelineDataTransfer.cpp
```diff
@@ -309,11 +309,18 @@ void PipelineDataTransfer::runOnAffineForOp(AffineForOp forOp) {
       LLVM_DEBUG(llvm::dbgs() << "tag double buffering failed\n";);
```
joker-eph: Out-of-scope for this diff, but FYI there is no coverage in this branch.

lib/Transforms/PipelineDataTransfer.cpp
```diff
@@ -309,11 +309,18 @@ void PipelineDataTransfer::runOnAffineForOp(AffineForOp forOp) {
...
-    // If the old tag has no more uses, remove its 'dead' alloc if it was
-    // alloc'ed.
-    if (oldTagMemRef->use_empty())
-      if (auto *allocInst = oldTagMemRef->getDefiningOp())
-        allocInst->erase();
+    // If the old tag has no uses or a single dealloc use, remove it.
+    // (canonicalization handles more complex cases).
+    if (auto *tagAllocInst = oldTagMemRef->getDefiningOp()) {
+      if (oldTagMemRef->use_empty()) {
+        tagAllocInst->erase();
```
joker-eph: We don't see coverage for this branch at the moment apparently.

bondhugula: All test cases have a matching dealloc for the tag alloc, and so this was never reached. Removed a tag dealloc from a test case - this is covered now.

lib/Transforms/PipelineDataTransfer.cpp
bondhugula: This will be reached if the tag associated with the DMA transfer escapes in between (almost no use case in practice) - so missed a test case here, but I can add one in this PR; it's very much related to the other changes.

bondhugula: Done.

PR #108, opened 2019-08-28
Merge Ch3 of the Toy tutorial into chapter 2.

River707:
This effectively rewrites Ch.2 to introduce dialects, operations, and registration instead of deferring to Ch.3. This allows for introducing the best practices up front(using ODS, registering operations, etc.), and limits the opaque API to the chapter document instead of the code.

### DIFF
--- a/include/mlir/Dialect/SPIRV/SPIRVControlFlowOps.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVControlFlowOps.td
@@ -64,6 +64,10 @@ def SPV_BranchOp : SPV_Op<"Branch", [Terminator]> {
 
   let skipDefaultBuilders = 1;
 
+  let extraClassDeclaration = [{
+    Block *getTarget() { return getOperation()->getSuccessor(0); }
+  }];
+
   let autogenSerialization = 0;
 }
 
@@ -114,13 +118,19 @@ def SPV_BranchConditionalOp : SPV_Op<"BranchConditional", [Terminator]> {
 
   let builders = [
     OpBuilder<
-      "Builder *, OperationState *state, Value *condition, "
-      "Block *trueBranch, Block *falseBranch, /*optional*/ArrayAttr weights",
+      "Builder *builder, OperationState *state, Value *condition, "
+      "Block *trueBranch, Block *falseBranch, "
+      "Optional<std::pair<uint32_t, uint32_t>> weights",
       [{
         state->addOperands(condition);
         state->addSuccessor(trueBranch, {});
         state->addSuccessor(falseBranch, {});
-        state->addAttribute("branch_weights", weights);
+        if (weights) {
+          auto attr =
+              builder->getI32ArrayAttr({static_cast<int32_t>(weights->first),
+                                        static_cast<int32_t>(weights->second)});
+          state->addAttribute("branch_weights", attr);
+        }
       }]
     >
   ];
@@ -132,6 +142,10 @@ def SPV_BranchConditionalOp : SPV_Op<"BranchConditional", [Terminator]> {
   let extraClassDeclaration = [{
     // Branch indices into the successor list.
     enum { kTrueIndex = 0, kFalseIndex = 1 };
+
+    Block *getTrueBlock() { return getOperation()->getSuccessor(kTrueIndex); }
+
+    Block *getFalseBlock() { return getOperation()->getSuccessor(kFalseIndex); }
   }];
 }
 
@@ -173,6 +187,21 @@ def SPV_LoopOp : SPV_Op<"loop"> {
   let results = (outs);
 
   let regions = (region AnyRegion:$body);
+
+  let extraClassDeclaration = [{
+    // Returns the loop header block.
+    Block *getHeaderBlock();
+
+    // Returns the loop continue block.
+    Block *getContinueBlock();
+
+    // Returns the loop merge block.
+    Block *getMergeBlock();
+
+    // Adds an empty entry block and loop merge block containing one
+    // spv._merge op.
+    void addEntryAndMergeBlock();
+  }];
 
   let hasOpcode = 0;
 }
--- a/include/mlir/IR/Function.h
+++ b/include/mlir/IR/Function.h
@@ -104,6 +104,10 @@ public:
   /// returned.
   Block *addEntryBlock();
 
+  /// Add a normal block to the end of the function's block list. The function
+  /// should at least already have an entry block.
+  Block *addBlock();
+
 private:
   // This trait needs access to `getNumFuncArguments` and `verifyType` hooks
   // defined below.
--- a/lib/Dialect/SPIRV/SPIRVOps.cpp
+++ b/lib/Dialect/SPIRV/SPIRVOps.cpp
@@ -591,9 +591,9 @@ static void print(spirv::BranchConditionalOp branchOp, OpAsmPrinter *printer) {
 
   if (auto weights = branchOp.branch_weights()) {
     *printer << " [";
-    mlir::interleaveComma(
-        weights->getValue(), printer->getStream(),
-        [&](Attribute a) { *printer << a.cast<IntegerAttr>().getInt(); });
+    interleaveComma(weights->getValue(), *printer, [&](Attribute a) {
+      *printer << a.cast<IntegerAttr>().getInt();
+    });
     *printer << "]";
   }
 
@@ -860,8 +860,7 @@ static void print(spirv::EntryPointOp entryPointOp, OpAsmPrinter *printer) {
            << entryPointOp.fn();
   if (auto interface = entryPointOp.interface()) {
     *printer << ", ";
-    mlir::interleaveComma(interface.getValue().getValue(), printer->getStream(),
-                          [&](Attribute a) { printer->printAttribute(a); });
+    interleaveComma(interface.getValue().getValue(), *printer);
   }
 }
 
@@ -908,8 +907,8 @@ static void print(spirv::ExecutionModeOp execModeOp, OpAsmPrinter *printer) {
     return;
   }
   *printer << ", ";
-  mlir::interleaveComma(
-      values.getValue().cast<ArrayAttr>(), printer->getStream(),
+  interleaveComma(
+      values.getValue().cast<ArrayAttr>(), *printer,
       [&](Attribute a) { *printer << a.cast<IntegerAttr>().getInt(); });
 }
 
@@ -1083,7 +1082,7 @@ static inline bool isMergeBlock(Block &block) {
 /// given `dstBlock`.
 static inline bool hasOneBranchOpTo(Block &srcBlock, Block &dstBlock) {
   // Check that there is only one op in the `srcBlock`.
-  if (std::next(srcBlock.begin()) != srcBlock.end())
+  if (srcBlock.empty() || std::next(srcBlock.begin()) != srcBlock.end())
     return false;
 
   auto branchOp = dyn_cast<spirv::BranchOp>(srcBlock.back());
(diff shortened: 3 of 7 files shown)


sana-damani:
Maybe just what a dialect contains? That is, custom operations, types and
transformations? That gives insight into what's coming next when defining
the Toy dialect.

joker-eph:
> If ODS is the preferred method, is there any reason to introduce the non-TableGen approach first? In fact, could we not skip it altogether?

We discussed it I think, the reason I chose to introduce the C++ class originally is for simplicity.
While you can argue that defining an operation in ODS is simpler, this is still one extra language/tool/infrastructure that has to be mentally mapped to understand the system. Moreover any code that will be written will have to use the C++ API generated from ODS (the `Op` class). My take is that I would learn better by starting with the simple case of writing myself the Op class and using it ("what you see is what you get") instead of hiding it behind a pretty DSL and a code-generator.
YMMV.

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
-In MLIR, the core unit of abstraction and computation is an `Operation`, similar
-in many ways to LLVM instructions. Operations can be used to represent all of
-the core IR structures in LLVM: instructions, globals(like functions), modules,
-etc; however MLIR does not have a closed set of operations. Instead, the MLIR
-operation set is fully extensible and operations can have application-specific
-semantics.
+[Language reference](../../LangRef.md)
 
+MLIR is designed to be a completely extensible infrastructure; there is no
+closed set of attributes(think always constant metadata), operations, or types.
```
joker-eph:
```suggestion
closed set of attributes (think always constant metadata), operations, or types.
```

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
 
-MLIR is designed to be a completely extensible system, as such the
-infrastructure has the capability to opaquely represent operations (as well as
-attributes, types, etc.) that have not been registered. This allows MLIR to
-parse, represent, and round-trip any valid IR. For example, the following can
-round-trip through *mlir-opt*:
+MLIR is designed to be a completely extensible system, and as such, the
+infrastructure has the capability to opaquely represent all of its core
+components: attributes, operations, types, etc. This allows MLIR to parse,
+represent, and round-trip any valid IR. For example, we could inject our toy
```
joker-eph: inject? This sounds strange to me right now

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
 ```
 
-Here MLIR will enforce some structural constraints (SSA, block termination,
-etc.) but otherwise the types and the `custom.operation` are completely opaque.
+In the cases of unregistered attributes, operations, types, MLIR will enforce
+some structural constraints (SSA, block termination, etc.) but otherwise they
+are completely opaque. This can be useful for bootstrapping purposes, but it is
+generally advised against. Opaque operations must be treated conservatively by
+transformations and analyses, cannot be properly verified for custom semantics,
+and are much harder to construct and manipulate.
```
joker-eph: I find this quite underwhelming since this is intended to replace a very large piece of what Ch2 was intended to. It loses the motivation for dialect and registration by papering over the problem. I find this to make the chapter less accessible to beginners (which was the point of having two separate chapters). Can you re-read around line 201 in the original Ch2 to get the spirit?

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
--   A list of operand values.
--   A list of types for result values.
--   A list of attributes.
--   A list of successors blocks (for branches mostly).
--   A list of regions (for structural operations like functions).
+```c++
+/// This is the definition of the Toy dialect. A dialect inherits from
+/// mlir::Dialect and registers custom attributes, operations, and types (in its
+/// constructor). It can also override some general behavior exposed via virtual
+/// methods.
```
joker-eph:
```suggestion
/// methods, which will be demonstrated in later chapters of the tutorial.
```

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
+class ToyDialect : public mlir::Dialect {
+ public:
+  explicit ToyDialect(mlir::MLIRContext *ctx);
+
+  /// Provide a utility accessor to the dialect namespace. This is used by
+  /// several utilities for casting between dialects.
```
joker-eph:
```suggestion
  /// several utilities.
```

I'm not sure what "casting between dialects" means here, and there are likely other uses so we can just drop I think.

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
-To build the `custom.operation` from the listing above, assuming you have a
-`Value *` handle to `%arg0`, is as simple as:
+The dialect can now be registered in the global registry:
 
 ```c++
-// Creation of the state defining the operation:
-mlir::OperationState state(location, "custom.operation");
-state.addOperands(arg0);
+  mlir::registerDialect<ToyDialect>();
+```
```
joker-eph: Side note: I would favor an API where we explicitly load dialects in the context. This seems more controlled (two MLIR based system in the same process would not want to share the list of dialects/patterns in their context necessarily).

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
-  // Create the transpose operation.
-  Operation *newTransposeOp = builder->createOperation(result);
-  return newTransposeOp;
+ToyDialect::ToyDialect(mlir::MLIRContext *ctx)
+    : mlir::Dialect(getDialectNamespace(), ctx) {
+  addOperations<ConstantOp>();
+}
+```
+
+### Op vs Operation: Using MLIR Operations
```
joker-eph: did we brainstorm naming recently? I don't remember if we concluded anything. What about renaming Operation into OpaqueOp for instance?

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
+Operation is the actual opaque instance of the operation, and represents the
+general API into an operation instance. An `Op` is the base class of a derived
+operation, like `ConstantOp`, and acts as smart pointer wrapper around a
+`Operation*`. This means that when we define our Toy operations, we are actually
+providing a clean interface for building and interfacing with the `Operation`
+class; this is why our `ConstantOp` defines no class fields. Therefore, we
+always pass these classes around by-value, instead of by reference or pointer
+(passing by-value is a common idiom and applies similarly to attributes, types,
+etc). We can always get an instance of our toy operation by using LLVM's casting
+infrastructure:
```
joker-eph: This kind of paragraph is exactly why I prefer to keep things separated: now we're not only introducing IR concept, we also introduce a bunch of details around the software engineering structure, mixed with implementation detail (link missing for LLVM casting infrastructure by the way).

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
+ %4 = "toy.constant"() {value = dense<1.0> : tensor<2x3xf64>} : () -> tensor<2x3xf64>
+```
 
-// An operation is not an SSA value (unlike LLVM), because it can return
-// multiple SSA values, the resulting value can be obtained:
-Value *result = customOperation->getResult(0);
+This operation takes zero operands, a
+[dense elements](../../LangRef.md#dense-elements-attribute) attribute named
+`value`, and returns a single result of
+[TensorType](../../LangRef.md#tensor-type). An operation inherits from the CRTP
```
joker-eph: CRTP is not an obvious or pervasive concept IMO

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
+  if (!op)
+    return;
+
+  // Get the internal operation instance back.
+  mlir::Operation *internalOp = op.getOperation();
+  assert(internalOp == op && "these operation instances are the same");
+}
+```
+
+### Using the Operation Definition Specification framework
```
joker-eph:
```suggestion
### Using the Operation Definition Specification (ODS) framework
```

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
+         { value = dense<[[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]> : tensor<2x3xf64> }
+        : () -> tensor<2x3xf64>
+
+    This is only valid if a function named "my_func" exists and takes two
+    arguments.
+  }];
+
+  // The generic call operation takes a symbol reference attribute as the
+  // callee, and two operands for the call.
+  let arguments = (ins SymbolRefAttr:$callee, F64Tensor:$lhs, F64Tensor:$rhs);
```
joker-eph: This does not make sense here.

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
+  /// Build a constant and reuse the type from the given 'value'.
+  static void build(mlir::Builder *builder, mlir::OperationState *state,
+                    mlir::DenseElementsAttr value);
+  /// Build a constant by broadcasting the given 'value'.
+  static void build(mlir::Builder *builder, mlir::OperationState *state,
+                    double value);
+
+  /// Return the value of the constant by fetching it from the attribute.
+  mlir::DenseElementsAttr getValue();
+};
```
joker-eph: This is too complex IMO: we need to build up concept and knowledge more slowly and progressively. That mean no need for more than strictly needed to understand the concept at hand (multiple builders don't seem needed here).

g3doc/Tutorials/Toy/Ch-2.md
```diff
@@ -23,139 +23,320 @@ addresses this issue by being designed for extensibility. As such, there are
...
+    }]>,
+    // Build a constant with a given constant floating-point value.
+    OpBuilder<"Builder *builder, OperationState *result, double value", [{
+      // Forward to a static function in the .cpp file.
+      buildConstantOp(builder, result, value);
+    }]>
+  ];
+
+  // Invoke a static `verify` method to verify this constant operation.
+  let verifier = [{ return ::verify(*this); }];
```
joker-eph: Again, this seems all too complex to me for newcomers / beginners

g3doc/Tutorials/Toy/Ch-2.md
River707: Whoops, thanks.

g3doc/Tutorials/Toy/Ch-2.md
River707: I agree, I'd like to add a step-by-step intro to this in a followup.

g3doc/Tutorials/Toy/Ch-2.md
River707: +1

g3doc/Tutorials/Toy/Ch-2.md
River707: Beefed it up to include a more step-by-step introduction to the tablegen side.

PR #109, opened 2019-08-28
Add a builder for AffineIfOp

nmostafa:
Add the missing builder to AffineIfOp and builder tests.


include/mlir/Dialect/AffineOps/AffineOps.td
```diff
@@ -212,9 +212,10 @@ def AffineIfOp : Affine_Op<"if", [ImplicitAffineTerminator]> {
   let regions = (region SizedRegion<1>:$thenRegion, AnyRegion:$elseRegion);
 
   let skipDefaultBuilders = 1;
+  
   let builders = [
     OpBuilder<"Builder *builder, OperationState *result, "
-              "Value *cond, bool withElseRegion">
+              "ArrayRef<Value *> args, IntegerSet set, bool withElseRegion">
```
bondhugula: The order "IntegerSet set, ArrayRef<Value *> operands" will be more natural and consistent with other builders.

test/EDSC/builder-api-test.cpp
```diff
@@ -22,7 +22,9 @@
 #include "mlir/EDSC/Builders.h"
 #include "mlir/EDSC/Helpers.h"
 #include "mlir/EDSC/Intrinsics.h"
+#include "mlir/IR/AffineExpr.h"
```
bondhugula: You don't need this include - because IntegerSet.h already has it.

lib/Dialect/AffineOps/AffineOps.cpp
```diff
@@ -1573,6 +1573,18 @@ void AffineIfOp::setIntegerSet(IntegerSet newSet) {
   setAttr(getConditionAttrName(), IntegerSetAttr::get(newSet));
 }
 
+void AffineIfOp::build(Builder *builder, OperationState *result,
+                       ArrayRef<Value *> args, IntegerSet set,
+                       bool withElseRegion) {
+  result->addOperands(args);
+  result->addAttribute(getConditionAttrName(), IntegerSetAttr::get(set));
+  Region *thenRegion = result->addRegion();
+  Region *elseRegion = result->addRegion();
```
bondhugula: Shouldn't this be under the withElseRegion conditional below?

test/EDSC/builder-api-test.cpp
```diff
@@ -746,6 +748,40 @@ TEST_FUNC(empty_map_load_store) {
...
+  SmallVector<bool, 4> isEq = {false, false};
+  SmallVector<AffineExpr, 4> affineExprs = {
+    builder.getAffineDimExpr(0) + builder.getAffineSymbolExpr(0), // d0 + s0 >= 0
+    builder.getAffineDimExpr(1) - builder.getAffineSymbolExpr(1), // d1 - s1 >= 0
+  };
+  auto intSet = builder.getIntegerSet(2, 2, affineExprs, isEq);
+
+  LoopNestBuilder({&i, &j}, {zero, zero}, {ten, ten}, {1, 1})([&]{
+    SmallVector<Value*, 4> affineIfArgs = {i, j, zero, ten};
+    intrinsics::affine_if(affineIfArgs, intSet, false);
```
bondhugula: ..., intSet, /*withElseRegion=*/false);

bondhugula: Also, missing test with an 'else' region. You could  perhaps just update this test to include an else region.

lib/Dialect/AffineOps/AffineOps.cpp
bondhugula: > AffneIfOp always has an else region, it is just empty if there is no else.

I see - thanks.

nmostafa:
PR feedback addressed. Please let me know if anything else is needed.

test/EDSC/builder-api-test.cpp
```diff
@@ -746,6 +747,39 @@ TEST_FUNC(empty_map_load_store) {
...
+
+  OpBuilder builder(f.getBody());
+  ScopedContext scope(builder, f.getLoc());
+  // clang-format off
+  ValueHandle zero = constant_index(0), ten = constant_index(10);
+  
+  SmallVector<bool, 4> isEq = {false, false};
+  SmallVector<AffineExpr, 4> affineExprs = {
+    builder.getAffineDimExpr(0) + builder.getAffineSymbolExpr(0), // d0 + s0 >= 0
+    builder.getAffineDimExpr(1) - builder.getAffineSymbolExpr(1), // d1 - s1 >= 0
```
River707: Do we need a complicated example here? Can it be a trivial set?

nmostafa: Just a case where we use both dims and symbols and passing them to the op args. I will eliminate the arithmetic ops in the inequalities (e.g. `d0 >=0`, `s0 >=0` ... )

nmostafa: Done.

test/EDSC/builder-api-test.cpp
```diff
@@ -746,6 +747,41 @@ TEST_FUNC(empty_map_load_store) {
...
+  using namespace edsc;
+  using namespace edsc::intrinsics;
+  using namespace edsc::op;
+  auto f32Type = FloatType::getF32(&globalContext());
+  auto memrefType = MemRefType::get({-1, -1}, f32Type, {}, 0);
+  auto f = makeFunction("affine_if_op", {}, {memrefType});
+
+  OpBuilder builder(f.getBody());
+  ScopedContext scope(builder, f.getLoc());
+  // clang-format off
```
joker-eph: Is this intentional? Can you add a comment on why?
It is usually done exceptionally for well scoped block of code, but here it isn't clear to me that the following declaration needs it and I don't see the end of the unformatted region.

nmostafa: I removed the clang-format. Are you also referring to ScopedContext ? Without it the EDSC builders crash.

joker-eph: No I was just referring to the clang-format directive. Thanks.

test/EDSC/builder-api-test.cpp
```diff
@@ -746,6 +747,41 @@ TEST_FUNC(empty_map_load_store) {
...
+  SmallVector<AffineExpr, 4> affineExprs = {
+    builder.getAffineDimExpr(0),    // d0 >= 0
+    builder.getAffineDimExpr(1),    // d1 >= 0
+    builder.getAffineSymbolExpr(0), // s0 >= 0
+    builder.getAffineSymbolExpr(1)  // s1 >= 0
+  };
+  auto intSet = builder.getIntegerSet(2, 2, affineExprs, isEq);
+  
+  SmallVector<Value*, 4> affineIfArgs = {zero, zero, ten, ten};
+  intrinsics::affine_if(intSet, affineIfArgs, /* withElseRegion */ false);
```
River707: The parameter comments should follow /*...=*/value

joker-eph: To expand a bit, this isn't just nit-picking: this syntax is recognized by clang and it can warn if the name does not match the parameter in the declaration (if you change the declaration later for example), and clang-tidy knows how to maintain these.

nmostafa: I see. Good to know. Fixed.

test/EDSC/builder-api-test.cpp
```diff
@@ -746,6 +747,41 @@ TEST_FUNC(empty_map_load_store) {
   f.erase();
 }
 
+// CHECK-LABEL: func @affine_if_op
+// CHECK:       affine.if ([[d0:.*]], [[d1:.*]]){{\[}}[[s0:.*]], [[s1:.*]]{{\]}}
+// CHECK-NEXT:  }
```
River707: final nit: This 'Check-NEXT: }' doesn't seem necessary.

nmostafa: It verifies absence of else block

River707: No it doesn't, CHECKs don't check the entire line. This means that if it did have a " else {" afterwards, the check would still pass. Change to `CHECK-NOT: else`, if you want to make sure there is no else.

PR #110, opened 2019-08-30
update vim syntax file

bondhugula:
- more highlighting: numbers, elemental types inside shaped types
- add some more keywords


bondhugula:
**Before:**
before



**After:**
after

PR #112, opened 2019-08-30
Integer set + operands / affine if op canonicalization

bondhugula:
- turn canonicalizeMapAndOperands into a template that works on both
  sets and maps, and use it to introduce a utility to canonicalize an
  affine integer set and its operands
- add pattern to canonicalize affine if op's.
- rename IntegerSet::getNumOperands -> IntegerSet::getNumInputs to be
  consistent with AffineMap
- add missing accessors for IntegerSet

Doesn't need extensive testing since canonicalizeSetAndOperands just
reuses canonicalizeMapAndOperands' logic, and the latter is tested on
affine.apply map + operands; the new method works the same way on an
integer set + operands of an affine if op for example.

### DIFF
--- a/include/mlir/Dialect/AffineOps/AffineOps.h
+++ b/include/mlir/Dialect/AffineOps/AffineOps.h
@@ -522,6 +522,10 @@ bool isValidSymbol(Value *value);
 /// 2. drop unused dims and symbols from map
 void canonicalizeMapAndOperands(AffineMap *map,
                                 llvm::SmallVectorImpl<Value *> *operands);
+/// Canonicalizes an integer set the same way canonicalizeMapAndOperands does
+/// for affine maps.
+void canonicalizeSetAndOperands(IntegerSet *set,
+                                llvm::SmallVectorImpl<Value *> *operands);
 
 /// Returns a composed AffineApplyOp by composing `map` and `operands` with
 /// other AffineApplyOps supplying those operands. The operands of the resulting
--- a/include/mlir/Dialect/AffineOps/AffineOps.td
+++ b/include/mlir/Dialect/AffineOps/AffineOps.td
@@ -223,6 +223,11 @@ def AffineIfOp : Affine_Op<"if", [ImplicitAffineTerminator]> {
     IntegerSet getIntegerSet();
     void setIntegerSet(IntegerSet newSet);
 
+    /// Sets the integer set with its operands. The size of 'operands' must not
+    /// exceed the current number of operands for this instance, as the operands
+    /// list of AffineIf is not resizable.
+    void setConditional(IntegerSet set, ArrayRef<Value *> operands);
+
     OpBuilder getThenBodyBuilder() {
       assert(!thenRegion().empty() && "Unexpected empty 'then' region.");
       Block &body = thenRegion().front();
@@ -234,6 +239,8 @@ def AffineIfOp : Affine_Op<"if", [ImplicitAffineTerminator]> {
       return OpBuilder(&body, std::prev(body.end()));
     }
   }];
+
+  let hasCanonicalizer = 1;
 }
 
 def AffineTerminatorOp :
--- a/include/mlir/IR/IntegerSet.h
+++ b/include/mlir/IR/IntegerSet.h
@@ -72,12 +72,22 @@ public:
   /// Returns true if this is the canonical integer set.
   bool isEmptyIntegerSet() const;
 
+  /// This method substitutes any uses of dimensions and symbols (e.g.
+  /// dim#0 with dimReplacements[0]) in subexpressions and returns the modified
+  /// integer set.  Because this can be used to eliminate dims and
+  /// symbols, the client needs to specify the number of dims and symbols in
+  /// the result.  The returned map always has the same number of results.
+  IntegerSet replaceDimsAndSymbols(ArrayRef<AffineExpr> dimReplacements,
+                                   ArrayRef<AffineExpr> symReplacements,
+                                   unsigned numResultDims,
+                                   unsigned numResultSyms);
+
   explicit operator bool() { return set; }
   bool operator==(IntegerSet other) const { return set == other.set; }
 
   unsigned getNumDims() const;
   unsigned getNumSymbols() const;
-  unsigned getNumOperands() const;
+  unsigned getNumInputs() const;
   unsigned getNumConstraints() const;
   unsigned getNumEqualities() const;
   unsigned getNumInequalities() const;
@@ -95,6 +105,10 @@ public:
   bool isEq(unsigned idx) const;
 
   MLIRContext *getContext() const;
+
+  /// Walk all of the AffineExpr's in this set's constraints. Each node in an
+  /// expression tree is visited in postorder.
+  void walkExprs(llvm::function_ref<void(AffineExpr)> callback) const;
 
   void print(raw_ostream &os) const;
   void dump() const;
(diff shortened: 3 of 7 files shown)


include/mlir/Dialect/AffineOps/AffineOps.h
```diff
@@ -522,6 +522,10 @@ bool isValidSymbol(Value *value);
 /// 2. drop unused dims and symbols from map
 void canonicalizeMapAndOperands(AffineMap *map,
                                 llvm::SmallVectorImpl<Value *> *operands);
+/// Canonicalizes an integer set the same way canonicalizeMapAndOperands does
+/// for affine maps.
+void canonicalizeSetAndOperands(IntegerSet *set,
+                                llvm::SmallVectorImpl<Value *> *operands);
```
River707: We should really be passing by reference here.

include/mlir/Dialect/AffineOps/AffineOps.td
```diff
@@ -223,6 +223,8 @@ def AffineIfOp : Affine_Op<"if", [ImplicitAffineTerminator]> {
     IntegerSet getIntegerSet();
     void setIntegerSet(IntegerSet newSet);
 
+    void setConditional(IntegerSet set, ArrayRef<Value *>);
```
River707: The ArrayRef needs a variable name

include/mlir/IR/IntegerSet.h
```diff
@@ -96,6 +106,10 @@ class IntegerSet {
 
   MLIRContext *getContext() const;
 
+  /// Walk all of the AffineExpr's in this set's constraints. Each node in an
+  /// expression tree is visited in postorder.
+  void walkExprs(std::function<void(AffineExpr)> callback) const;
```
River707: use llvm::function_ref instead, constructing std::function is costly.

lib/Dialect/AffineOps/AffineOps.cpp
```diff
@@ -1585,6 +1603,49 @@ void AffineIfOp::setIntegerSet(IntegerSet newSet) {
...
+    SmallVector<Value *, 4> operands(ifOp.getOperands());
+
+    canonicalizeSetAndOperands(&set, &operands);
+
+    // Any canonicalization change always leads to either a reduction in the
+    // number of operands or a change in the number of symbolic operands
+    // (promotion of dims to symbols).
+    bool updated = false;
+    if (operands.size() < ifOp.getIntegerSet().getNumInputs() ||
+        set.getNumSymbols() > ifOp.getIntegerSet().getNumSymbols()) {
```
River707: Just early exit here:

if (...)
  return matchFailure();

...

include/mlir/Dialect/AffineOps/AffineOps.h
bondhugula: canonicalizeMapAndOperands and canonicalizeSetAndOperands work in the same way, and if we change this one, it'd require changing the other one as well to be consistent and at the numerous places that's called from. I think it should only be done in another NFC PR. 

On another note, the ampersands really help with readability since they indicate output parameters, and make it obvious at the call site that the passed things are being changed. If we'd like to move to pass by reference for all non-nullable inputs or outputs consistently, I think this should be stated in the style guide. 

```
auto map = ...
canonicalizeMapAndOperands(&map, &operands);
...
```

lib/Dialect/AffineOps/AffineOps.cpp
bondhugula: Done.

include/mlir/IR/IntegerSet.h
bondhugula: Done.

include/mlir/Dialect/AffineOps/AffineOps.td
bondhugula: Done.

include/mlir/Dialect/AffineOps/AffineOps.h
joker-eph: Done in 57ebdf105583cb39bd8d29b83dc50dd0cf50868e

include/mlir/Dialect/AffineOps/AffineOps.td
```diff
@@ -223,6 +223,8 @@ def AffineIfOp : Affine_Op<"if", [ImplicitAffineTerminator]> {
     IntegerSet getIntegerSet();
     void setIntegerSet(IntegerSet newSet);
 
+    void setConditional(IntegerSet set, ArrayRef<Value *> operands);
```
River707: The AffineIf builders will need to be updated to allow for a variadic operands list.

bondhugula: I assume you mean resizable operands list? The canonicalization here always shrinks the list. So I didn't make update it - there isn't a use case in MLIR yet that could make the operands list of AffineIf grow.

River707: We should guard against that then, can you document the expectation here that there are less operands? There should also be an assert for this case.

bondhugula: Documented this. There is already the guard in setOperands which will assert if one tries to grow the list. Do you think we should add one more here?

River707: Can you reword the constraint a bit? It reads a little weird to me, something like "The size of 'operands' must not exceed the current number operands for this instance, as the operands list of AffineIf is not resizable"

lib/Dialect/AffineOps/AffineOps.cpp
```diff
@@ -1600,6 +1618,45 @@ void AffineIfOp::setIntegerSet(IntegerSet newSet) {
...
+    canonicalizeSetAndOperands(&set, &operands);
+
+    // Any canonicalization change always leads to either a reduction in the
+    // number of operands or a change in the number of symbolic operands
+    // (promotion of dims to symbols).
+    if (operands.size() < ifOp.getIntegerSet().getNumInputs() ||
+        set.getNumSymbols() > ifOp.getIntegerSet().getNumSymbols()) {
+      ifOp.setConditional(set, operands);
+      rewriter.updatedRootInPlace(ifOp);
+      // If the conditional was changed, we return success.
```
River707: This comment seems unnecessary.

include/mlir/Dialect/AffineOps/AffineOps.td
bondhugula: Fixed - thanks.

lib/Dialect/AffineOps/AffineOps.cpp
bondhugula: Dropped.

lib/Dialect/AffineOps/AffineOps.cpp
```diff
@@ -1600,6 +1618,44 @@ void AffineIfOp::setIntegerSet(IntegerSet newSet) {
...
+      rewriter.updatedRootInPlace(ifOp);
+      return matchSuccess();
+    }
+
+    return matchFailure();
+  }
+};
+} // end anonymous namespace
+
+void AffineIfOp::getCanonicalizationPatterns(
```
ftynse: Could you please reflow this as 

```
void AffineIfOp::getCanonicalizationPatterns(OwningRewritePatternList &results,
    MLIRContext *context) {
```

to comply with the style?

bondhugula: Done, thanks.

PR #116, opened 2019-09-01
Remove unused variables

alexst07:
Remove unused variables on mlir/lib/Dialect/Linalg/Transforms/LowerToLLVMDialect.cpp

### DIFF
--- a/lib/Dialect/Linalg/Transforms/LowerToLLVMDialect.cpp
+++ b/lib/Dialect/Linalg/Transforms/LowerToLLVMDialect.cpp
@@ -185,20 +185,19 @@ public:
   BaseViewConversionHelper(Operation *op, ViewType viewType,
                            ConversionPatternRewriter &rewriter,
                            LLVMTypeConverter &lowering)
-      : indexType(rewriter.getIndexType()), viewType(viewType),
-        elementTy(getPtrToElementType(viewType, lowering)),
+      : elementTy(getPtrToElementType(viewType, lowering)),
         int64Ty(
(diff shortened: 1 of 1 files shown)


lib/Dialect/Linalg/Transforms/LowerToLLVMDialect.cpp
```diff
@@ -594,10 +594,7 @@ class TransposeOpConversion : public LLVMOpLowering {
     auto tranposeOp = cast<TransposeOp>(op);
     BaseViewConversionHelper helper(op, tranposeOp.getViewType(), rewriter,
                                     lowering);
-    IndexType indexType = helper.indexType;
-    ViewType viewType = helper.viewType;
-    LLVMType elementTy = helper.elementTy, int64Ty = helper.int64Ty,
-             viewDescriptorPtrTy = helper.viewDescriptorPtrTy;
```
joker-eph: Seems like some of these (like `indexType`) can be removed as a member of `BaseViewConversionHelper` as well?

alexst07: Yes, there are some attributes that no other part of the code is using. I will remove it and update the pull request.

joker-eph: Thanks

nicolasvasilache:
Thanks for the cleanup, didn't get time to finish refactoring that piece
before leaving on PTO :)

Issue #60, opened 2019-08-01
Add mlir-vulkan-runner


antiagainst:
It would be nice to have a simple testing harness to exercise the SPIR-V blob generated and serialized from SPIR-V CodeGen on Vulkan-capable devices, similar to mlir-cuda-runner.

Ideally the runner should inspect the SPIR-V module to help building the resource interface for a Vulkan compute pipeline, run the compute pipeline by feeding in test data and then check the result data.

denis0x0D:
@antiagainst if noone is working on it, I can take it, I've already have a kinda "Vulkan runtime" to test spir-v compute shaders generated by hands written SPIR-V IRBuilder, I learned it from the blog of Neil Henning http://www.duskborn.com/posts/a-simple-vulkan-compute-example/ and code is mostly the same.

antiagainst:
Please feel free to proceed if you'd like to take on it! Thanks a lot, @denis0x0D!

denis0x0D:
@antiagainst thanks!

MaheshRavishankar:
Fantastic! Thanks Denis!

denis0x0D:
@antiagainst @MaheshRavishankar can you please comment on the following, to make sure I understand the specifications of the task in right way,  and please fix me if I'm wrong.

From the high-level view the mlir-vulkan-runner must consits of 2 main parts:
compile-time passes and runtime machinery.

1. Pipeline of the passes:
a. The pass which outlines gpu.launch body to kernel function. (GpuKernelOutLiningPass)
b. The pass which converts GPU dialect into SPIR-V dialect. (GPUToSPIRVPass)
c. The codegen stage which generates spir-v binary from spir-v kernel. (SPIR-V serialization) is similiar to ConvertGPUKernelToCubinPass.
d. The pass which replaces "gpu.launch_function" with call into runtime part, and instruments code with calls to runtime machinery simliar to GpuLanchFunctToCudaCallsPass does.

2. Runtime machinery:
a. Consists of the interceptors which must use Vulkan API.
b. Takes responsibilities for the memory management, buffer creation, device creation, kernel launch and so on.

So, in this case I can reuse existing mlir::JitRunnerMain which uses mlir::ExecuteEngine to run the program.
For example the representation of the module after GPU->SPIRV converter pass.

```
module {
  spv.module "Logical" "VulkanKHR" {
    %0 = spv.Variable bind(0, 0) : !spv.ptr<f32, StorageBuffer>
[...]

PR #118, opened 2019-09-02
[WIP] [spirv] Add Vulkan runtime.

denis0x0D:
This patch is addressing https://github.com/tensorflow/mlir/issues/60
Implements initial version of Vulkan runtime to test spirv::ModuleOp.
Creates and runs Vulkan computation pipeline.

Provides one unititest to test multiplication of two spv.arrays of float types.

Requirements:
- Vulkan capable device and drivers installed.
- Vulkan SDK installed and VULKAN_SDK environment variable is set.

How to build:
- Provide -DMLIR_VULKAN_RUNNER_ENABLED as cmake variable.

Note:
- Tested only under the GNU/Linux distribution with official Vulkan SDK provided by https://vulkan.lunarg.com/doc/sdk/latest/linux/getting_started.html

Thanks!

### DIFF
--- a/tools/mlir-vulkan-runner/CMakeLists.txt
+++ b/tools/mlir-vulkan-runner/CMakeLists.txt
@@ -1,0 +1,45 @@
+set(LLVM_OPTIONAL_SOURCES
+  VulkanRuntime.cpp
+  )
+
+if (MLIR_VULKAN_RUNNER_ENABLED)
+  message(STATUS "Building the Vulkan runner")
+
+  # At first try "FindVulkan" from:
+  # https://cmake.org/cmake/help/v3.7/module/FindVulkan.html
+  if (NOT CMAKE_VERSION VERSION_LESS 3.7.0)
+    find_package(Vulkan)
+  endif()
+
+  # If Vulkan is not found try a path specified by VULKAN_SDK.
+  if (NOT Vulkan_FOUND)
+    if ("$ENV{VULKAN_SDK}" STREQUAL "")
+      message(FATAL_ERROR "VULKAN_SDK environment variable is empty")
+    endif()
+
+    find_library(Vulkan_LIBRARY vulkan HINTS "$ENV{VULKAN_SDK}/lib" REQUIRED)
+    if (Vulkan_LIBRARY)
+      set(Vulkan_FOUND ON)
+      set(Vulkan_INCLUDE_DIR "$ENV{VULKAN_SDK}/include")
+      message(STATUS "Found Vulkan: " ${Vulkan_LIBRARY})
+    endif()
+  endif()
+
+  if (NOT Vulkan_FOUND)
+    message(FATAL_ERROR "Cannot find Vulkan library")
+  endif()
+
+  # Create archive for this moment.
+  add_llvm_library(MLIRVulkanRuntime
+    VulkanRuntime.cpp
+    )
+
+  target_include_directories(MLIRVulkanRuntime PRIVATE ${Vulkan_INCLUDE_DIR})
+
+  target_link_libraries(MLIRVulkanRuntime
+    MLIRSPIRVSerialization
+    LLVMCore
+    LLVMSupport
+    ${Vulkan_LIBRARY}
+    )
+endif()
--- a/tools/mlir-vulkan-runner/VulkanRuntime.cpp
+++ b/tools/mlir-vulkan-runner/VulkanRuntime.cpp
@@ -1,0 +1,1029 @@
+//===- VulkanRuntime.cpp - MLIR Vulkan runtime ------------------*- C++ -*-===//
+//
+// Copyright 2019 The MLIR Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+//
+// This file provides a library for running a module on a Vulkan device.
+// Implements a Vulkan runtime to run a spirv::ModuleOp. It also defines a few
+// utility functions to extract information from a spirv::ModuleOp.
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Analysis/Passes.h"
+#include "mlir/Dialect/SPIRV/SPIRVOps.h"
+#include "mlir/Dialect/SPIRV/Serialization.h"
+#include "mlir/IR/Module.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Support/StringExtras.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/Support/ToolOutputFile.h"
+
+#include <vulkan/vulkan.h>
+
+using namespace mlir;
+
+using DescriptorSetIndex = uint32_t;
+using BindingIndex = uint32_t;
+
+/// Struct containing information regarding to a device memory buffer.
+struct VulkanDeviceMemoryBuffer {
+  BindingIndex bindingIndex{0};
+  VkDescriptorType descriptorType{VK_DESCRIPTOR_TYPE_MAX_ENUM};
+  VkDescriptorBufferInfo bufferInfo;
+  VkBuffer buffer;
+  VkDeviceMemory deviceMemory;
+};
+
+/// Struct containing information regarding to a host memory buffer.
+struct VulkanHostMemoryBuffer {
+  /// Pointer to a host memory.
+  void *ptr{nullptr};
+  /// Size of a host memory in bytes.
+  uint32_t size{0};
+};
+
+/// Struct containing the number of local workgroups to dispatch for each
+/// dimension.
+struct NumWorkGroups {
+  uint32_t x{1};
+  uint32_t y{1};
+  uint32_t z{1};
+};
+
+/// Struct containing information regarding to a descriptor set.
+struct DescriptorSetInfo {
+  /// Index of a descriptor set in descriptor sets.
+  DescriptorSetIndex descriptorSet{0};
+  /// Number of desriptors in a set.
+  uint32_t descriptorSize{0};
+  /// Type of a descriptor set.
+  VkDescriptorType descriptorType{VK_DESCRIPTOR_TYPE_MAX_ENUM};
+};
+
+/// VulkanHostMemoryBuffer mapped into a descriptor set and a binding.
+using ResourceData =
+    llvm::DenseMap<DescriptorSetIndex,
+                   llvm::DenseMap<BindingIndex, VulkanHostMemoryBuffer>>;
+
+/// StorageClass mapped into a descriptor set and a binding.
+using ResourceStorageClassData =
+    llvm::DenseMap<DescriptorSetIndex,
+                   llvm::DenseMap<BindingIndex, mlir::spirv::StorageClass>>;
+
+inline void emitVulkanError(const llvm::Twine &message, VkResult error) {
+  llvm::errs()
+      << message.concat(" failed with error code ").concat(llvm::Twine{error});
+}
+
+#define RETURN_ON_VULKAN_ERROR(result, msg)                                    \
+  if ((result) != VK_SUCCESS) {                                                \
+    emitVulkanError(msg, (result));                                            \
+    return failure();                                                          \
+  }
+
+namespace {
+/// Processes spv.module and collects all needed information for VulkanRuntime.
+class SPIRVModuleInfoCollector {
+public:
+  explicit SPIRVModuleInfoCollector() = default;
+  std::string getEntryPoint() { return entryPoint; }
+  ResourceStorageClassData &getResourceStorageClassData() {
+    return resourceStorageClassData;
+  }
+  void processModule(spirv::ModuleOp module);
+
+private:
+  SPIRVModuleInfoCollector(const SPIRVModuleInfoCollector &) = delete;
+  SPIRVModuleInfoCollector &operator=(const SPIRVModuleInfoCollector &) = delete;
+  void processGlobalVariable(spirv::GlobalVariableOp varOp);
+  void processEntryPoint(spirv::EntryPointOp op);
+
+  std::string entryPoint;
+  ResourceStorageClassData resourceStorageClassData;
+};
+
+/// Processes ModuleOp to collect module specific information.
+/// Note: ModuleOp must be valid.
+void SPIRVModuleInfoCollector::processModule(spirv::ModuleOp module) {
+  for (auto &op : module.getBlock()) {
+    if (auto entryPointOp = dyn_cast<spirv::EntryPointOp>(op)) {
+      processEntryPoint(entryPointOp);
+    } else if (auto varOp = dyn_cast<spirv::GlobalVariableOp>(op)) {
+      processGlobalVariable(varOp);
+    }
+  }
+}
+
+/// Processes EntryPointOp to collect entry point.
+void SPIRVModuleInfoCollector::processEntryPoint(spirv::EntryPointOp op) {
+  entryPoint = op.fn();
+}
+
+/// Processes GlobalVariableOp to collect storage classes for resource data.
+void SPIRVModuleInfoCollector::processGlobalVariable(
+    spirv::GlobalVariableOp varOp) {
+  auto descriptorSetName =
+      convertToSnakeCase(stringifyDecoration(spirv::Decoration::DescriptorSet));
+  auto bindingName =
+      convertToSnakeCase(stringifyDecoration(spirv::Decoration::Binding));
+  auto descriptorSet = varOp.getAttrOfType<IntegerAttr>(descriptorSetName);
+  auto binding = varOp.getAttrOfType<IntegerAttr>(bindingName);
+
+  if (descriptorSet && binding) {
+    if (auto ptrType = varOp.type().dyn_cast<spirv::PointerType>()) {
+      auto descriptorBindingIndex = binding.getInt();
+      auto descriptorSetIndex = descriptorSet.getInt();
+      resourceStorageClassData[descriptorSetIndex][descriptorBindingIndex] =
+          ptrType.getStorageClass();
+    }
+  }
+}
+} // namespace
+
+namespace {
+
+/// Vulkan runtime.
+/// The purpose of this class is to run spir-v computaiton shader on Vulkan
+/// device.
+/// Before the run, user must provide and set resource data with descriptors,
+/// spir-v shader, number of work groups and entry point. After the creation of
+/// VulkanRuntime, special methods must be called in the following
+/// sequence: initRuntime(), run(), updateHostMemoryBuffers(), destroy();
+/// each method in the sequence returns sussses or failure depends on the Vulkan
+/// result code.
+class VulkanRuntime {
+public:
+  explicit VulkanRuntime() = default;
+  VulkanRuntime(const VulkanRuntime &) = delete;
+  VulkanRuntime &operator=(const VulkanRuntime &) = delete;
+
+  /// Sets needed data for Vulkan runtime.
+  void setResourceData(const ResourceData &resData);
+  void setShaderModule(llvm::ArrayRef<uint32_t> binaryRef);
+  void setNumWorkGroups(const NumWorkGroups &nWorkGroups);
+  void setResourceStorageClassData(const ResourceStorageClassData &stClassData);
+  void setEntryPoint(llvm::StringRef entryPointName);
+
+  /// Runtime initialization.
+  LogicalResult initRuntime();
+
+  /// Runs runtime.
+  LogicalResult run();
+
+  /// Updates host memory buffers.
+  LogicalResult updateHostMemoryBuffers();
+
+  /// Destroys all created vulkan objects and resources.
+  LogicalResult destroy();
+
+private:
+  //===--------------------------------------------------------------------===//
+  // Pipeline creation methods.
+  //===--------------------------------------------------------------------===//
+
+  LogicalResult createInstance();
+  LogicalResult createDevice();
+  LogicalResult getBestComputeQueue(const VkPhysicalDevice &physicalDevice);
+  LogicalResult createMemoryBuffers();
+  LogicalResult createShaderModule();
+  void initDescriptorSetLayoutBindingMap();
+  LogicalResult createDescriptorSetLayout();
+  LogicalResult createPipelineLayout();
+  LogicalResult createComputePipeline();
+  LogicalResult createDescriptorPool();
+  LogicalResult allocateDescriptorSets();
+  LogicalResult setWriteDescriptors();
+  LogicalResult createCommandPool();
+  LogicalResult createComputeCommandBuffer();
+  LogicalResult submitCommandBuffersToQueue();
+
+  //===--------------------------------------------------------------------===//
+  // Helper methods.
+  //===--------------------------------------------------------------------===//
+
+  /// Maps storage class to a descriptor type.
+  LogicalResult
+  mapStorageClassToDescriptorType(spirv::StorageClass storageClass,
+                                  VkDescriptorType &descriptorType);
+
+  /// Maps storage class to buffer usage flags.
+  LogicalResult
+  mapStorageClassToBufferUsageFlag(spirv::StorageClass storageClass,
+                                   VkBufferUsageFlagBits &bufferUsage);
+
+  LogicalResult countDeviceMemorySize();
+
+  //===--------------------------------------------------------------------===//
+  // Vulkan objects.
+  //===--------------------------------------------------------------------===//
+
+  VkInstance instance;
+  VkDevice device;
+  VkQueue queue;
+
+  /// Specifies VulkanDeviceMemoryBuffers devided into sets.
+  llvm::DenseMap<DescriptorSetIndex,
+                 llvm::SmallVector<VulkanDeviceMemoryBuffer, 1>>
+      deviceMemoryBufferMap;
+
+  /// Specifies shader module.
+  VkShaderModule shaderModule;
+
+  /// Specifies layout bindings.
+  llvm::DenseMap<DescriptorSetIndex,
+                 llvm::SmallVector<VkDescriptorSetLayoutBinding, 1>>
+      descriptorSetLayoutBindingMap;
+
+  /// Specifies layouts of descriptor sets.
+  llvm::SmallVector<VkDescriptorSetLayout, 1> descriptorSetLayouts;
+  VkPipelineLayout pipelineLayout;
+
+  /// Specifies descriptor sets.
+  llvm::SmallVector<VkDescriptorSet, 1> descriptorSets;
+
+  /// Specifies a pool of descriptor set info, each descriptor set must have
+  /// information such as type, index and amount of bindings.
+  llvm::SmallVector<DescriptorSetInfo, 1> descriptorSetInfoPool;
+  VkDescriptorPool descriptorPool;
+
+  /// Computation pipeline.
+  VkPipeline pipeline;
+  VkCommandPool commandPool;
+  llvm::SmallVector<VkCommandBuffer, 1> commandBuffers;
+
+  //===--------------------------------------------------------------------===//
+  // Vulkan memory context.
+  //===--------------------------------------------------------------------===//
+
+  uint32_t queueFamilyIndex{0};
+  uint32_t memoryTypeIndex{VK_MAX_MEMORY_TYPES};
+  VkDeviceSize memorySize{0};
+
+  //===--------------------------------------------------------------------===//
+  // Vulkan execution context.
+  //===--------------------------------------------------------------------===//
+
+  NumWorkGroups numWorkGroups;
+  std::string entryPoint;
+  llvm::SmallVector<uint32_t, 0> binary;
+
+  //===--------------------------------------------------------------------===//
+  // Vulkan resource data and storage classes.
+  //===--------------------------------------------------------------------===//
+
+  ResourceData resourceData;
+  ResourceStorageClassData resourceStorageClassData;
+};
+
+void VulkanRuntime::setNumWorkGroups(const NumWorkGroups &nWorkGroups) {
+  numWorkGroups = nWorkGroups;
+}
+
+void VulkanRuntime::setResourceStorageClassData(
+    const ResourceStorageClassData &stClassData) {
+  resourceStorageClassData = stClassData;
+}
+
+void VulkanRuntime::setEntryPoint(llvm::StringRef entryPointName) {
+  entryPoint = entryPointName;
+}
+
+void VulkanRuntime::setResourceData(const ResourceData &resData) {
+  resourceData = resData;
+}
+
+void VulkanRuntime::setShaderModule(llvm::ArrayRef<uint32_t> binaryRef) {
+  binary = SmallVector<uint32_t, 0>(binaryRef.begin(), binaryRef.end());
+}
+
+LogicalResult VulkanRuntime::mapStorageClassToDescriptorType(
+    spirv::StorageClass storageClass, VkDescriptorType &descriptorType) {
+  switch (storageClass) {
+  case spirv::StorageClass::StorageBuffer:
+    descriptorType = VK_DESCRIPTOR_TYPE_STORAGE_BUFFER;
+    break;
+  case spirv::StorageClass::Uniform:
+    descriptorType = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+    break;
+  default:
+    return failure();
+  }
+  return success();
+}
+
+LogicalResult VulkanRuntime::mapStorageClassToBufferUsageFlag(
+    spirv::StorageClass storageClass, VkBufferUsageFlagBits &bufferUsage) {
+  switch (storageClass) {
+  case spirv::StorageClass::StorageBuffer:
+    bufferUsage = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
+    break;
+  case spirv::StorageClass::Uniform:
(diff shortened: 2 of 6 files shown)


tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,841 @@
...
+  deviceCreateInfo.ppEnabledLayerNames = nullptr;
+  deviceCreateInfo.enabledExtensionCount = 0;
+  deviceCreateInfo.ppEnabledExtensionNames = nullptr;
+  deviceCreateInfo.pEnabledFeatures = nullptr;
+
+  RETURN_ON_VULKAN_ERROR(
+      vkCreateDevice(physicalDevice, &deviceCreateInfo, 0, &device),
+      "vkCreateDevice");
+
+  VkPhysicalDeviceMemoryProperties properties;
```
antiagainst: Let's initialize these local variables with `= {}` otherwise MSAN might be unhappy.

tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,841 @@
...
+  vkGetPhysicalDeviceMemoryProperties(physicalDevice, &properties);
+
+  for (uint32_t i = 0, e = properties.memoryTypeCount; i < e; ++i) {
```
antiagainst: Can you put comments here on what we are doing here so it's easier for others to follow?

antiagainst (commented):
Awesome, thanks @denis0x0D!

I've quite a few inline comments. But generally, I think we should add more documentation and improve how errors are reported. :)

denis0x0D:
@antiagainst thanks a lot for review! I will update the patch regarding to the comments.

include/mlir/Support/VulkanRuntime.h
```diff
@@ -0,0 +1,59 @@
+//===- VulkanRuntime.h - MLIR Vulkan runtime ------------------------------===//
```
River707: This file should be in /Support.

denis0x0D: @River707 thanks, I'll delete it, since it's just for the test purpose, I think I can declare the function inside the test file.

tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,841 @@
...
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Support/StringExtras.h"
+#include "mlir/Support/VulkanRuntime.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/Support/ToolOutputFile.h"
+
+using namespace mlir;
+
+inline void emit_vulkan_error(const llvm::Twine &message, VkResult error) {
+  llvm::errs()
```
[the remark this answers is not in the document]
denis0x0D: @antiagainst thanks, I'll update it, I was thinkg about to be independent from MLIR context for error messages, but it's wrong.

unittests/Dialect/SPIRV/Runtime/RuntimeTests.cpp
```diff
@@ -0,0 +1,150 @@
...
+    if (failed(runOnModule(moduleRef.get(), vars, numWorkGroups))) {
+      return failure();
+    }
+
+    return success();
+  }
+
+  void createResourceVarFloat(uint32_t id, uint32_t elementCount) {
+    std::srand(unsigned(std::time(0)));
+    float *ptr = new float[elementCount];
```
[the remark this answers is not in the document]
denis0x0D: @antiagainst can you please specify, do I unrerstand you right, you want to update VulkanHostMemoryBuffe to hold unique ptr instead pointer to void*?

denis0x0D:
@antiagainst I've updated the patch regarding to your comments and have some thoughts about the error reporting machinery in particular and about the vulkan-runner in general,  
the following text may contain repetitions with what has already been agreed, sorry about it, but I want make sure that I unrerstand all in the right way. Can you please comment on the following, please fix me if I'm wrong.

1. I was looking at the current realization of other runtime libs inside MLIR and they are using `llvm:errs()`, for example cuda runtime wrappers https://github.com/tensorflow/mlir/blob/master/tools/mlir-cuda-runner/cuda-runtime-wrappers.cpp#L34

2. It could be better for Vulkan runtime to be independent from MLIR context, for example the full pipeline for vulkan-runner could look like this:

2.1. Compile-time passes:

a. Pass which serializes spir-v module into binary and inserts it as an attribute, similar to GpuToCubinPass: 
```
function.setAttr(kCubinAnnotation,
                   builder.getStringAttr({cubin->data(), cubin->size()}));
```
b. Pass which creates global constant containing spir-v binary, similar to GpuGenerateCubinAccessorsPass:
```
Value *startPtr = LLVM::createGlobalString(
        loc, builder, StringRef(nameBuffer), blob.getValue(), llvmDialect);
    builder.create<LLVM::ReturnOp>(loc, startPtr);
```
c. And the final pass which converts launch call into calls for Vulkan runtime wrappers, instruments calls for buffers registration into Vulkan runtime and so on.

2.2. Runtime part must consist of two parts: Vulkan runtime wrappers over actual runtime and actual Vulkan runtime which manages memory, descriptors, layouts, pipeline and so on.
In this case the mlir module could look like this:

```
module {
  func @main() {
     // ...
    call @vkPopulateBuffer(%buff : memref<x?f32>, %desriptorSet : i32, %descriptorBinding : i32, %storageClass : i32) ...
    // ...
    call @vkLaunchShader(...) ...
    // ...
 }
// extern @vkPopulateBuffer
// extern @vkLaunchShader
```

Vulkan runtime wrappers could look like this:

```
extern "C" void vkPopulateBuffer(const memref_t buff, int32_t descriptorSet,
 int32_t decsriptorBinding, int32_t stroageClass) {

     VulkanRuntimeManager::instance()->registerBuffer(
      vulkanHostMemoryBuffer(buff.values, buff.length * sizeof(float)),
      decsriptorSet, descriptorBinding, storageClass);
}
```
VulkanRuntimeManager could look like this:

```
class VulkanRuntimeManager {
public:
  static VulkanRuntimeManager *instance() {
  static VulkanRuntimeManager *mng = new VulkanRuntimeManager;
  return mng;
}
void registerBuffer(VulkanHostMemoryBuffer buffer, int32_t set, int32_t binding, int32_t storageclass) {
 lock(m);
 runtime.registerBuffer(...)
}
void laucnhShader(...) ;
private:
mutex m;
VulkanRuntime runtime;
};
```

In this case we can at first register all needed information for Vulkan runtime such as resources (set, binding), entry point, number of work groups, then create device memory buffers, descriptor set layouts, pipeline layout and bind it to computation pipeline, create command buffer and finaly submit command buffer to the working queue, what do you think?
Thanks.

tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,1029 @@
...
+          ptrType.getStorageClass();
+    }
+  }
+}
+} // namespace
+
+namespace {
+
+/// Vulkan runtime.
+/// The purpose of this class is to run spir-v computaiton shader on Vulkan
```
kiszk: nit: `computaiton` -> `computation`

tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,1029 @@
...
+/// device.
+/// Before the run, user must provide and set resource data with descriptors,
+/// spir-v shader, number of work groups and entry point. After the creation of
+/// VulkanRuntime, special methods must be called in the following
+/// sequence: initRuntime(), run(), updateHostMemoryBuffers(), destroy();
+/// each method in the sequence returns sussses or failure depends on the Vulkan
```
kiszk: nit: `sussses` -> `success`

tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,1029 @@
...
+
+  //===--------------------------------------------------------------------===//
+  // Vulkan objects.
+  //===--------------------------------------------------------------------===//
+
+  VkInstance instance;
+  VkDevice device;
+  VkQueue queue;
+
+  /// Specifies VulkanDeviceMemoryBuffers devided into sets.
```
kiszk: nit: `devided` -> `divided`

tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,1029 @@
...
+}
+
+LogicalResult VulkanRuntime::run() {
+  // Create logical device, shader module and memory buffers.
+  if (failed(createInstance()) || failed(createDevice()) ||
+      failed(createMemoryBuffers()) || failed(createShaderModule())) {
+    return failure();
+  }
+
+  // Descriptor bindings devided into sets. Each descriptor binding
```
kiszk: nit: `devided` -> `divided`

unittests/Dialect/SPIRV/Runtime/RuntimeTests.cpp
```diff
@@ -0,0 +1,235 @@
...
+protected:
+  llvm::DenseMap<DescriptorSetIndex,
+                 llvm::DenseMap<BindingIndex, VulkanHostMemoryBuffer>>
+      vars;
+  std::random_device randomDevice;
+};
+
+TEST_F(RuntimeTest, SimpleTest) {
+  // SPIRV module embedded into the string.
+  // This module contains 4 resource variables devided into 2 sets.
```
kiszk: nit: `devided` -> `divided`

tools/mlir-vulkan-runner/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+  )
+
+if (MLIR_VULKAN_RUNNER_ENABLED)
+  message(STATUS "Building the Vulkan runner")
+
+  # At first try "FindVulkan" from:
+  # https://cmake.org/cmake/help/v3.7/module/FindVulkan.html
+  if (NOT CMAKE_VERSION VERSION_LESS 3.7.0)
+    find_package(Vulkan)
+  endif()
```
joker-eph: I'd add a else and a proper error message here.

unittests/Dialect/SPIRV/Runtime/RuntimeTests.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+// Struct containing the number of local workgroups to dispatch for each
+// dimension.
+struct NumWorkGroups {
+  uint32_t x{1};
+  uint32_t y{1};
+  uint32_t z{1};
+};
+
+// This is a temporary function and will be removed in the future.
```
joker-eph: Can you clarify the temporary aspect: there is a layering violation right now where you're poking at the implementation of a tools.

unittests/Dialect/SPIRV/Runtime/RuntimeTests.cpp
```diff
@@ -0,0 +1,235 @@
...
+  ASSERT_TRUE(succeeded(parseAndRunModule(spirvModuleSource, numWorkGroups)));
+  ASSERT_TRUE(isEqualFloat(expected, vars[1][1]));
+
+  destroyResourceVarFloat(vars[0][0]);
+  destroyResourceVarFloat(vars[0][1]);
+  destroyResourceVarFloat(vars[1][0]);
+  destroyResourceVarFloat(vars[1][1]);
+  destroyResourceVarFloat(fmulResult);
+  destroyResourceVarFloat(expected);
+}
```
joker-eph: Can we avoid C++ unit-tests and use a regular lit+FileCheck testing please? I think that is the whole point of using a runner.

See the cuda testing for reference.

tools/mlir-vulkan-runner/CMakeLists.txt
antiagainst: I think the error message is at L29. :) The probe is not done yet here.

tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,1029 @@
...
+using namespace mlir;
+
+using DescriptorSetIndex = uint32_t;
+using BindingIndex = uint32_t;
+
+/// Struct containing information regarding to a device memory buffer.
+struct VulkanDeviceMemoryBuffer {
+  BindingIndex bindingIndex{0};
+  VkDescriptorType descriptorType{VK_DESCRIPTOR_TYPE_MAX_ENUM};
+  VkDescriptorBufferInfo bufferInfo;
```
antiagainst: Zero initialize the other fields too? (Using `VK_NULL_HANDLE` and other reasonable defaults)

tools/mlir-vulkan-runner/VulkanRuntime.cpp
```diff
@@ -0,0 +1,1029 @@
...
+
+/// Struct containing the number of local workgroups to dispatch for each
+/// dimension.
+struct NumWorkGroups {
+  uint32_t x{1};
+  uint32_t y{1};
+  uint32_t z{1};
+};
+
+/// Struct containing information regarding to a descriptor set.
```
antiagainst: s/to//

tools/mlir-vulkan-runner/VulkanRuntime.cpp
antiagainst: s/spir-v/SPIR-V/

globally

unittests/Dialect/SPIRV/Runtime/RuntimeTests.cpp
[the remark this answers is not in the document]
antiagainst: No. Actually it's fine to have VulkanHostMemoryBuffer to hold raw pointer. I was just saying that we should wrap these raw pointers in `unique_ptr` to avoid potential leaks. For example, you can let this function to return the `unique_ptr` and assign it to a local variable in the test so we don't need to explicitly call `delete` later.

unittests/Dialect/SPIRV/Runtime/RuntimeTests.cpp
antiagainst: I think this was just a temporary step to have functionalities implemented incrementally. We don't have all the [shims](https://github.com/tensorflow/mlir/tree/master/tools/mlir-cuda-runner) and [conversions](https://github.com/tensorflow/mlir/blob/master/lib/Conversion/GPUToCUDA/ConvertLaunchFuncToCudaCalls.cpp) like the CUDA side yet. We've already >1k LOCs here; having all the above, we are expecting another >1k LOCs. So I'm fine of landing this as-is. But if you are very uncomfortable about this, I see two ways going forward: 1) remove the C++ tests and land the functionality; 2) leaving this PR open and have following-up commits to implement the rest of the functionality so finally we can squash the intermediate state away.

antiagainst (changes requested):
Sorry for the late reply! On my side I just have a few more nits.

antiagainst:
Re: https://github.com/tensorflow/mlir/pull/118#issuecomment-529228871

1. Okay, using `llvm::errs()` sgtm. :)
2. What you said is reasonable for me!

Again, sorry for the delay on reviewing! There are quite a few interesting things happened in the meantime that relates to this. Our prototyping on the Vulkan runtime side, [IREE](https://github.com/google/iree), was open sourced. (I see you've already noticed that because you posted questions there. ;-P) It is a reflection of how we actually want to approach in a more Vulkan-native way. Compared to CUDA, which provides a more developer-friendly and middle-level host API abstractions, Vulkan's host API is more low-level and much verbose. We think by modelling it in IR form (not literally but selectively on core features), we can gain the benefits of running compiler optimizations on it. It is not fully proven yet but looks quite promising thus far. We are also thinking about how to structure MLIR core and IREE regarding the components. SPIR-V dialect is in MLIR core right now, but the lowering from high-level dialect used by IREE is not; it's inside IREE. That is partially because we don't have a proper modelling of Vulkan host side in MLIR core. Here different from CUDA again, a simple `gpu.launch` op that connects the host and device code does not really work for Vulkan (I mean yes we can still lower it to Vulkan but it's not gaining all expected benefits from Vulkan performance-wise). So it would be nice to have proper modelling on Vulkan in core so we can have the lowering in core and potentially share with other code paths for functionalities. Just wanted to point out recent developments to keep you informed where we are going; what you've here is certainly very valuable for testing and running in core and as building blocks for future Vulkan modelling. Thanks for the contribution again! :)

unittests/Dialect/SPIRV/Runtime/RuntimeTests.cpp
joker-eph: I'm still on the same opinion as in September: build the right tool step by step with the right testing at every step along the way.

tools/mlir-vulkan-runner/CMakeLists.txt
joker-eph: I see that there is a fallback, so throwing error here is not possible indeed, but I still feel that the "you're not using the right CMake version should be something surfaced at some point as a root cause of failure.
Something like adding here: an ```else if ("$ENV{VULKAN_SDK}" STREQUAL "")  message(FATAL_ERROR "Please use at least Make 3.7.0 or provide the VULKAN_SDK path as an environment variable")```

joker-eph (commented):
Actually I just noticed that the PR wasn't updated by Denis recently, I got mislead because it popped up in my email box after recent comments from kiszk@

denis0x0D:
@kiszk @joker-eph @antiagainst @MaheshRavishankar thanks for review, actually I was thinking that IREE covers that PR as well, and this PR is not needed anymore, but if it is still relevant to test lowering - that sounds great to me!

joker-eph:
IREE is an independent project, I am not aware of a plan to integrate IREE within MLIR, so having end-to-end codegen story and testing  capability seems still important to me.

denis0x0D:
@joker-eph @antiagainst 

> having end-to-end codegen story and testing capability seems still important to me.

sounds great to me!

If it's ok, on the next iteration I'll rebase current patch on trunk, update according to the comments, delete c++ unit tests and leave this PR under WIP. 
After that I can start to work on a separate PR, the PR will implement a pass which:
1. Translates module into spv.module.
2. Serializes spv.module into binary form.
3. Sets serialized module as an attribute.
This is similar to GpuKernelToCubinPass does https://github.com/tensorflow/mlir/blob/master/lib/Conversion/GPUToCUDA/ConvertKernelFuncToCubin.cpp

Seems to me, to implement fully working mlir-vulkan-runner also requires one more PR to cover similar functionality with https://github.com/tensorflow/mlir/blob/master/lib/Conversion/GPUToCUDA/ConvertLaunchFuncToCudaCalls.cpp and to update runtime part to be consistent with new passes and mlir-vulkan-runner. I can cover it on the future PRs.
What do you think? If you see the better plan, please fix me in this case.
Thanks!

antiagainst:
+1 to what @joker-eph said in https://github.com/tensorflow/mlir/pull/118#issuecomment-557054792.

@denis0x0D: what you said SGTM! Just wanted to let you know, we are working on enabling the lowering flow from Linalg dialect to SPIR-V right now. I think that has a higher priority. We'll create issues for tasks soon and would really appreciate it if you can help there first. :) Sorry for keeping this thread open for even longer time, but I guess because this is entirely new code so it won't be much a problem to rebase against master branch later. Thanks!

denis0x0D:
@antiagainst 

> we are working on enabling the lowering flow from Linalg dialect to SPIR-V right now. I think that has a higher priority. We'll create issues for tasks soon and would really appreciate it if you can help there first.  

Sounds great, I would like to help on it if possible!
Thanks!

MaheshRavishankar:
Just catching up on all the discussion here now. Apologies for not paying more attention here.

As everyone agrees here, having a mlir-vulkan-runner in MLIR core is extremely useful for testing. So thanks again @denis0x0D for taking this up. Just to provide more context, as @antiagainst  mentioned we have been trying to get all the pieces in place to convert from Linalg ops to GPU dialect to SPIR-V dialect. Some of the changes w.r.t to these will land soon (in a couple of days). I have also been making some fairly significant change to the SPIR-V lowering infrastructure to make it easy to build upon. (these should also land in a couple of days). While these changes don't directly affect the development of a vulkan-runner (the runner should be only be concerned about the spir-v binary generated by the dialect), they do motivate some requirements on the vulkan-runner summarized below.

1) As structured right now, going to GPU dialect will result in two sub-modules within the main module

```
module {
   module attributes {gpu.container_module} {
      // Host side code which will contain gpu.launch_func op
  }
  module @fmul_kernel attributes {gpu.kernel_module} {
     // Kernel functions ,i.e. FuncOps with the gpu.kernel attribute
  }
}
```

The functions in the `gpu.kernel_module` with the `gpu.kernel` attribute are targeted for SPIR-V lowering. THe lowering will create a new `spv.module` within the `gpu.kernel_module`


```
module {
   module attributes {gpu.container_module} {
      // Host side code which will contain gpu.launch_func op
  }
  module @fmul_kernel attributes {gpu.kernel_module} {
    spv.module {
       ...
    }
     // Kernel functions ,i.e. FuncOps with the gpu.kernel attribute
  }
}
```

So it would be useful if the vulkan runner can handle such module. Given this I have some follow up below on a previous comment

> @joker-eph @antiagainst
> 
> > having end-to-end codegen story and testing capability seems still important to me.
> 
> sounds great to me!
> 
> If it's ok, on the next iteration I'll rebase current patch on trunk, update according to the comments, delete c++ unit tests and leave this PR under WIP.
> After that I can start to work on a separate PR, the PR will implement a pass which:
> 
> 1. Translates module into spv.module.

I am not sure we can translate a `module` into `spv.module` directly since the `module` contains both host and device side components. Do you mean extracting spv.module from within a module and running them?

> 2. Serializes spv.module into binary form.
> 3. Sets serialized module as an attribute.
>    This is similar to GpuKernelToCubinPass does https://github.com/tensorflow/mlir/blob/master/lib/Conversion/GPUToCUDA/ConvertKernelFuncToCubin.cpp
> 
> Seems to me, to implement fully working mlir-vulkan-runner also requires one more PR to cover similar functionality with https://github.com/tensorflow/mlir/blob/master/lib/Conversion/GPUToCUDA/ConvertLaunchFuncToCudaCalls.cpp and to update runtime part to be consistent with new passes and mlir-vulkan-runner. I can cover it on the future PRs.
> What do you think? If you see the better plan, please fix me in this case.
> Thanks!

joker-eph:
On Sat, Nov 23, 2019 at 11:45 PM MaheshRavishankar <notifications@github.com>
wrote:

> Just catching up on all the discussion here now. Apologies for not paying
> more attention here.
>
> As everyone agrees here, having a mlir-vulkan-runner in MLIR core is
> extremely useful for testing. So thanks again @denis0x0D
> <https://github.com/denis0x0D> for taking this up. Just to provide more
> context, as @antiagainst <https://github.com/antiagainst> mentioned we
> have been trying to get all the pieces in place to convert from Linalg ops
> to GPU dialect to SPIR-V dialect. Some of the changes w.r.t to these will
> land soon (in a couple of days). I have also been making some fairly
> significant change to the SPIR-V lowering infrastructure to make it easy to
> build upon. (these should also land in a couple of days). While these
> changes don't directly affect the development of a vulkan-runner (the
> runner should be only be concerned about the spir-v binary generated by the
> dialect), they do motivate some requirements on the vulkan-runner
> summarized below.
>
>    1. As structured right now, going to GPU dialect will result in two
>    sub-modules within the main module
>
> module {
>    module attributes {gpu.container_module} {
>       // Host side code which will contain gpu.launch_func op
>   }
>   module @fmul_kernel attributes {gpu.kernel_module} {
>      // Kernel functions ,i.e. FuncOps with the gpu.kernel attribute
>   }
> }
>
>
As far as I can tell, at the moment the host code is in the module
enclosing the kernel module, they aren't siblings:
https://github.com/tensorflow/mlir/blob/master/g3doc/Dialects/GPU.md#gpulaunch_func

Did I miss anything?


>  The functions in the gpu.kernel_module with the gpu.kernel attribute are
> targeted for SPIR-V lowering. THe lowering will create a new spv.module
> within the gpu.kernel_module
>
> module {
>    module attributes {gpu.container_module} {
>       // Host side code which will contain gpu.launch_func op
>   }
>   module @fmul_kernel attributes {gpu.kernel_module} {
>     spv.module {
>        ...
>     }
>      // Kernel functions ,i.e. FuncOps with the gpu.kernel attribute
>   }
> }
>
>
Why is the spv.module nested inside the kernel_module instead lowering the
kernel_module into a spv.module?
What do the "FuncOps with the gpu.kernel" attribute becomes in your
representation above?

(is there an example I could follow in the repo maybe?)

Thanks,

-- 
Mehdi

denis0x0D:
@MaheshRavishankar thanks for the feedback,

> I am not sure we can translate a module into spv.module directly since the module contains both host and device side components. Do you mean extracting spv.module from within a module and running them?

I was wrong about translate to spv.module inside this pass, sorry about it, I lost the context of this task :) 
The lowering part GPU -> SPIR-V is already exists and it should be added to pass manager in `mlir-vulkan-runner`.

So to enalbe `mlir-vulkan-runner` it needs:
1. A pass which enables right after "-convert-gpu-to-spirv", it consumes `module` with {gpu.container_module} attribute, serializes `spv.module` into binary form, using existing `spirv::Serializer` and attaches serialized `spv.module` as an attribute, simliar to this
https://github.com/tensorflow/mlir/blob/master/lib/Conversion/GPUToCUDA/ConvertKernelFuncToCubin.cpp#L85

2. A pass which instruments host part, the part which contains `gpu.launch_func`, with calls to runtime wrappers, simliar to this (the API should be discussed):
```
module {
  func @main() {
     // ...
    call @vkPopulateBuffer(%buff : memref<x?f32>, %desriptorSet : i32, %descriptorBinding : i32, %storageClass : i32) ...
    // ...
    call @vkLaunchShader(...) ...
    // ...
 }
// extern @vkPopulateBuffer
// extern @vkLaunchShader
```
The host part is lowering to llvm ir (llvm.dialect), the serialized module becomes a `llvm.global` https://github.com/tensorflow/mlir/blob/master/lib/Conversion/GPUToCUDA/ConvertLaunchFuncToCudaCalls.cpp#L321 with just a binary data, so we can pass a pointer to that data to runtime https://github.com/tensorflow/mlir/pull/118/files#diff-13d6d98d70fab37eda97625d19b84998R685, actually to create a shaderModule we just need to populate `VkShaderModuleCreateInfo` with size and pointer to binary shader.

So as result after the second pass we get a `module` which can be run under `JitRunner`, the pipeline must look similiar to this code https://github.com/tensorflow/mlir/blob/master/tools/mlir-cuda-runner/mlir-cuda-runner.cpp#L112.
Thanks!

MaheshRavishankar:
Thanks @denis0x0D . Overall what you are suggesting makes sense to me, but I am not that fluent in Vulkan speak to provide more feedback on specifics. But the interaction with the kernel to SPIR-V binary compilation makes sense.

@joker-eph : Thanks for catching my mistake. Edited my post to reflect the actual layout. The main point i was raising though was that you cannot serialize the entire module to a spv.module.

MaheshRavishankar:
> On Sat, Nov 23, 2019 at 11:45 PM MaheshRavishankar ***@***.***> wrote: Just catching up on all the discussion here now. Apologies for not paying more attention here. As everyone agrees here, having a mlir-vulkan-runner in MLIR core is extremely useful for testing. So thanks again @denis0x0D <https://github.com/denis0x0D> for taking this up. Just to provide more context, as @antiagainst <https://github.com/antiagainst> mentioned we have been trying to get all the pieces in place to convert from Linalg ops to GPU dialect to SPIR-V dialect. Some of the changes w.r.t to these will land soon (in a couple of days). I have also been making some fairly significant change to the SPIR-V lowering infrastructure to make it easy to build upon. (these should also land in a couple of days). While these changes don't directly affect the development of a vulkan-runner (the runner should be only be concerned about the spir-v binary generated by the dialect), they do motivate some requirements on the vulkan-runner summarized below. 1. As structured right now, going to GPU dialect will result in two sub-modules within the main module module { module attributes {gpu.container_module} { // Host side code which will contain gpu.launch_func op } module @fmul_kernel attributes {gpu.kernel_module} { // Kernel functions ,i.e. FuncOps with the gpu.kernel attribute } }
> As far as I can tell, at the moment the host code is in the module enclosing the kernel module, they aren't siblings: https://github.com/tensorflow/mlir/blob/master/g3doc/Dialects/GPU.md#gpulaunch_func Did I miss anything?
> The functions in the gpu.kernel_module with the gpu.kernel attribute are targeted for SPIR-V lowering. THe lowering will create a new spv.module within the gpu.kernel_module module { module attributes {gpu.container_module} { // Host side code which will contain gpu.launch_func op } module @fmul_kernel attributes {gpu.kernel_module} { spv.module { ... } // Kernel functions ,i.e. FuncOps with the gpu.kernel attribute } }
> Why is the spv.module nested inside the kernel_module instead lowering the kernel_module into a spv.module? What do the "FuncOps with the gpu.kernel" attribute becomes in your representation above? (is there an example I could follow in the repo maybe?) Thanks,

The translation from FuncOp with gpu.kernel to spv.module predates the existence of kernel_module. So when the change to include the FuncOp with gpu.kernel was added, the change to convert a kernel_module to spv.module was not looked into. It is more natural to translate a kernel_module to a spv.module, and should be fairly straight-forward to achieve. It's on my list of things to do.

MaheshRavishankar:
> On Sat, Nov 23, 2019 at 11:45 PM MaheshRavishankar ***@***.***> wrote: Just catching up on all the discussion here now. Apologies for not paying more attention here. As everyone agrees here, having a mlir-vulkan-runner in MLIR core is extremely useful for testing. So thanks again @denis0x0D <https://github.com/denis0x0D> for taking this up. Just to provide more context, as @antiagainst <https://github.com/antiagainst> mentioned we have been trying to get all the pieces in place to convert from Linalg ops to GPU dialect to SPIR-V dialect. Some of the changes w.r.t to these will land soon (in a couple of days). I have also been making some fairly significant change to the SPIR-V lowering infrastructure to make it easy to build upon. (these should also land in a couple of days). While these changes don't directly affect the development of a vulkan-runner (the runner should be only be concerned about the spir-v binary generated by the dialect), they do motivate some requirements on the vulkan-runner summarized below. 1. As structured right now, going to GPU dialect will result in two sub-modules within the main module module { module attributes {gpu.container_module} { // Host side code which will contain gpu.launch_func op } module @fmul_kernel attributes {gpu.kernel_module} { // Kernel functions ,i.e. FuncOps with the gpu.kernel attribute } }
> As far as I can tell, at the moment the host code is in the module enclosing the kernel module, they aren't siblings: https://github.com/tensorflow/mlir/blob/master/g3doc/Dialects/GPU.md#gpulaunch_func Did I miss anything?
> The functions in the gpu.kernel_module with the gpu.kernel attribute are targeted for SPIR-V lowering. THe lowering will create a new spv.module within the gpu.kernel_module module { module attributes {gpu.container_module} { // Host side code which will contain gpu.launch_func op } module @fmul_kernel attributes {gpu.kernel_module} { spv.module { ... } // Kernel functions ,i.e. FuncOps with the gpu.kernel attribute } }
> Why is the spv.module nested inside the kernel_module instead lowering the kernel_module into a spv.module? What do the "FuncOps with the gpu.kernel" attribute becomes in your representation above? (is there an example I could follow in the repo maybe?) Thanks,
> […](#)
> -- Mehdi

The conversion of a FuncOp with gpu.kernel attribute to spv.module predates the existence of kernel_module. It is pretty straight-forward (and would actually clean up the conversion a little bit) to lower a kernel_module to a spv.module. It is on my list of things to do. Will probably get to that in a day or so.

@denis0x0D : I will try to send the change to lower kernel_module to spv.module as a pull request to the mlir github. So that might be something to look out for w.r.t to the vulkan runner. Thanks!

denis0x0D:
@MaheshRavishankar 

>  I will try to send the change to lower kernel_module to spv.module as a pull request to the mlir github. So that might be something to look out for w.r.t to the vulkan runner. Thanks!

Sounds great to me! By the way, those convertions looks great https://github.com/tensorflow/mlir/tree/master/test/Conversion/GPUToSPIRV

antiagainst:
Thannks @MaheshRavishankar for the good catch! I was actually implicitly thinking that way and didn't notice that it is not obvious to everyone. :)

I think we are already creating the `spv.module` inside the top-level model, so not embedded in `gpu.kernel_module`: https://github.com/tensorflow/mlir/blob/d342ff9507f952697f537309f73746ae78394bd3/lib/Conversion/GPUToSPIRV/ConvertGPUToSPIRVPass.cpp#L48-L57
But yes going directly from `gpu.kernel_module` instead of `gpu.kernel` is cleaner and we can have multiple entry points in the same `spv.module`! (We need to do more verification along the way too.)

MaheshRavishankar:
@antiagainst  : In top of tree now it is added just before the FuncOp

https://github.com/tensorflow/mlir/blob/5614db039b28026b5fe38aa96a21c4e6835d352b/lib/Conversion/GPUToSPIRV/ConvertGPUToSPIRVPass.cpp#L57

But the correct way is to convert the entire module into a spv.Module.

@denis0x0D : The conversion from GPU to SPIRV is changed now with 8c152c5771d590fca9c6b64b0bc6e9563175efb8
The CL introduces attributes that you use to specify the ABI for entry point functions. So when you convert from GPU to SPIR-V dialect you add these attributes to the function arguments and function itself that describe various ABI related aspects (like storage class, descriptor_set, binding, etc for spv.globalVariables), as well as the workgroup size for the entry point function.
A later pass (just before serialization) will materialize this ABI.

denis0x0D:
@MaheshRavishankar thanks for mention this!

sherhut:
Late to the game and just wanted to comment that it would be really _awesome_ if we had a lowering for the host-side of the gpu dialect in the vulkan/SPIRV context, as well. I think by using a C++ library to implement a higher-level API should make the host-side code-generation part relatively straight forward. However, as far as I have been told, it still requires quite some code to write such C++ library.

I would not spend too much energy on the design of the C++ library's API and just build what makes sense for easy code generation. The host side of the dialect will likely evolve a fair bit so these abstractions won't be forever and the mlir-vulkan-runner is not meant to replace a real runtime, so simplicity is more important.

antiagainst:
> I would not spend too much energy on the design of the C++ library's API and just build what makes sense for easy code generation. 

That's the approach that has been taken in the current pull request. You'll need thousands of lines of code to bootstrap a basic Vulkan compute shader. Right now that's all hidden in a single entry point `runOnVulkan`.

PR #123, opened 2019-09-06
Set mlir-cpu-runner JIT codegen opt level correctly

bondhugula:
- the JIT codegen was being run at the default -O0 level; instead,
  propagate the opt level from the cmd line.

### DIFF
--- a/include/mlir/ExecutionEngine/ExecutionEngine.h
+++ b/include/mlir/ExecutionEngine/ExecutionEngine.h
@@ -72,13 +72,15 @@ public:
 
   /// Creates an execution engine for the given module.  If `transformer` is
   /// provided, it will be called on the LLVM module during JIT-compilation and
-  /// can be used, e.g., for reporting or optimization.
-  /// If `sharedLibPaths` are provided, the underlying JIT-compilation will open
-  /// and link the shared libraries for symbol resolution.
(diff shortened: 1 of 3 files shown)


include/mlir/ExecutionEngine/ExecutionEngine.h
```diff
@@ -72,13 +72,15 @@ class ExecutionEngine {
...
-  /// generated for the given module.
+  /// can be used, e.g., for reporting or optimization. `jitCodeGenOptLevel`,
+  /// when provided, is used as the optimization level for target code
+  /// generation. If `sharedLibPaths` are provided, the underlying
+  /// JIT-compilation will open and link the shared libraries for symbol
+  /// resolution. If `objectCache` is provided, JIT compiler will use it to
+  /// store the object generated for the given module.
   static llvm::Expected<std::unique_ptr<ExecutionEngine>> create(
       ModuleOp m, std::function<llvm::Error(llvm::Module *)> transformer = {},
+      Optional<unsigned> jitCodeGenOptLevel = llvm::None,
```
River707: Why is this not llvm::CodeGenOpt::Level?

bondhugula: There's no good reason. Fixed it now.

PR #126, opened 2019-09-08
add missing memref cast fold pattern for dim op

bondhugula:
- add missing canonicalization pattern to fold memref_cast + dim to
  dim (needed to propagate constant when folding a dynamic shape to
  a static one)

- also fix an outdated/inconsistent comment in StandardOps/Ops.td

### DIFF
--- a/test/Transforms/canonicalize.mlir
+++ b/test/Transforms/canonicalize.mlir
@@ -256,21 +256,25 @@ func @xor_self_tensor(%arg0: tensor<4x5xi32>) -> tensor<4x5xi32> {
 
 // CHECK-LABEL: func @memref_cast_folding
 func @memref_cast_folding(%arg0: memref<4 x f32>, %arg1: f32) -> f32 {
+  // CHECK-NOT: memref_cast
   %1 = memref_cast %arg0 : memref<4xf32> to memref<?xf32>
-
-  // CHECK-NEXT: %c0 = constant 0 : index
   %c0 = constant 0 : index
+  // CHECK-NOT: dim
+  %dim = dim %1, 0 : memref<? x f32>
+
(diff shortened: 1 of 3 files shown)


test/Transforms/canonicalize.mlir
```diff
@@ -261,16 +261,21 @@ func @memref_cast_folding(%arg0: memref<4 x f32>, %arg1: f32) -> f32 {
   // CHECK-NEXT: %c0 = constant 0 : index
   %c0 = constant 0 : index
 
+  %dim = dim %1, 0 : memref<? x f32>
+
+  // CHECK-NEXT: affine.load %arg0[3]
```
jpienaar: This is not passing presubmit testing (seems you are missing "constant 4" op in between previous check and next). Could you verify locally?

bondhugula: You are right (this was because my local copy had another pending pull request, and the further simplification was a combined effect of both). Fixed now.

bondhugula:
@jpienaar Anything further on this? Thanks!

joker-eph:
Something is not stable across toolchains, we see a different ordering between constant internally:

```
 // CHECK-NEXT: %c0 = constant 0 : index
                ^
<stdin>:117:2: note: 'next' match was here
 %c0 = constant 0 : index
 ^
<stdin>:115:27: note: previous match ended here
 func @memref_cast_folding(%arg0: memref<4xf32>, %arg1: f32) -> f32 {
                          ^
<stdin>:116:1: note: non-matching line after previous match is here
 %c4 = constant 4 : index
^

```

joker-eph:
Seems like we got it passing after all! Sorry for the delay

PR #129, opened 2019-09-11
Update CMakeLists.txt for shared library build

stephenneuendorffer:
There are several circular dependencies between libraries.  These
are explicitly marked as STATIC.  Note that add_llvm_library is redundant
with add_dependencies, so we don't have to have libraries in both places.


jpienaar:
Hey Stephen,

Could you verify if you've signed the CLA? 

Thanks,

Jacques

Issue #127, opened 2019-09-11
[spirv] Add OpControlBarrier and OpMemoryBarrier


antiagainst:
We need the definitions of these two ops and their (de)serialization.

antiagainst:
@denis0x0D are you willing to help on this? :)

denis0x0D:
@antiagainst yes, of course! Thanks a lot!

antiagainst:
This is done via #130. Thanks @denis0x0D!

PR #130, opened 2019-09-11
[spirv] Add OpControlBarrier and OpMemoryBarrier

denis0x0D:
This patch addresses https://github.com/tensorflow/mlir/issues/127 

ebnf:
`scope ::=` `"CrossDevice"` | `"Device"` | `"Workgroup"` | `"Subgroup"`
              | `"Invocation"`

    memory-semantics ::= `"None"` | `"Acquire"` | `"Release"`
                         | `"AcquireReleae"` | `"SequentiallyConsistent"`
                         | `"UniformMemory"`  | `"SubgroupMemory"`
                         | `"WorkgroupMemory"`  | `"CrossWorkgroupMemory"`
                         | `"AtomicCounterMemory"` | `"ImageMemory"`

    control-barrier-op ::= `spv.ControlBarrier` scope, scope, memory-semantics

    memory-barrier-op ::= `spv.MemoryBarrier` scope, memory-semantics

Example:
``` 
 spv.ControlBarrier "Workgroup", "Device", "Acquire|UniformMemory"
 spv.MemoryBarrier "Device", "Acquire|UniformMemory"

```
@antiagainst @MaheshRavishankar can you please, take a look on this?
Thanks!

### DIFF
--- a/include/mlir/Dialect/SPIRV/SPIRVBase.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVBase.td
@@ -148,6 +148,8 @@ def SPV_OC_OpFOrdLessThanEqual      : I32EnumAttrCase<"OpFOrdLessThanEqual", 188
 def SPV_OC_OpFUnordLessThanEqual    : I32EnumAttrCase<"OpFUnordLessThanEqual", 189>;
 def SPV_OC_OpFOrdGreaterThanEqual   : I32EnumAttrCase<"OpFOrdGreaterThanEqual", 190>;
 def SPV_OC_OpFUnordGreaterThanEqual : I32EnumAttrCase<"OpFUnordGreaterThanEqual", 191>;
+def SPV_OC_OpControlBarrier         : I32EnumAttrCase<"OpControlBarrier", 224>;
+def SPV_OC_OpMemoryBarrier          : I32EnumAttrCase<"OpMemoryBarrier", 225>;
 def SPV_OC_OpLoopMerge              : I32EnumAttrCase<"OpLoopMerge", 246>;
 def SPV_OC_OpLabel                  : I32EnumAttrCase<"OpLabel", 248>;
 def SPV_OC_OpBranch                 : I32EnumAttrCase<"OpBranch", 249>;
@@ -180,8 +182,9 @@ def SPV_OpcodeAttr :
       SPV_OC_OpFUnordLessThan, SPV_OC_OpFOrdGreaterThan, SPV_OC_OpFUnordGreaterThan,
       SPV_OC_OpFOrdLessThanEqual, SPV_OC_OpFUnordLessThanEqual,
       SPV_OC_OpFOrdGreaterThanEqual, SPV_OC_OpFUnordGreaterThanEqual,
-      SPV_OC_OpLoopMerge, SPV_OC_OpLabel, SPV_OC_OpBranch,
-      SPV_OC_OpBranchConditional, SPV_OC_OpReturn, SPV_OC_OpReturnValue
+      SPV_OC_OpControlBarrier, SPV_OC_OpMemoryBarrier, SPV_OC_OpLoopMerge,
+      SPV_OC_OpLabel, SPV_OC_OpBranch, SPV_OC_OpBranchConditional, SPV_OC_OpReturn,
+      SPV_OC_OpReturnValue
       ]> {
     let returnType = "::mlir::spirv::Opcode";
     let convertFromStorage = "static_cast<::mlir::spirv::Opcode>($_self.getInt())";
@@ -1031,6 +1034,52 @@ def SPV_MemoryModelAttr :
     ]> {
   let returnType = "::mlir::spirv::MemoryModel";
   let convertFromStorage = "static_cast<::mlir::spirv::MemoryModel>($_self.getInt())";
+  let cppNamespace = "::mlir::spirv";
+}
+
+def SPV_MS_None                   : BitEnumAttrCase<"None", 0x0000>;
+def SPV_MS_Acquire                : BitEnumAttrCase<"Acquire", 0x0002>;
+def SPV_MS_Release                : BitEnumAttrCase<"Release", 0x0004>;
+def SPV_MS_AcquireRelease         : BitEnumAttrCase<"AcquireRelease", 0x0008>;
+def SPV_MS_SequentiallyConsistent : BitEnumAttrCase<"SequentiallyConsistent", 0x0010>;
+def SPV_MS_UniformMemory          : BitEnumAttrCase<"UniformMemory", 0x0040>;
+def SPV_MS_SubgroupMemory         : BitEnumAttrCase<"SubgroupMemory", 0x0080>;
+def SPV_MS_WorkgroupMemory        : BitEnumAttrCase<"WorkgroupMemory", 0x0100>;
+def SPV_MS_CrossWorkgroupMemory   : BitEnumAttrCase<"CrossWorkgroupMemory", 0x0200>;
+def SPV_MS_AtomicCounterMemory    : BitEnumAttrCase<"AtomicCounterMemory", 0x0400>;
+def SPV_MS_ImageMemory            : BitEnumAttrCase<"ImageMemory", 0x0800>;
+def SPV_MS_OutputMemory           : BitEnumAttrCase<"OutputMemory", 0x1000>;
+def SPV_MS_MakeAvailable          : BitEnumAttrCase<"MakeAvailable", 0x2000>;
+def SPV_MS_MakeVisible            : BitEnumAttrCase<"MakeVisible", 0x4000>;
+def SPV_MS_Volatile               : BitEnumAttrCase<"Volatile", 0x8000>;
+
+def SPV_MemorySemanticsAttr :
+    BitEnumAttr<"MemorySemantics", "valid SPIR-V MemorySemantics", [
+      SPV_MS_None, SPV_MS_Acquire, SPV_MS_Release, SPV_MS_AcquireRelease,
+      SPV_MS_SequentiallyConsistent, SPV_MS_UniformMemory, SPV_MS_SubgroupMemory,
+      SPV_MS_WorkgroupMemory, SPV_MS_CrossWorkgroupMemory,
+      SPV_MS_AtomicCounterMemory, SPV_MS_ImageMemory, SPV_MS_OutputMemory,
+      SPV_MS_MakeAvailable, SPV_MS_MakeVisible, SPV_MS_Volatile
+    ]> {
+  let returnType = "::mlir::spirv::MemorySemantics";
+  let convertFromStorage = "static_cast<::mlir::spirv::MemorySemantics>($_self.getInt())";
+  let cppNamespace = "::mlir::spirv";
+}
+
+def SPV_S_CrossDevice : I32EnumAttrCase<"CrossDevice", 0>;
+def SPV_S_Device      : I32EnumAttrCase<"Device", 1>;
+def SPV_S_Workgroup   : I32EnumAttrCase<"Workgroup", 2>;
+def SPV_S_Subgroup    : I32EnumAttrCase<"Subgroup", 3>;
+def SPV_S_Invocation  : I32EnumAttrCase<"Invocation", 4>;
+def SPV_S_QueueFamily : I32EnumAttrCase<"QueueFamily", 5>;
+
+def SPV_ScopeAttr :
+    I32EnumAttr<"Scope", "valid SPIR-V Scope", [
+      SPV_S_CrossDevice, SPV_S_Device, SPV_S_Workgroup, SPV_S_Subgroup,
+      SPV_S_Invocation, SPV_S_QueueFamily
+    ]> {
+  let returnType = "::mlir::spirv::Scope";
+  let convertFromStorage = "static_cast<::mlir::spirv::Scope>($_self.getInt())";
   let cppNamespace = "::mlir::spirv";
 }
 
--- a/include/mlir/Dialect/SPIRV/SPIRVOps.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVOps.td
@@ -172,6 +172,72 @@ def SPV_CompositeExtractOp : SPV_Op<"CompositeExtract", [NoSideEffect]> {
 
 // -----
 
+def SPV_ControlBarrierOp : SPV_Op<"ControlBarrier", []> {
+  let summary = [{
+    Wait for other invocations of this module to reach the current point of
+    execution.
+  }];
+
+  let description = [{
+    All invocations of this module within Execution scope must reach this
+    point of execution before any invocation will proceed beyond it.
+
+    When Execution is Workgroup or larger, behavior is undefined if this
+    instruction is used in control flow that is non-uniform within
+    Execution. When Execution is Subgroup or Invocation, the behavior of
+    this instruction in non-uniform control flow is defined by the client
+    API.
+
+    If Semantics is not None, this instruction also serves as an
+    OpMemoryBarrier instruction, and must also perform and adhere to the
+    description and semantics of an OpMemoryBarrier instruction with the
+    same Memory and Semantics operands.  This allows atomically specifying
+    both a control barrier and a memory barrier (that is, without needing
+    two instructions). If Semantics is None, Memory is ignored.
+
+    Before version 1.3, it is only valid to use this instruction with
+    TessellationControl, GLCompute, or Kernel execution models. There is no
+    such restriction starting with version 1.3.
+
+    When used with the TessellationControl execution model, it also
+    implicitly synchronizes the Output Storage Class:  Writes to Output
+    variables performed by any invocation executed prior to a
+    OpControlBarrier will be visible to any other invocation after return
+    from that OpControlBarrier.
+
+    ### Custom assembly form
+
+    ``` {.ebnf}
+    scope ::= `"CrossDevice"` | `"Device"` | `"Workgroup"` | ...
+
+    memory-semantics ::= `"None"` | `"Acquire"` | "Release"` | ...
+
+    control-barrier-op ::= `spv.ControlBarrier` scope, scope, memory-semantics
+    ```
+
+    For example:
+
+    ```
+    spv.ControlBarrier "Workgroup", "Device", "Acquire|UniformMemory"
+
+    ```
+  }];
+
+  let arguments = (ins
+    SPV_ScopeAttr:$execution_scope,
+    SPV_ScopeAttr:$memory_scope,
+    SPV_MemorySemanticsAttr:$memory_semantics
+  );
+
+  let results = (outs);
+
+  let verifier = [{ return verifyMemorySemantics(*this); }];
+
+  let autogenSerialization = 0;
+}
+
+// -----
+
 def SPV_ExecutionModeOp : SPV_Op<"ExecutionMode", [InModuleScope]> {
   let summary = "Declare an execution mode for an entry point.";
 
@@ -262,6 +328,53 @@ def SPV_LoadOp : SPV_Op<"Load", []> {
   let results = (outs
     SPV_Type:$value
   );
+}
+
+// -----
+
+def SPV_MemoryBarrierOp : SPV_Op<"MemoryBarrier", []> {
+  let summary = "Control the order that memory accesses are observed.";
+
+  let description = [{
+    Ensures that memory accesses issued before this instruction will be
+    observed before memory accesses issued after this instruction. This
+    control is ensured only for memory accesses issued by this invocation
+    and observed by another invocation executing within Memory scope.
+
+    Semantics declares what kind of memory is being controlled and what kind
+    of control to apply.
+
+    To execute both a memory barrier and a control barrier, see
+    OpControlBarrier.
+
+    ### Custom assembly form
+
+    ``` {.ebnf}
+    scope ::= `"CrossDevice"` | `"Device"` | `"Workgroup"` | ...
+
+    memory-semantics ::= `"None"` | `"Acquire"` | `"Release"` | ...
+
+    memory-barrier-op ::= `spv.MemoryBarrier` scope, memory-semantics
+    ```
+
+    For example:
+
+    ```
+    spv.MemoryBarrier "Device", "Acquire|UniformMemory"
+
+    ```
+  }];
+
+  let arguments = (ins
+    SPV_ScopeAttr:$memory_scope,
+    SPV_MemorySemanticsAttr:$memory_semantics
+  );
+
+  let results = (outs);
+
+  let verifier = [{ return verifyMemorySemantics(*this); }];
+
+  let autogenSerialization = 0;
 }
 
 // -----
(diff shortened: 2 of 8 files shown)


include/mlir/Dialect/SPIRV/SPIRVBase.td
```diff
@@ -227,6 +230,20 @@ class SPV_Optional<Type type> : Variadic<type>;
...
+// executing invocations taking part in the operation.
+def SPV_Scope : IntOfWidths<[32]>;
+
+// Memory semantics define memory-order constraints, and on what storage classes
+// those constraints apply to. The memory order constrains the allowed orders in
+// which memory operations in this invocation can made visible to another
+// invocation. The storage classes specify to which subsets of memory these
+// constraints are to be applied. Storage classes not selected are not being
+// constrained.
+def SPV_MemorySemantics : IntOfWidths<[32]>;
```
MaheshRavishankar: Not sure about this approach. Should really be using an enum to describe all the valid semantics.. The enum would be an I32Attr (see other enum definitions in SPIRVBase.td). You could then let the operation take a list of these attributes (i.e. and ArrayAttr) and then having the instructions take an array of these

include/mlir/Dialect/SPIRV/SPIRVBase.td
```diff
@@ -227,6 +230,20 @@ class SPV_Optional<Type type> : Variadic<type>;
 // TODO(ravishankarm): From 1.4, this should also include Composite type.
 def SPV_SelectType : AnyTypeOf<[SPV_Scalar, SPV_Vector, SPV_AnyPtr]>;
 
+// The execution scope or memory scope of an operation. When used as a memory
+// scope, it specifies the distance of synchronization from the current
+// invocation. When used as an execution scope, it specifies the set of
+// executing invocations taking part in the operation.
+def SPV_Scope : IntOfWidths<[32]>;
```
MaheshRavishankar: ot sure about this approach. Should really be using an enum to describe all the valid semantics.. The enum would be an I32Attr (see other enum definitions in SPIRVBase.td). You could then let the operation take a list of these attributes (i.e. and ArrayAttr) and then having the instructions take an array of these

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -166,6 +166,68 @@ def SPV_CompositeExtractOp : SPV_Op<"CompositeExtract", [NoSideEffect]> {
...
+    %execScope = spv.constant 2 : i32
+    %memScope = spv.constant 1 : i32
+    %memSemantics = spv.constant 2376 : i32
+    spv.ControlBarrier %execScope, %memScope, %memSemantics
+
+    ```
+  }];
+
+  let arguments = (ins
+    SPV_Scope:$executionScope,
```
MaheshRavishankar: If the scope and memory semantics are attributes, then these would attributes as well.

include/mlir/Dialect/SPIRV/SPIRVBase.td
denis0x0D: @MaheshRavishankar thanks for review, since that mask it would require to describe all possible combination of those bits  https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Memory_Semantics_-id- to describe a type.
Also, regarding to the spec the operands must be a ssa-id of integer:
OpControlBarrier %int_2, %int_1, %int_2376 
it would require to creating of constants  while serizalizing.
If it's ok, I will update the patch.
Thanks.

antiagainst: Technically this memory semantics `<id>` can be generated from another instruction like `OpGenericPtrMemSemantics`. But that's for OpenCL. For Vulkan, which supports `Shader` capability, the spec [requires](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_shadervalidation_a_validation_rules_for_shader_a_href_capability_capabilities_a) "All `<id>` used for Scope and Memory Semantics must be of an OpConstant." So we should be fine here to embed this constant directly into the instruction as an attribute and provide better assembly sugar for it. So +1 to what Mehesh said. If later somebody would like to add support for OpenCL, we can then revisit this. :)

Generally we are missing a principled approach to these "bit enums". I'm thinking introducing something like `BitEnumAttr`. Similar to `Int32EnumAttr`, it describes each possible bit using a `BitEnumAttrCase`. Also similar to `EnumGen`, we can have a TableGen backend to generate conversion functions for them. It should generate conversion functions that take in a int32_t and returns a vector containing the names for each set bit, and vice versa. With that, it would be nice to have something like

```mlir
spv.ControlBarrier "Workgroup", "WorkGroup", "Acquire|UniformMemory"
```

or similar.

And this can be used for other SPIR-V enums like FunctionControl, etc. 

WDYT?

antiagainst:
Hey @denis0x0D, FYI, BitEnumAttr is available now after https://github.com/tensorflow/mlir/commit/f317907227b727e6c811f6a1e5d09336486b15ae. :)

denis0x0D:
@antiagainst thanks a lot!

denis0x0D:
@antiagainst can you please comment on following?
Regarding to spec the "Scope" type is integer that contains bit mask.
https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Scope_-id-
In this case we need some mechanism to disable autogen for AttrName definition in BitEnumAttr, for example for OpControlBarrier the same type of bit mask "Scope" used for the first operand and for the second. https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpControlBarrier
Auto gen generates by default "scope", so we get "redifinition error".
In other case, I was looking on examples in DirextXShaderCompiler and have not seen the uses-cases for that type with multiple bits enabled.
Can you please suggest, what is the better way, to make it as BitEnumAttr and update autogen utils or to use it as IntEnumAttr.
Thanks a lot!

denis0x0D:
@antiagainst @MaheshRavishankar I've updated the patch based on BitEnums, rebased on master and force pushed since it completely different version. Can you please take a look? Thanks a lot!

joker-eph:
That looks really nice! 
Nicer than the original one with the opaque constant and the SSA value which is still in the PR description (can you update the description by the way?)

denis0x0D:
@joker-eph thanks a lot for looking on this! Updated!

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -227,6 +227,33 @@ static LogicalResult verifyMemoryAccessAttribute(LoadStoreOpTy loadStoreOp) {
...
+    bitCount += static_cast<uint32_t>(
+        spirv::bitEnumContains(memorySemantics, memorySemanticsBit));
+  }
+
+  if (!bitCount) {
+    return op.emitError(
+        "expected at least one constraint for the memory order");
+  }
+
+  if (bitCount > 1) {
```
MaheshRavishankar: I am not sure of whether this is part of SPIR-V spec itself. It might be for specific use cases of SPIR-V, but if its not part of the spec itself, better to not have this validation.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -1523,6 +1553,29 @@ Serializer::processOp<spirv::ExecutionModeOp>(spirv::ExecutionModeOp op) {
...
+template <>
+LogicalResult
+Serializer::processOp<spirv::MemoryBarrierOp>(spirv::MemoryBarrierOp op) {
+  auto loc = op.getLoc();
+  auto memoryScopeIntAttr =
+      op.getAttrOfType<IntegerAttr>(spirv::attributeName<spirv::MemoryScope>());
+  auto memorySemanticsIntAttr = op.getAttrOfType<IntegerAttr>(
+      spirv::attributeName<spirv::MemorySemantics>());
+
+  auto memoryScopeID = prepareConstantInt(loc, memoryScopeIntAttr);
```
MaheshRavishankar: Same here. Maybe the processing of the memoryScope and memorySemantics can be commoned out?

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: @MaheshRavishankar thanks a lot for review, I was looking to this link: https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_memory_semantics__id_a_memory_semantics_lt_id_gt

> Despite being a mask and allowing multiple bits to be combined, it is invalid for more than one of these four bits to be set: Acquire, Release, AcquireRelease, or SequentiallyConsistent. Requesting both Acquire and Release semantics is done by setting the AcquireRelease bit, not by setting two bits.

> Memory semantics define memory-order constraints, and on what storage classes those constraints apply to. The memory order constrains the allowed orders in which memory operations in this invocation can made visible to another invocation. The storage classes specify to which subsets of memory these constraints are to be applied. Storage classes not selected are not being constrained.

Do I miss something?

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: @MaheshRavishankar thanks for mention this! I'll update it!

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -1910,6 +1965,46 @@ Deserializer::processOp<spirv::FunctionCallOp>(ArrayRef<uint32_t> operands) {
...
+    return emitError(unknownLoc, "OpMemoryBarrier memory semantics <id> ")
+           << memorySemanticsID << " must come from a normal constant";
+  }
+  auto memorySemantics =
+      memorySemanticsInfo->first.dyn_cast<IntegerAttr>();
+  if (!memorySemantics) {
+    return emitError(unknownLoc, "expected 32 bit integer constant for memory "
+                                 "semantics <id> ")
+           << memorySemanticsID << " in OpMemoryBarrier";
+  }
```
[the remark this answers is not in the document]
denis0x0D: @MaheshRavishankar thanks, I'll update it.

include/mlir/Dialect/SPIRV/SPIRVBase.td
```diff
@@ -1034,6 +1037,56 @@ def SPV_MemoryModelAttr :
...
+      SPV_MS_SequentiallyConsistent, SPV_MS_UniformMemory, SPV_MS_SubgroupMemory,
+      SPV_MS_WorkgroupMemory, SPV_MS_CrossWorkgroupMemory, SPV_MS_AtomicCounterMemory,
+      SPV_MS_ImageMemory
+    ]> {
+  let returnType = "::mlir::spirv::MemorySemantics";
+  let convertFromStorage = "static_cast<::mlir::spirv::MemorySemantics>($_self.getInt())";
+  let cppNamespace = "::mlir::spirv";
+}
+
+def SPV_S_CrossDevice : I32EnumAttrCase<"CrossDevice", 0x0000>;
```
antiagainst: Did you use the `utils/define_enum.sh` script to add these fields automatically from the spec? It seems strange here. IIRC, for `I32EnumAttr`, the value here should just be decimal values.

include/mlir/Dialect/SPIRV/SPIRVBase.td
```diff
@@ -1034,6 +1037,56 @@ def SPV_MemoryModelAttr :
...
+
+def SPV_MemorySemanticsAttr :
+    BitEnumAttr<"MemorySemantics", "valid SPIR-V MemorySemantics", [
+      SPV_MS_None, SPV_MS_Acquire, SPV_MS_Release, SPV_MS_AcquireRelease,
+      SPV_MS_SequentiallyConsistent, SPV_MS_UniformMemory, SPV_MS_SubgroupMemory,
+      SPV_MS_WorkgroupMemory, SPV_MS_CrossWorkgroupMemory, SPV_MS_AtomicCounterMemory,
+      SPV_MS_ImageMemory
+    ]> {
+  let returnType = "::mlir::spirv::MemorySemantics";
+  let convertFromStorage = "static_cast<::mlir::spirv::MemorySemantics>($_self.getInt())";
```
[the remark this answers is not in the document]
antiagainst: This is useful to generate getter method for a attribute of this kind.

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -172,6 +172,78 @@ def SPV_CompositeExtractOp : SPV_Op<"CompositeExtract", [NoSideEffect]> {
...
+    OpControlBarrier will be visible to any other invocation after return
+    from that OpControlBarrier.
+
+    ### Custom assembly form
+
+    ``` {.ebnf}
+    scope ::= `"CrossDevice"` | `"Device"` | `"Workgroup"` | `"Subgroup"`
+              | `"Invocation"`
+
+    memory-semantics ::= `"None"` | `"Acquire"` | `"Release"`
```
antiagainst: No need to list all of them (since this list will keep growing potentially). We can just list a few and then use `| ...` to omit the rest.

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -172,6 +172,78 @@ def SPV_CompositeExtractOp : SPV_Op<"CompositeExtract", [NoSideEffect]> {
...
+
+  let arguments = (ins
+    SPV_ExecutionScopeAttr:$execution_scope,
+    SPV_MemoryScopeAttr:$memory_scope,
+    SPV_MemorySemanticsAttr:$memory_semantics
+  );
+
+  let results = (outs);
+
+  let verifier = [{ return verifyMemorySemantics(*this); }];
```
antiagainst: I think this is just the default and can be removed?

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -172,6 +172,78 @@ def SPV_CompositeExtractOp : SPV_Op<"CompositeExtract", [NoSideEffect]> {
...
+
+  let autogenSerialization = 0;
+
```
antiagainst: Redundant empty line.

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -266,6 +338,59 @@ def SPV_LoadOp : SPV_Op<"Load", []> {
...
+    To execute both a memory barrier and a control barrier, see
+    OpControlBarrier.
+
+    ### Custom assembly form
+
+    ``` {.ebnf}
+    scope ::= `"CrossDevice"` | `"Device"` | `"Workgroup"` | `"Subgroup"`
+               | `"Invocation"`
+
+    memory-semantics ::= `"None"` | `"Acquire"` | `"Release"`
```
antiagainst: Same here.

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -266,6 +338,59 @@ def SPV_LoadOp : SPV_Op<"Load", []> {
...
+  }];
+
+  let arguments = (ins
+    SPV_MemoryScopeAttr:$memory_scope,
+    SPV_MemorySemanticsAttr:$memory_semantics
+  );
+
+  let results = (outs);
+
+  let verifier = [{ return verifyMemorySemantics(*this); }];
```
antiagainst: Same here.

include/mlir/Dialect/SPIRV/SPIRVOps.td
```diff
@@ -266,6 +338,59 @@ def SPV_LoadOp : SPV_Op<"Load", []> {
...
+
+  let autogenSerialization = 0;
+
```
antiagainst: Same here.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -227,6 +227,33 @@ static LogicalResult verifyMemoryAccessAttribute(LoadStoreOpTy loadStoreOp) {
...
+      spirv::MemorySemantics::AcquireRelease,
+      spirv::MemorySemantics::SequentiallyConsistent};
+
+  uint32_t bitCount = 0;
+  for (auto &memorySemanticsBit : memorySemanticsBits) {
+    bitCount += static_cast<uint32_t>(
+        spirv::bitEnumContains(memorySemantics, memorySemanticsBit));
+  }
+
+  if (!bitCount) {
```
antiagainst: Is this correct? What is the spec rule for this? I think we can have `None`/`Relaxed` memory semantics?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -227,6 +227,33 @@ static LogicalResult verifyMemoryAccessAttribute(LoadStoreOpTy loadStoreOp) {
...
 
+template <typename BarrierOp>
+static LogicalResult verifyMemorySemantics(BarrierOp op) {
+  auto memorySemantics = op.memory_semantics();
+  SmallVector<spirv::MemorySemantics, 4> memorySemanticsBits{
+      spirv::MemorySemantics::Acquire, spirv::MemorySemantics::Release,
+      spirv::MemorySemantics::AcquireRelease,
+      spirv::MemorySemantics::SequentiallyConsistent};
+
+  uint32_t bitCount = 0;
```
antiagainst: I think we can do ```
auto atMostOneInSet = spirv::MemorySemantics::... | ...;

auto bitCount = llvm::countPopulation(memorySemantics & atMostOneInSet);
```

We need to generate `operator&` for bit enums to do the above; but it should be fairly straightforward: https://github.com/tensorflow/mlir/blob/master/tools/mlir-tblgen/SPIRVUtilsGen.cpp#L707

File: lib/Dialect/SPIRV/SPIRVOps.cpp (hunk 4 above)
Comment (antiagainst) [reply lib/Dialect/SPIRV/SPIRVOps.cpp]: Yeah, this is from the spec.

**Comment by antiagainst** [comment]
Sorry for the delay! Regarding https://github.com/tensorflow/mlir/pull/130#issuecomment-532372947, I feel I don't quite understand here:

> In this case we need some mechanism to disable autogen for AttrName definition in BitEnumAttr, for example for OpControlBarrier the same type of bit mask "Scope" used for the first operand and for the second. https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpControlBarrier
Auto gen generates by default "scope", so we get "redifinition error".

Do you mean the auto-gen'd `attributeName`? I think it should be fine? Could you elaborate a bit?

**Comment by antiagainst** [comment]
I think we should just have one `ScopeAttr` definition and fix the autogen tools if they are causing problems. :)

File: lib/Dialect/SPIRV/SPIRVOps.cpp (hunk 4 above)
Comment (joker-eph) [reply lib/Dialect/SPIRV/SPIRVOps.cpp]: I think it would be nice to quote and refer to the spec everywhere appropriate as code comment in general. Some part of the clang parser are doing this and this is quite helpful reading code.

Comment (denis0x0D) [reply lib/Dialect/SPIRV/SPIRVOps.cpp]: @joker-eph I'll add the comment with link to the spec.

File: lib/Dialect/SPIRV/SPIRVOps.cpp (hunk 15 above)
Comment (denis0x0D) [reply lib/Dialect/SPIRV/SPIRVOps.cpp]: @antiagainst oh, I missed that none is relaxed, sorry about it. Thanks for mention this!

File: lib/Dialect/SPIRV/SPIRVOps.cpp (hunk 16 above)
Comment (denis0x0D) [reply lib/Dialect/SPIRV/SPIRVOps.cpp]: @antiagainst thanks, I'll update it!

File: include/mlir/Dialect/SPIRV/SPIRVOps.td (hunk 10 above)
Comment (denis0x0D) [reply include/mlir/Dialect/SPIRV/SPIRVOps.td]: @antiagainst can you please comment, do I get you right? I have to make `verifyMemorySemantics` code to be autogenerated?

**Comment by denis0x0D** [comment]
@antiagainst I was wrong about a special mechanism to disable autogen for attr name, sorry about it. I was relying on parseEnumAttribute function https://github.com/tensorflow/mlir/blob/master/lib/Dialect/SPIRV/SPIRVOps.cpp#L96 , it uses attr name generated by default for Enums https://github.com/tensorflow/mlir/blob/master/tools/mlir-tblgen/SPIRVUtilsGen.cpp#L660
, in case SPV_ScopeAttr, the name is `scope` and if we have the arguments like this:

```
let arguments = (ins
    SPV_ScopeAttr:$execution_scope,
    SPV_ScopeAttr:$memory_scope,
    SPV_MemorySemanticsAttr:$memory_semantics
  );
```
it parses and sets the attribute with default name which is `scope` and not  `execution_scope` or `memory_scope`, so just need to update a little bit a parseEnumAttribute function to allow to parse attribute with custom name.
Thanks.

**Comment by denis0x0D** [comment]
@antiagainst @MaheshRavishankar updated regarding to the comments, can you please take a look?
Thanks!

File: lib/Dialect/SPIRV/SPIRVOps.cpp [hunk 17]
```diff
@@ -229,23 +231,20 @@ static LogicalResult verifyMemoryAccessAttribute(LoadStoreOpTy loadStoreOp) {
 
 template <typename BarrierOp>
 static LogicalResult verifyMemorySemantics(BarrierOp op) {
+  // Regarding to the SPIR-V specification:
```
antiagainst: // According to ...
// "..."

We use `` for quoting code.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -132,6 +132,9 @@ class Deserializer {
   /// Gets the constant's attribute and type associated with the given <id>.
   Optional<std::pair<Attribute, Type>> getConstant(uint32_t id);
 
+  /// Gets the constants's integer attribute with the given <id>.
```
antiagainst: /// ... Returns a null IntegerAttr if the given <id> is not registered or does not correspond to an integer constant.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -1875,49 +1886,20 @@ Deserializer::processOp<spirv::ControlBarrierOp>(ArrayRef<uint32_t> operands) {
...
-  if (!memorySemanticsInfo) {
-    return emitError(unknownLoc, "OpControlBarrier memory semantics <id> ")
-           << memorySemanticsID << " must come from a normal constant";
-  }
-  auto memorySemantics = memorySemanticsInfo->first.dyn_cast<IntegerAttr>();
-  if (!memorySemantics) {
-    return emitError(unknownLoc, "expected 32 bit integer constant for memory "
-                                 "semantics <id> ")
-           << memorySemanticsID << " in OpControlBarrier";
+  SmallVector<IntegerAttr, 3> argAttrs;
```
antiagainst: Much nicer! Thanks!

include/mlir/Dialect/SPIRV/SPIRVOps.td
antiagainst: Nvm, this is fine.  Previously I meant that if we delete this line here, because of the SPV_Op definition, we will need to supply a `verify()` method for `spv::ControlBarrierOp` anyway. But that means `verify()` needs to  wrap `verifyMemorySemantics()` here.

denis0x0D:
@antiagainst thanks for review! Updated.

PR #134, opened 2019-09-14
Add integer sign- and zero-extension and truncation to standard.

ombre5733:
Currently sext, zext and trunc operations for integers are missing in the standard dialect.
If one wants to combine integers of different bit-widths, one has to use the LLVM operations
for extending/truncating and introduce custom casts to convert LLVM integer types to
standard types. Exposing sext, zext and trunc directly in the standard dialect allows to 
perfom integer conversions without this hassle.

Since this is my first addition to MLIR, I'd like some feedback for the implementation, especially:
- Naming of the operations and their memonics
- Are the tests sufficient as is or should I add more (if so, where)?

### DIFF
--- a/include/mlir/Dialect/StandardOps/Ops.td
+++ b/include/mlir/Dialect/StandardOps/Ops.td
@@ -821,7 +821,7 @@ def SelectOp : Std_Op<"select", [NoSideEffect, SameOperandsAndResultShape]> {
   let hasFolder = 1;
 }
 
-def SignExtendIOp : Std_Op<"sexti", [NoSideEffect]> {
+def SignExtendIOp : Std_Op<"sexti", [NoSideEffect, SameOperandsAndResultShape]> {
   let summary = "integer sign extension operation";
   let description = [{
     The integer sign extension operation takes an integer input of
@@ -990,7 +990,7 @@ def TensorStoreOp : Std_Op<"tensor_store",
   let verifier = ?;
 }
 
-def TruncateIOp : Std_Op<"trunci", [NoSideEffect]> {
+def TruncateIOp : Std_Op<"trunci", [NoSideEffect, SameOperandsAndResultShape]> {
   let summary = "integer truncation operation";
   let description = [{
     The integer truncation operation takes an integer input of
@@ -1031,7 +1031,7 @@ def XOrOp : IntArithmeticOp<"xor", [Commutative]> {
   let hasFolder = 1;
 }
 
-def ZeroExtendIOp : Std_Op<"zexti", [NoSideEffect]> {
+def ZeroExtendIOp : Std_Op<"zexti", [NoSideEffect, SameOperandsAndResultShape]> {
   let summary = "integer zero extension operation";
   let description = [{
     The integer zero extension operation takes an integer input of
--- a/include/mlir/Dialect/StandardOps/Ops.td
+++ b/include/mlir/Dialect/StandardOps/Ops.td
@@ -820,6 +820,45 @@ def SelectOp : Std_Op<"select", [NoSideEffect, SameOperandsAndResultShape]> {
 
   let hasFolder = 1;
 }
+
+def SignExtendIOp : Std_Op<"sexti", [NoSideEffect]> {
+  let summary = "integer sign extension operation";
+  let description = [{
+    The integer sign extension operation takes an integer input of
+    width M and an integer destination type of width N. The destination
+    bit-width must be larger than the input bit-width (N > M).
+    The top-most (N - M) bits of the output are filled with copies
+    of the most-significant bit of the input.
+
+      %1 = constant 5 : i3            // %1 is 0b101
+      %2 = sexti %1 : i3 to i6        // %2 is 0b111101
+      %3 = constant 2 : i3            // %3 is 0b010
+      %4 = sexti %3 : i3 to i6        // %4 is 0b000010
+  }];
+
+  let arguments = (ins IntegerLike:$value);
+
+  let results = (outs IntegerLike);
+
+  let builders = [OpBuilder<
+    "Builder *builder, OperationState *result, Value *value, Type destType", [{
+      result->addOperands(value);
+      result->addTypes(destType);
+  }]>];
+
+  let extraClassDeclaration = [{
+    /// Returns the result type of the sign extension operation.
+    IntegerType getType() { return getResult()->getType().cast<IntegerType>(); }
+  }];
+
+  let parser = [{
+    return impl::parseCastOp(parser, result);
+  }];
+  let printer = [{
+    return printStandardCastOp(this->getOperation(), p);
+  }];
+}
+
 def ShlISOp : IntArithmeticOp<"shlis"> {
   let summary = "signed integer shift left";
 }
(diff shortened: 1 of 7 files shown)


include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -820,6 +820,45 @@ def SelectOp : Std_Op<"select", [NoSideEffect, SameOperandsAndResultShape]> {
 
   let hasFolder = 1;
 }
+
+def SignExtendIOp : Std_Op<"sexti", [NoSideEffect]> {
```
joker-eph: I think you're need SameOperandsAndResultShape here?

(assuming you support more than scalar)

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -951,10 +990,82 @@ def TensorStoreOp : Std_Op<"tensor_store",
...
+      %3 = constant 2 : i3            // %3 is 0b010
+      %4 = zexti %3 : i3 to i6        // %4 is 0b000010
+  }];
+
+  let arguments = (ins IntegerLike:$value);
+
+  let results = (outs IntegerLike);
+
+  let builders = [OpBuilder<
+    "Builder *builder, OperationState *result, Value *value, Type destType", [{
```
joker-eph: There is already a builder generated with ` Type destType, Value *value`, do we need this extra one?

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -2021,6 +2021,24 @@ OpFoldResult SelectOp::fold(ArrayRef<Attribute> operands) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// SignExtendIOp
+//===----------------------------------------------------------------------===//
+
+static LogicalResult verify(SignExtendIOp op) {
+  auto opType = op.getOperand()->getType().dyn_cast_or_null<IntegerType>();
+  if (!opType)
+    return op.emitError("operand type ") << opType << " is not an integer ";
```
joker-eph: Because you use a constraint on the operand in ODS, there is already a verifier generated for you.
And the actual verifier is accepting Tensors and Vector (you can change the constraint to limit to only scalar if desirable).

You can always check the generated code by running:

`{build_path}/bin/mlir-tblgen -gen-op-defs {src_path}/include/mlir/Dialect/StandardOps/Ops.td -I {src_path}/include/ `

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -820,6 +820,45 @@ def SelectOp : Std_Op<"select", [NoSideEffect, SameOperandsAndResultShape]> {
...
+    The integer sign extension operation takes an integer input of
+    width M and an integer destination type of width N. The destination
+    bit-width must be larger than the input bit-width (N > M).
+    The top-most (N - M) bits of the output are filled with copies
+    of the most-significant bit of the input.
+
+      %1 = constant 5 : i3            // %1 is 0b101
+      %2 = sexti %1 : i3 to i6        // %2 is 0b111101
+      %3 = constant 2 : i3            // %3 is 0b010
+      %4 = sexti %3 : i3 to i6        // %4 is 0b000010
```
joker-eph: An example with a vector would be nice as well

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -2021,6 +2021,24 @@ OpFoldResult SelectOp::fold(ArrayRef<Attribute> operands) {
...
+  auto resType = op.getType().dyn_cast_or_null<IntegerType>();
+  if (!resType)
+    return op.emitError("result type ") << resType << " is not an integer ";
+  if (opType.getWidth() >= resType.getWidth())
+    return op.emitError("result type ")
+           << resType << " must be wider than operand type " << opType;
```
joker-eph: I think you need in the verifier:

1) Get the element type:
```
  Type dstEltType = getType();
  if (auto shapedType = dstEltType.dyn_cast<ShapedType>()) 
    dstEltType = shapedType.getElementType();
  Type srcEltType = getType();
  if (auto shapedType = srcEltType.dyn_cast<ShapedType>()) 
    srcEltType = shapedType.getElementType();
```

(I didn't find a helper for this, I think we should have one)

2) Forbid index type for destination (what about source?):

```
   if (dstEltType.isa<IndexType>()) 
      return op.emitError() << "index is not a valid destination type";
```

3) The bit-width comparison you're doing.

lib/Dialect/StandardOps/Ops.cpp
ombre5733: Thanks for this hint. Now I understood, how the verifiers are chained and which checks are already auto-generated.

lib/Dialect/StandardOps/Ops.cpp
ombre5733: Understood. The helper I've found is called `getElementTypeOrSelf()` from `IR/TypeUtilities.h`.

For now I'd like to forbid index as both source and destination type. Since the width of index is never stated explicitly, it can be that 

    zexti %0 : index to i64

works on one machine, where the index type is 32 bits wide but it would not when the index type is 64 (or even 128) bits. Treating `zexti` & co dynamically during lowering like `index_cast` (either sign extend or truncate) seems strange to me. I probably would not like that my `zexti` is converted to `llvm.trunc`.

Does this sound reasonable?

include/mlir/Dialect/StandardOps/Ops.td
ombre5733: Actually not.

include/mlir/Dialect/StandardOps/Ops.td
ombre5733: Agreed.

include/mlir/Dialect/StandardOps/Ops.td
ombre5733: Ok.

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -155,7 +156,8 @@ static void printStandardCastOp(Operation *op, OpAsmPrinter *p) {
 }
 
 /// A custom cast operation verifier.
-template <typename T> static LogicalResult verifyCastOp(T op) {
+template <typename T>
+static LogicalResult verifyCastOp(T op) {
```
joker-eph: our internal clang-format put this on a single line apparently

ombre5733: Do I need to react on this? Was just not sure because the issue has already the "ready to pull" label.

River707:
Can you update for the recent changes to head?

lib/Dialect/StandardOps/Ops.cpp
joker-eph: Sorry I missed your answer, yes this block our integration internally. If you had fixed it then it would have gone through (I could also have cloned and fix myself)

joker-eph: I could have fixed it myself by cloning and patching it, but indeed this was blocking the integration (after we mark it "ready to pull", the PRs are going through some tooling to integrate this internally before we push it out to the master branch again)

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -951,10 +987,76 @@ def TensorStoreOp : Std_Op<"tensor_store",
...
+      %1 = constant 5 : i3            // %1 is 0b101
+      %2 = zexti %1 : i3 to i6        // %2 is 0b000101
+      %3 = constant 2 : i3            // %3 is 0b010
+      %4 = zexti %3 : i3 to i6        // %4 is 0b000010
+
+      %5 = zexti %0 : vector<2 x i32> to vector<2 x i64>
+  }];
+
+  let arguments = (ins IntegerLike:$value);
+
```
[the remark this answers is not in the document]
ombre5733: Sorry, was not fast enough to react on this before the commit was merged. I'll do the cleaning when adding right shifts.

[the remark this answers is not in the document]
joker-eph: I fix them on integration (but I could have missed some)

PR #136, opened 2019-09-16
Upgrade/fix/simplify store to load forwarding

bondhugula:
- fix store to load forwarding for a certain set of cases (where
  forwarding shouldn't have happened); use AffineValueMap difference
  based MemRefAccess equality checking; utility logic is also greatly
  simplified

- add missing equality/inequality operators for AffineExpr ==/!= ints

- add == != operators on MemRefAccess

### DIFF
--- a/lib/Analysis/AffineStructures.cpp
+++ b/lib/Analysis/AffineStructures.cpp
@@ -2717,51 +2717,6 @@ void FlatAffineConstraints::projectOut(Value *id) {
   FourierMotzkinEliminate(pos);
 }
 
-bool FlatAffineConstraints::isRangeOneToOne(unsigned start,
-                                            unsigned limit) const {
-  assert(start <= getNumIds() - 1 && "invalid start position");
-  assert(limit > start && limit <= getNumIds() && "invalid limit");
-
-  FlatAffineConstraints tmpCst(*this);
-
-  if (start != 0) {
-    // Move [start, limit) to the left.
-    for (unsigned r = 0, e = getNumInequalities(); r < e; ++r) {
-      for (unsigned c = 0, f = getNumCols(); c < f; ++c) {
-        if (c >= start && c < limit)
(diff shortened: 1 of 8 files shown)


River707:
@andydavis1 Can you review?

lib/Analysis/AffineStructures.cpp
```diff
@@ -2717,51 +2717,6 @@ void FlatAffineConstraints::projectOut(Value *id) {
   FourierMotzkinEliminate(pos);
 }
 
-bool FlatAffineConstraints::isRangeOneToOne(unsigned start,
```
andydavis1: Was this just not used anywhere?

bondhugula: It just had one user and that's now gone with this update.

test/Transforms/memref-dataflow-opt.mlir
```diff
@@ -247,3 +247,35 @@ func @store_load_store_nested_fwd(%N : index) -> f32 {
...
+func @should_not_fwd(%A: memref<100xf32>, %M : index, %N : index) -> f32 {
+  %cf = constant 0.0 : f32
+  affine.store %cf, %A[%M] : memref<100xf32>
+  // CHECK: affine.load %{{.*}}[%{{.*}}]
+  %v = affine.load %A[%N] : memref<100xf32>
+  return %v : f32
+}
+
+// Can store forward to A[%j, %i], but no forwarding to load on %A[%i, %j]
+func @refs_not_known_to_be_equal(%A : memref<100 x 100 x f32>, %M : index) {
```
River707: Missing a CHECK-LABEL here.

bondhugula: Thanks, fixed.

bondhugula: I can't find any information on the failed check below.

Issue #131, opened 2019-09-12
[spirv] Add support for function calls


antiagainst:
Thus far the SPIR-V dialect is able to support multiple functions in a module and their (de)serialization. It would be nice to add `spv.FunctionCall` so we can make it really useful!

antiagainst:
This is done via #137. Thanks @denis0x0D!

PR #137, opened 2019-09-16
[spirv] Add support for function calls.

denis0x0D:
This patch is addressing https://github.com/tensorflow/mlir/issues/131

Thanks!

### DIFF
--- a/include/mlir/Dialect/SPIRV/SPIRVControlFlowOps.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVControlFlowOps.td
@@ -151,6 +151,52 @@ def SPV_BranchConditionalOp : SPV_Op<"BranchConditional", [Terminator]> {
 
 // -----
 
+def SPV_FunctionCallOp : SPV_Op<"FunctionCall", [InFunctionScope]> {
+  let summary = "Call a function.";
+
+  let description = [{
+    Result Type is the type of the return value of the function. It must be
+    the same as the Return Type operand of the Function Type operand of the
+    Function operand.
+
+    Function is an OpFunction instruction.  This could be a forward
+    reference.
+
+    Argument N is the object to copy to parameter N of Function.
+
+    Note: A forward call is possible because there is no missing type
+    information: Result Type must match the Return Type of the function, and
+    the calling argument types must match the formal parameter types.
+
+    ### Custom assembly form
+
+    ``` {.ebnf}
+    function-call-op ::= `spv.FunctionCall` function-id `(` ssa-use-list `)`
+                     `:` function-type
+    ```
+
+    For example:
+
+    ```
+    spv.FunctionCall @f_void(%arg0) : (i32) ->  ()
+    %0 = spv.FunctionCall @f_iadd(%arg0, %arg1) : (i32, i32) -> i32
+    ```
+  }];
+
+  let arguments = (ins
+    SymbolRefAttr:$callee,
+    Variadic<SPV_Type>:$arguments
+  );
+
+  let results = (outs
+    SPV_Optional<SPV_Type>:$result
+  );
+
+  let autogenSerialization = 0;
+}
+
+// -----
+
 def SPV_LoopOp : SPV_Op<"loop"> {
   let summary = "Define a structured loop.";
 
--- a/lib/Dialect/SPIRV/SPIRVOps.cpp
+++ b/lib/Dialect/SPIRV/SPIRVOps.cpp
@@ -35,6 +35,7 @@ using namespace mlir;
 // TODO(antiagainst): generate these strings using ODS.
 static constexpr const char kAlignmentAttrName[] = "alignment";
 static constexpr const char kBranchWeightAttrName[] = "branch_weights";
+static constexpr const char kCallee[] = "callee";
 static constexpr const char kDefaultValueAttrName[] = "default_value";
 static constexpr const char kFnNameAttrName[] = "fn";
 static constexpr const char kIndicesAttrName[] = "indices";
@@ -910,6 +911,108 @@ static void print(spirv::ExecutionModeOp execModeOp, OpAsmPrinter *printer) {
   interleaveComma(
       values.getValue().cast<ArrayAttr>(), *printer,
       [&](Attribute a) { *printer << a.cast<IntegerAttr>().getInt(); });
+}
+
+//===----------------------------------------------------------------------===//
+// spv.FuncionCall
+//===----------------------------------------------------------------------===//
+
+static ParseResult parseFunctionCallOp(OpAsmParser *parser,
+                                       OperationState *state) {
+  SymbolRefAttr calleeAttr;
+  FunctionType type;
+  SmallVector<OpAsmParser::OperandType, 4> operands;
+  auto loc = parser->getNameLoc();
+  if (parser->parseAttribute(calleeAttr, kCallee, state->attributes) ||
+      parser->parseOperandList(operands, OpAsmParser::Delimiter::Paren) ||
+      parser->parseColonType(type)) {
+    return failure();
+  }
+
+  auto funcType = type.dyn_cast<FunctionType>();
+  if (!funcType) {
+    return parser->emitError(loc, "expected function type, but provided ")
+           << type;
+  }
+
+  if (funcType.getNumResults() > 1) {
+    return parser->emitError(loc, "expected callee function to have 0 or 1 "
+                                  "result, but provided ")
+           << funcType.getNumResults();
+  }
+
+  return failure(parser->addTypesToList(funcType.getResults(), state->types) ||
+                 parser->resolveOperands(operands, funcType.getInputs(), loc,
+                                         state->operands));
+}
+
+static void print(spirv::FunctionCallOp functionCallOp, OpAsmPrinter *printer) {
+  SmallVector<Type, 4> argTypes(functionCallOp.getOperandTypes());
+  SmallVector<Type, 1> resultTypes(functionCallOp.getResultTypes());
+  Type functionType =
+      FunctionType::get(argTypes, resultTypes, functionCallOp.getContext());
+
+  *printer << spirv::FunctionCallOp::getOperationName() << ' '
+           << functionCallOp.getAttr(kCallee) << '(';
+  printer->printOperands(functionCallOp.arguments());
+  *printer << ") : " << functionType;
+}
+
+static LogicalResult verify(spirv::FunctionCallOp functionCallOp) {
+  auto fnName = functionCallOp.callee();
+
+  auto moduleOp = functionCallOp.getParentOfType<spirv::ModuleOp>();
+  if (!moduleOp) {
+    return functionCallOp.emitOpError(
+        "must appear in a function inside 'spv.module'");
+  }
+
+  auto funcOp = moduleOp.lookupSymbol<FuncOp>(fnName);
+  if (!funcOp) {
+    return functionCallOp.emitOpError("callee function '")
+           << fnName << "' not found in 'spv.module'";
+  }
+
+  auto functionType = funcOp.getType();
+
+  if (functionCallOp.getNumResults() > 1) {
+    return functionCallOp.emitOpError(
+               "expected callee function to have 0 or 1 result, but provided ")
+           << functionCallOp.getNumResults();
+  }
+
+  if (functionType.getNumInputs() != functionCallOp.getNumOperands()) {
+    return functionCallOp.emitOpError(
+               "has incorrect number of operands for callee: expected ")
+           << functionType.getNumInputs() << ", but provided "
+           << functionCallOp.getNumOperands();
+  }
+
+  for (uint32_t i = 0, e = functionType.getNumInputs(); i != e; ++i) {
+    if (functionCallOp.getOperand(i)->getType() != functionType.getInput(i)) {
+      return functionCallOp.emitOpError(
+                 "operand type mismatch: expected operand type ")
+             << functionType.getInput(i) << ", but provided "
+             << functionCallOp.getOperand(i)->getType()
+             << " for operand number " << i;
+    }
+  }
+
+  if (functionType.getNumResults() != functionCallOp.getNumResults()) {
+    return functionCallOp.emitOpError(
+               "has incorrect number of results has for callee: expected ")
+           << functionType.getNumResults() << ", but provided "
+           << functionCallOp.getNumResults();
+  }
+
+  if (functionCallOp.getNumResults() &&
+      (functionCallOp.getResult(0)->getType() != functionType.getResult(0))) {
+    return functionCallOp.emitOpError("result type mismatch: expected ")
+           << functionType.getResult(0) << ", but provided "
+           << functionCallOp.getResult(0)->getType();
+  }
+
+  return success();
 }
 
 //===----------------------------------------------------------------------===//
--- a/lib/Dialect/SPIRV/Serialization/Deserializer.cpp
+++ b/lib/Dialect/SPIRV/Serialization/Deserializer.cpp
@@ -128,6 +128,11 @@ private:
   /// Gets the constant's attribute and type associated with the given <id>.
   Optional<std::pair<Attribute, Type>> getConstant(uint32_t id);
 
+  /// Returns a symbol to be used for the function name with the given
+  /// result <id>. This tries to use the function's OpName if
+  /// exists; otherwise creates one based on the <id>.
+  std::string getFunctionSymbol(uint32_t id);
+
   /// Returns a symbol to be used for the specialization constant with the given
   /// result <id>. This tries to use the specialization constant's OpName if
   /// exists; otherwise creates one based on the <id>.
@@ -637,10 +642,7 @@ LogicalResult Deserializer::processFunction(ArrayRef<uint32_t> operands) {
            << functionType << " and return type " << resultType << " specified";
   }
 
-  std::string fnName = nameMap.lookup(operands[1]).str();
-  if (fnName.empty()) {
-    fnName = "spirv_fn_" + std::to_string(operands[2]);
-  }
+  std::string fnName = getFunctionSymbol(operands[1]);
   auto funcOp = opBuilder.create<FuncOp>(unknownLoc, fnName, functionType,
                                          ArrayRef<NamedAttribute>());
   curFunction = funcMap[operands[1]] = funcOp;
@@ -760,6 +762,14 @@ Optional<std::pair<Attribute, Type>> Deserializer::getConstant(uint32_t id) {
   if (constIt == constantMap.end())
     return llvm::None;
   return constIt->getSecond();
+}
+
+std::string Deserializer::getFunctionSymbol(uint32_t id) {
+  auto funcName = nameMap.lookup(id).str();
+  if (funcName.empty()) {
+    funcName = "spirv_fn_" + std::to_string(id);
+  }
+  return funcName;
 }
 
 std::string Deserializer::getSpecConstantSymbol(uint32_t id) {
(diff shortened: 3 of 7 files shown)


lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -517,8 +547,80 @@ Serializer::processMemberDecoration(uint32_t structID, uint32_t memberIndex,
...
+             << ", the size in words of the `binary` vector "
+             << functions.size();
+    }
+  }
+  return success();
+}
 } // namespace
 
+LogicalResult Serializer::processFunctionCallOp(spirv::FunctionCallOp op) {
+  auto funcAttrSymbolRef = op.getAttrOfType<SymbolRefAttr>("callee");
```
antiagainst: `op.callee()` ?

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -517,8 +547,80 @@ Serializer::processMemberDecoration(uint32_t structID, uint32_t memberIndex,
...
+
+  auto funcName = funcAttrSymbolRef.getValue();
+  auto funcID = findFunctionID(funcName);
+  auto funcCallID = getNextID();
+  SmallVector<uint32_t, 3> operands{resTypeID, funcCallID, funcID};
+  uint32_t argIndex = 0;
+
+  for (auto *value : op.arguments()) {
+    auto valueID = findValueID(value);
+    if (!valueID) {
```
antiagainst: The IR is verified so this cannot happen? I think we can turn it into an `assert` here.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -517,8 +547,80 @@ Serializer::processMemberDecoration(uint32_t structID, uint32_t memberIndex,
...
+    }
+    operands.push_back(valueID);
+    ++argIndex;
+  }
+
+  if (resultTypes.size()) {
+    valueIDMap[op.getResult(0)] = funcCallID;
+  }
+
+  // The spec allows to invoke FunctionCallOp as a forward call.
```
antiagainst: Instead of doing this, what about having a `getOrCreateFunctionID(StringRef)` function, which internally uses `getNextID` as the ID for that function if it is not seen yet? I think that will be simpler.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -1374,6 +1476,9 @@ LogicalResult Serializer::processOperation(Operation *op) {
   if (auto constOp = dyn_cast<spirv::ConstantOp>(op)) {
     return processConstantOp(constOp);
   }
+  if (auto funcCallOp = dyn_cast<spirv::FunctionCallOp>(op)) {
```
antiagainst: Similarly, if we do `processOp<spirv::FunctionCallOp>` we don't need to explicitly register here.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -517,8 +547,80 @@ Serializer::processMemberDecoration(uint32_t structID, uint32_t memberIndex,
...
+  if (failed(processType(op.getLoc(),
+                         (resultTypes.empty() ? getVoidType() : resultTypes[0]),
+                         resTypeID))) {
+    return failure();
+  }
+
+  auto funcName = funcAttrSymbolRef.getValue();
+  auto funcID = findFunctionID(funcName);
+  auto funcCallID = getNextID();
+  SmallVector<uint32_t, 3> operands{resTypeID, funcCallID, funcID};
```
antiagainst: 8 instead of 3 to avoid allocation for the common cases?

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -517,8 +547,80 @@ Serializer::processMemberDecoration(uint32_t structID, uint32_t memberIndex,
...
+    if (!valueID) {
+      return op.emitOpError("cannot find a value for the "
+                            "argument with number ")
+             << argIndex;
+    }
+    operands.push_back(valueID);
+    ++argIndex;
+  }
+
+  if (resultTypes.size()) {
```
antiagainst: !resultTypes.empty()

test/Dialect/SPIRV/control-flow-ops.mlir
```diff
@@ -144,6 +144,119 @@ func @weights_cannot_both_be_zero() -> () {
...
+  func @fmain(%arg0 : vector<4xf32>, %arg1 : vector<4xf32>, %arg2 : i32) -> () {
+    // CHECK: {{%.*}} = spv.FunctionCall @f_0({{%.*}}, {{%.*}}) : (vector<4xf32>, vector<4xf32>) -> vector<4xf32>
+    %0 = spv.FunctionCall @f_0(%arg0, %arg1) : (vector<4xf32>, vector<4xf32>) -> vector<4xf32>
+    // CHECK: spv.FunctionCall @f_1({{%.*}}, {{%.*}}) : (vector<4xf32>, vector<4xf32>) -> ()
+    spv.FunctionCall @f_1(%0, %arg1) : (vector<4xf32>, vector<4xf32>) ->  ()
+    // CHECK: spv.FunctionCall @f_2() : () -> ()
+    spv.FunctionCall @f_2() : () -> ()
+    // CHECK: {{%.*}} = spv.FunctionCall @f_4({{%.*}}) : (i32) -> i32
+    %1 = spv.FunctionCall @f_4(%arg2) : (i32) -> i32
+    // CHECK: {{%.*}} = spv.FunctionCall @f_4({{%.*}}) : (i32) -> i32
```
antiagainst: What's the purpose of this check? Seems the same as above.

test/Dialect/SPIRV/control-flow-ops.mlir
```diff
@@ -144,6 +144,119 @@ func @weights_cannot_both_be_zero() -> () {
...
+  func @f_1(%arg0 : vector<4xf32>, %arg1 : vector<4xf32>) -> () {
+    spv.Return
+  }
+
+  func @f_2() -> () {
+    spv.Return
+  }
+
+  func @f_3(%arg0 : !spv.struct<!spv.array<4xf32>>) -> (!spv.struct<!spv.array<4xf32>>) {
+    // CHEKC: {{%.*}} = spv.FunctionCall @f_3({{%.*}}) : (!spv.struct<!spv.array<4xf32>>) -> !spv.struct<!spv.array<4xf32>>
```
antiagainst: Typo: CHECK.

Do we need this case? Seems already covered by fmain. (I mean we don't check new stuff here right?)

test/Dialect/SPIRV/Serialization/function_call.mlir
```diff
@@ -0,0 +1,68 @@
...
+    %1 = spv._address_of @var1 : !spv.ptr<!spv.array<4xf32>, Input>
+    // CHECK: {{%.*}} = spv.FunctionCall @f_0({{%.*}}) : (i32) -> i32
+    %3 = spv.FunctionCall @f_0(%0) : (i32) -> i32
+    // CHECK: spv.FunctionCall @f_1({{%.*}}, {{%.*}}) : (i32, !spv.ptr<!spv.array<4 x f32>, Input>) -> ()
+    spv.FunctionCall @f_1(%3, %1) : (i32, !spv.ptr<!spv.array<4xf32>, Input>) ->  ()
+    // CHECK: spv.FunctionCall @f_2() : () -> ()
+    spv.FunctionCall @f_2() : () -> ()
+    // CHECK: spv.FunctionCall @f_3({{%.*}}) : (!spv.ptr<!spv.array<4 x f32>, Input>) -> !spv.ptr<!spv.array<4 x f32>, Input>
+    %4 = spv.FunctionCall @f_3(%1) : (!spv.ptr<!spv.array<4xf32>, Input>) -> !spv.ptr<!spv.array<4xf32>, Input>
+    // CHECK: spv.FunctionCall @f_3({{%.*}}) : (!spv.ptr<!spv.array<4 x f32>, Input>) -> !spv.ptr<!spv.array<4 x f32>, Input>
```
antiagainst: Do we need this case? It doesn't cover something new compared to the above one.

test/Dialect/SPIRV/Serialization/function_call.mlir
```diff
@@ -0,0 +1,68 @@
...
+    spv.Return
+  }
+  func @f_0(%arg0 : i32) -> i32 {
+    spv.ReturnValue %arg0 : i32
+  }
+  func @f_1(%arg0 : i32, %arg1 : !spv.ptr<!spv.array<4xf32>, Input>) -> () {
+    spv.Return
+  }
+  func @f_2() -> () {
+    // CHECK: spv.FunctionCall @f_2() : () -> ()
```
antiagainst: This one is not needed. It's already covered in fmain.

test/Dialect/SPIRV/Serialization/function_call.mlir
```diff
@@ -0,0 +1,68 @@
...
+    spv.FunctionCall @f_2() : () -> ()
+    spv.Return
+  }
+  func @f_3(%arg0 : !spv.ptr<!spv.array<4xf32>, Input>) -> !spv.ptr<!spv.array<4xf32>, Input> {
+    spv.ReturnValue %arg0 : !spv.ptr<!spv.array<4xf32>, Input>
+  }
+
+  func @f_4() -> () {
+    %count = spv.constant 16 : i32
+    // CHECK: spv.FunctionCall @f_loop_with_function_call({{%.*}}) : (i32) -> ()
```
antiagainst: Don't need this one? It's already covered in fmain.

antiagainst (changes requested):
Awesome! Thanks for adding support for this, Denis!

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: oh, I see, thanks a lot!

test/Dialect/SPIRV/Serialization/function_call.mlir
denis0x0D: I was thinking to check, that we can pass a result value from FunctionCall as operand for the other FunctionCall.

test/Dialect/SPIRV/Serialization/function_call.mlir
denis0x0D: I was thinking to check that Function call can work under control flow ops such as loops.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -912,6 +912,116 @@ static void print(spirv::ExecutionModeOp execModeOp, OpAsmPrinter *printer) {
...
+      parser->resolveOperands(operands, funcType.getInputs(), loc,
+                              state->operands)) {
+    return failure();
+  }
+
+  return success();
+}
+
+static void print(spirv::FunctionCallOp functionCallOp, OpAsmPrinter *printer) {
+  SmallVector<Type, 0> argTypes(functionCallOp.getOperandTypes());
```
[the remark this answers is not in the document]
denis0x0D: @antiagainst can you please clarify, it seems like it uses under L962, we have to print function type, please fix me if I'm wrong.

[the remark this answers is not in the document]
antiagainst: Ah, right. Sorry, my bad.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -500,6 +503,70 @@ LogicalResult Deserializer::processExtension(ArrayRef<uint32_t> operands) {
...
+    resultTypes.push_back(resultType);
+  }
+
+  // Create and insert spv.FunctionCall.
+  auto opFunctionCall = opBuilder.create<spirv::FunctionCallOp>(
+      unknownLoc, resultTypes, opBuilder.getSymbolRefAttr(functionName),
+      arguments);
+
+  if (resultTypes.size()) {
+    valueMap[resultID] = opFunctionCall.getResult(0);
```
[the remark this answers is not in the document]
antiagainst: Yes, you are correct. Please ignore this comment. :)

test/Dialect/SPIRV/Serialization/function_call.mlir
antiagainst: The result of a function call is a normal SSA value, so nothing different from other SSA values. I don't think we need to check it. :)

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -912,6 +912,116 @@ static void print(spirv::ExecutionModeOp execModeOp, OpAsmPrinter *printer) {
...
+           << type;
+  }
+
+  if (funcType.getNumResults() > 1) {
+    return parser->emitError(
+               loc, "callee function must have 0 or 1 result, but provided ")
+           << funcType.getNumResults();
+  }
+
+  if (parser->addTypesToList(funcType.getResults(), state->types) ||
```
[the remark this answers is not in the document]
denis0x0D: @antiagainst can you please clarify, sorry, I probably didn't get it right, parser returns an instance of ParseResult, logical operator || returns bool, I, probably missing the convertion from bool to ParseResult.

test/Dialect/SPIRV/Serialization/function_call.mlir
antiagainst: But that's for the next function?

lib/Dialect/SPIRV/SPIRVOps.cpp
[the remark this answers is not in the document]
River707: failure(parser->... || parser->...)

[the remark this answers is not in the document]
antiagainst: Sorry I meant `return failure(parser->... || parser->...)`.

test/Dialect/SPIRV/Serialization/function_call.mlir
denis0x0D: @antiagainst it's just check fro my first invalid implementation, I was missing to put and id of OpFunctionCall into id valueMap for serializer. I'll delete, thanks!

lib/Dialect/SPIRV/SPIRVOps.cpp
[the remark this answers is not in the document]
denis0x0D: got it! Thanks!

test/Dialect/SPIRV/Serialization/function_call.mlir
antiagainst: Oh, I see. In that case, you can `spv.Return %3 : i32` or something. No need to create another `spv.FunctionCall` just to use the result of a `spv.FunctionCall`. :)

include/mlir/Dialect/SPIRV/SPIRVControlFlowOps.td
```diff
@@ -151,6 +151,53 @@ def SPV_BranchConditionalOp : SPV_Op<"BranchConditional", [Terminator]> {
 
 // -----
 
+def SPV_FunctionCallOp : SPV_Op<"FunctionCall"> {
```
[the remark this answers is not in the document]
denis0x0D: @antiagainst in this case should InFunctionScope be changed 
from 
`CPred<"llvm::isa_and_nonnull<FuncOp>($_op.getParentOp())">>;`
to something like this
`CPred<"llvm::isa_and_nonnull<FuncOp>($_op.getParentOfType<FuncOp>())">>;`

If we put op with this trait inside the loop op, we will get the LoopOp as parent, please fix me if I'm wrong.

lib/Dialect/SPIRV/Serialization/Serializer.cpp
MaheshRavishankar: +1

MaheshRavishankar (commented):
Thanks a lot @denis0x0D ! This is really useful

include/mlir/Dialect/SPIRV/SPIRVControlFlowOps.td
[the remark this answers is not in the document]
denis0x0D: Updated:
`CPred<"($_op.getParentOfType<FuncOp>())">>;`

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -912,6 +912,116 @@ static void print(spirv::ExecutionModeOp execModeOp, OpAsmPrinter *printer) {
...
+// spv.FuncionCall
+//===----------------------------------------------------------------------===//
+
+static ParseResult parseFunctionCallOp(OpAsmParser *parser,
+                                       OperationState *state) {
+  SymbolRefAttr calleeAttr;
+  FunctionType type;
+  SmallVector<OpAsmParser::OperandType, 4> operands;
+  auto loc = parser->getNameLoc();
+  if (parser->parseAttribute(calleeAttr, "callee", state->attributes) ||
```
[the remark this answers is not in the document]
denis0x0D: Oh, I see, thanks for mention this!

lib/Dialect/SPIRV/SPIRVOps.cpp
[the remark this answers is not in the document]
denis0x0D: @MaheshRavishankar thanks for mention it, I was actually learning from standard ops and llvm ops.

test/Dialect/SPIRV/Serialization/function_call.mlir
denis0x0D: @antiagainst thanks, got it!

test/Dialect/SPIRV/Serialization/function_call.mlir
denis0x0D: @antiagainst your are right! Thanks!

denis0x0D:
@antiagainst @MaheshRavishankar thanks a lot for review! I've updated the patch, can you please take a look?
Thanks.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -912,6 +913,108 @@ static void print(spirv::ExecutionModeOp execModeOp, OpAsmPrinter *printer) {
...
+           << funcType.getNumResults();
+  }
+
+  return failure(parser->addTypesToList(funcType.getResults(), state->types) ||
+                 parser->resolveOperands(operands, funcType.getInputs(), loc,
+                                         state->operands));
+}
+
+static void print(spirv::FunctionCallOp functionCallOp, OpAsmPrinter *printer) {
+  SmallVector<Type, 0> argTypes(functionCallOp.getOperandTypes());
```
antiagainst: 4 instead of 0?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -912,6 +913,108 @@ static void print(spirv::ExecutionModeOp execModeOp, OpAsmPrinter *printer) {
...
+  SmallVector<Type, 0> resultTypes(functionCallOp.getResultTypes());
```
antiagainst: 1 instead of 0?

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -1779,6 +1779,53 @@ Deserializer::processOp<spirv::ExecutionModeOp>(ArrayRef<uint32_t> words) {
...
+  if (functionName.empty()) {
+    functionName = "spirv_fn_" + std::to_string(functionID);
+  }
+
+  llvm::SmallVector<Value *, 4> arguments;
+  for (auto operand : llvm::drop_begin(operands, 3)) {
+    auto *value = getValue(operand);
+    if (!value) {
+      return emitError(unknownLoc, "unknown <id> ")
+             << operand << " used by spv.FunctionCall";
```
antiagainst: OpFunctionCall instead of spv.FunctionCall. Here we are deserializing SPIR-V blob so the error message should be talking about SPIR-V instructions.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -1779,6 +1779,53 @@ Deserializer::processOp<spirv::ExecutionModeOp>(ArrayRef<uint32_t> words) {
...
+  Type resultType = getType(operands[0]);
+  if (!resultType) {
+    return emitError(unknownLoc, "undefined result type from <id> ")
+           << operands[0];
+  }
+
+  auto resultID = operands[1];
+  auto functionID = operands[2];
+  
+  auto functionName = nameMap.lookup(functionID).str();
```
antiagainst: Can we have a `getFunctionSymbol` similar to `getSpecConstantSymbol` so the same logic can be reused in `processFunction`?

lib/Dialect/SPIRV/Serialization/Serializer.cpp
```diff
@@ -131,6 +131,15 @@ class Serializer {
     return funcIDMap.lookup(fnName);
   }
 
+  uint32_t getOrCreateFunctionID(StringRef fnName) {
```
antiagainst: We only use inline implementation when it's just a trivial oneliner. Can you move the implementation here out side of the decl?

antiagainst: Can you put a doc for this function? "// Gets the <id> for the function with the given name. Assigns the next available <id> if the function haven't been deserialized."

antiagainst (changes requested):
Nice, thanks a lot! Just a few more nits. :)

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: Oh, sorry, i've missed it again.

lib/Dialect/SPIRV/Serialization/Deserializer.cpp
denis0x0D: Thanks, fixed!

lib/Dialect/SPIRV/Serialization/Serializer.cpp
denis0x0D: Fixed, thanks!

denis0x0D:
@antiagainst thanks a lot for review, I've updated the patch, I also want to note, that I've updated the InFunctionScope trait, from

`CPred<"llvm::isa_and_nonnull<FuncOp>($_op.getParentOp())">>;
`

to 

`CPred<"($_op.getParentOfType<FuncOp>())">>;`

because previous one does not work if op is inside control flow op, I hope it's right.
Thanks.

antiagainst:
@denis0x0D: yes, that makes sense. All those ops marked as `InFunctionScope` should be allowed in loop regions too. Thanks for fixing this! :)

antiagainst:
Another recommendation regarding PR workflow: it would be nice if you can address comments by adding additional commits to the feature branch instead of force pushing. It would make code review easier so we can land PRs faster. :) We can do force pushing at the end after approval. Thanks! :)

denis0x0D:
> Another recommendation regarding PR workflow: it would be nice if you can address comments by adding additional commits to the feature branch instead of force pushing. It would make code review easier so we can land PRs faster. :) We can do force pushing at the end after approval. Thanks! :)

thanks a lot for mention this!

antiagainst (approved):

walli55 (commented):
Alomaisi

PR #139, opened 2019-09-17
Support symbolic operands for memref replacement; fix memrefNormalize

bondhugula:
- allow symbols in index remapping provided for memref replacement
- fix memref normalize

Reported-by: Alex Zinenko

### DIFF
--- a/include/mlir/Dialect/StandardOps/Ops.td
+++ b/include/mlir/Dialect/StandardOps/Ops.td
@@ -162,6 +162,12 @@ def AllocOp : Std_Op<"alloc"> {
     unsigned getNumSymbolicOperands() {
       return getNumOperands() - getType().getNumDynamicDims();
     }
+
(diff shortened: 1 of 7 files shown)


include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -162,6 +162,13 @@ def AllocOp : Std_Op<"alloc"> {
     unsigned getNumSymbolicOperands() {
       return getNumOperands() - getType().getNumDynamicDims();
     }
+
+    /// Returns the symbolic operands (the ones in square brackets), which bind
+    /// to the symbols of the memref's layout map.
+    operand_range getSymbolicOperands() {
+      return {getOperation()->operand_begin() + getType().getNumDynamicDims(),
```
River707: I don't think getOperation()-> is necessary here.

bondhugula: Fixed, thanks.

PR #141, opened 2019-09-21
Introduce splat op + provide its LLVM lowering

bondhugula:
- introduce splat op in standard dialect (currently for int/float input
  type, output type can be vector or statically shaped tensor)
- implement LLVM lowering (when result type is 1-d vector)
- add constant folding hook for it
- while on Ops.cpp, fix some stale names

### DIFF
--- a/include/mlir/Dialect/StandardOps/Ops.td
+++ b/include/mlir/Dialect/StandardOps/Ops.td
@@ -871,6 +871,33 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
   let summary = "signed integer shift left";
 }
 
+def SplatOp : Std_Op<"splat", [NoSideEffect]> {
+  let summary = "splat or broadcast operation";
+  let description = [{
+    The "splat" op reads a value of integer or float type and broadcasts it into
+    a vector or a tensor. The output of splat is thus a new value of either
+    vector or tensor type with elemental type being its operand's type.
+    When the result is a tensor, it has to be statically shaped.
+
+      %1 = splat %0 : vector<8xi32>
+      %2 = splat %0 : tensor<4x8xi32>
+
+    // TODO: handle broadcast to dynamically shaped tensors.
+  }];
+
+  let arguments = (ins AnyTypeOf<[AnyInteger, AnyFloat],
+                                 "integer or float type">:$input);
+
+  let results = (outs AnyTypeOf<[AnyVector, AnyStaticShapeTensor]>:$aggregate);
+
+  let builders =
+      [OpBuilder<"Builder *builder, OperationState &result, Value *element, "
+                  "Type aggregateType",
+                  [{ build(builder, result, aggregateType, element); }]>];
+
+  let hasFolder = 1;
+}
+
 def SubFOp : FloatArithmeticOp<"subf"> {
   let summary = "floating point subtraction operation";
   let hasFolder = 1;
--- a/lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
+++ b/lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
@@ -237,17 +237,6 @@ public:
     return builder.create<LLVM::ConstantOp>(loc, getIndexType(), attr);
   }
 
-  // Get the array attribute named "position" containing the given list of
-  // integers as integer attribute elements.
-  static ArrayAttr getIntegerArrayAttr(ConversionPatternRewriter &builder,
-                                       ArrayRef<int64_t> values) {
-    SmallVector<Attribute, 4> attrs;
-    attrs.reserve(values.size());
-    for (int64_t pos : values)
-      attrs.push_back(builder.getIntegerAttr(builder.getIndexType(), pos));
-    return builder.getArrayAttr(attrs);
-  }
-
   // Extract raw data pointer value from a value representing a memref.
   static Value *extractMemRefElementPtr(ConversionPatternRewriter &builder,
                                         Location loc,
@@ -260,7 +249,7 @@ public:
     else
       return builder.create<LLVM::ExtractValueOp>(
           loc, elementTypePtr, convertedMemRefValue,
-          getIntegerArrayAttr(builder, 0));
+          builder.getIndexArrayAttr(0));
     return buffer;
   }
 
@@ -1097,6 +1086,39 @@ struct CondBranchOpLowering
   using Super::Super;
 };
 
+// The Splat operation is lowered to an insertelement + a shufflevector
+// operation. Splat to only 1-d vector result types are lowered.
+struct SplatOpLowering : public LLVMLegalizationPattern<SplatOp> {
+  using LLVMLegalizationPattern<SplatOp>::LLVMLegalizationPattern;
+
+  PatternMatchResult
+  matchAndRewrite(Operation *op, ArrayRef<Value *> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto splatOp = cast<SplatOp>(op);
+    VectorType resultType = splatOp.getType().dyn_cast<VectorType>();
+    if (!resultType || resultType.getRank() != 1)
+      return matchFailure();
+
+    // First insert it into an undef vector so we can shuffle it.
+    auto vectorType = lowering.convertType(splatOp.getType());
+    Value *undef = rewriter.create<LLVM::UndefOp>(op->getLoc(), vectorType);
+    auto zero = rewriter.create<LLVM::ConstantOp>(
+        op->getLoc(), lowering.convertType(rewriter.getIntegerType(32)),
+        rewriter.getZeroAttr(rewriter.getIntegerType(32)));
+
+    auto v = rewriter.create<LLVM::InsertElementOp>(
+        op->getLoc(), vectorType, undef, splatOp.getOperand(), zero);
+
+    int64_t width = splatOp.getType().cast<VectorType>().getDimSize(0);
+    SmallVector<int32_t, 4> zeroValues(width, 0);
+
+    // Shuffle the value across the desired number of elements.
+    ArrayAttr zeroAttrs = rewriter.getI32ArrayAttr(zeroValues);
+    rewriter.replaceOpWithNewOp<LLVM::ShuffleVectorOp>(op, v, undef, zeroAttrs);
+    return matchSuccess();
+  }
+};
+
 } // namespace
 
 static void ensureDistinctSuccessors(Block &bb) {
@@ -1158,9 +1180,9 @@ void mlir::populateStdToLLVMConversionPatterns(
       DivFOpLowering, FuncOpConversion, IndexCastOpLowering, LoadOpLowering,
       MemRefCastOpLowering, MulFOpLowering, MulIOpLowering, OrOpLowering,
       RemISOpLowering, RemIUOpLowering, RemFOpLowering, ReturnOpLowering,
-      SelectOpLowering, SignExtendIOpLowering, SIToFPLowering, StoreOpLowering,
-      SubFOpLowering, SubIOpLowering, TruncateIOpLowering, XOrOpLowering,
-      ZeroExtendIOpLowering>(*converter.getDialect(), converter);
+      SelectOpLowering, SIToFPLowering, SignExtendIOpLowering, SplatOpLowering,
+      StoreOpLowering, SubFOpLowering, SubIOpLowering, TruncateIOpLowering,
+      XOrOpLowering, ZeroExtendIOpLowering>(*converter.getDialect(), converter);
 }
 
 // Convert types using the stored LLVM IR module.
--- a/lib/Dialect/StandardOps/Ops.cpp
+++ b/lib/Dialect/StandardOps/Ops.cpp
@@ -202,10 +202,10 @@ ParseResult mlir::parseDimAndSymbolList(OpAsmParser &parser,
   numDims = opInfos.size();
 
   // Parse the optional symbol operands.
-  auto affineIntTy = parser.getBuilder().getIndexType();
+  auto indexTy = parser.getBuilder().getIndexType();
   if (parser.parseOperandList(opInfos,
                               OpAsmParser::Delimiter::OptionalSquare) ||
-      parser.resolveOperands(opInfos, affineIntTy, operands))
+      parser.resolveOperands(opInfos, indexTy, operands))
     return failure();
   return success();
 }
@@ -1658,14 +1658,14 @@ static ParseResult parseExtractElementOp(OpAsmParser &parser,
   SmallVector<OpAsmParser::OperandType, 4> indexInfo;
   ShapedType type;
 
-  auto affineIntTy = parser.getBuilder().getIndexType();
+  auto indexTy = parser.getBuilder().getIndexType();
   return failure(
       parser.parseOperand(aggregateInfo) ||
       parser.parseOperandList(indexInfo, OpAsmParser::Delimiter::Square) ||
       parser.parseOptionalAttributeDict(result.attributes) ||
       parser.parseColonType(type) ||
       parser.resolveOperand(aggregateInfo, type, result.operands) ||
-      parser.resolveOperands(indexInfo, affineIntTy, result.operands) ||
+      parser.resolveOperands(indexInfo, indexTy, result.operands) ||
       parser.addTypeToList(type.getElementType(), result.types));
 }
 
@@ -1739,14 +1739,14 @@ static ParseResult parseLoadOp(OpAsmParser &parser, OperationState &result) {
   SmallVector<OpAsmParser::OperandType, 4> indexInfo;
   MemRefType type;
 
-  auto affineIntTy = parser.getBuilder().getIndexType();
+  auto indexTy = parser.getBuilder().getIndexType();
   return failure(
       parser.parseOperand(memrefInfo) ||
       parser.parseOperandList(indexInfo, OpAsmParser::Delimiter::Square) ||
       parser.parseOptionalAttributeDict(result.attributes) ||
       parser.parseColonType(type) ||
       parser.resolveOperand(memrefInfo, type, result.operands) ||
-      parser.resolveOperands(indexInfo, affineIntTy, result.operands) ||
+      parser.resolveOperands(indexInfo, indexTy, result.operands) ||
       parser.addTypeToList(type.getElementType(), result.types));
 }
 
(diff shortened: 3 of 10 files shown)


include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -836,6 +836,35 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
...
+    When the result is a tensor, it has to be statically shaped.
+
+      %1 = splat %0 : vector<8xi32>
+      %2 = splat %0 : tensor<4x8xi32>
+
+    // TODO: handle broadcast to dynamically shaped tensors.
+  }];
+
+  let arguments = (ins AnyTypeOf<[AnyInteger, Index, AnyFloat],
+                                 "integer, index, or float type">:$intOrFloat);
```
River707: intOrFloat -> input/operand

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -836,6 +836,35 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
...
+
+  let results = (outs AnyTypeOf<[ AnyVector, AnyTensor ]> : $aggregate);
```
River707: There are a few stray spaces here. [AnyVector, AnyTensor]>:$aggregate

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -836,6 +836,35 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
...
+
+  let builders =
+      [ OpBuilder<"Builder *builder, OperationState &result, Value *element, "
```
River707: Same here, there are stray spaces.

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -2018,6 +2018,61 @@ OpFoldResult SelectOp::fold(ArrayRef<Attribute> operands) {
...
+                 parser.parseOptionalAttributeDict(result.attributes) ||
+                 parser.parseColonType(shapedType) ||
+                 parser.resolveOperand(splatValueInfo,
+                                        shapedType.getElementType(),
+                                        result.operands) ||
+                 parser.addTypeToList(shapedType, result.types));
+}
+
+static LogicalResult verify(SplatOp op) {
+  if (!op.getOperand()->getType().isIntOrIndexOrFloat())
```
River707: This shouldn't be necessary.

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -2018,6 +2018,61 @@ OpFoldResult SelectOp::fold(ArrayRef<Attribute> operands) {
...
+    return op.emitError("operand should be of integer, index, or float type");
+
+  auto type = op.getType();
+
+  if (!type.isa<VectorType>() && !type.isa<TensorType>())
+    return op.emitError("result type should either be a vector or tensor type");
+
+  if (op.getOperand()->getType() != type.cast<ShapedType>().getElementType())
```
River707: Everything in this verify should be covered automatically.

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -2018,6 +2018,61 @@ OpFoldResult SelectOp::fold(ArrayRef<Attribute> operands) {
...
+
+  return success();
+}
+
+// Constant folding hook for SplatOp.
+OpFoldResult SplatOp::fold(ArrayRef<Attribute> operands) {
+  assert(operands.size() == 1 && "splat takes one operand");
+
+  auto constOperand = operands.front();
+  if (!constOperand.dyn_cast_or_null<IntegerAttr>() &&
```
River707: Change these to isa:

if (!constOperand || !constOperand.isa<> || ...

test/IR/core-ops.mlir
```diff
@@ -431,6 +431,17 @@ func @test_dimop(%arg0: tensor<4x4x?xf32>) {
   return
 }
 
+// CHECK-LABEL: func @test_splat_op
+// CHECK-SAME: [[S:%arg[0-9]+]]: f32
+func @test_splat_op(%s : f32) {
+  %v = splat %s : vector<8xf32>
+// CHECK: %{{.*}} = splat [[S]] : vector<8xf32>
```
River707: Drop all of the {{.*}} =

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -2018,6 +2018,61 @@ OpFoldResult SelectOp::fold(ArrayRef<Attribute> operands) {
...
+      !constOperand.dyn_cast_or_null<FloatAttr>())
+    return {};
+
+  auto shapedType = getType().cast<ShapedType>();
+  assert(shapedType.getElementType() == constOperand.getType() &&
+         "incorrect input attribute type for folding");
+
+  // SplatElementsAttr::get treats single value for second arg as being a splat.
+  return SplatElementsAttr::get(shapedType, {constOperand});
```
River707: You can't have a splat of index type.

include/mlir/Dialect/StandardOps/Ops.td
River707: This should be AnyStaticShapeTensor

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -836,6 +836,35 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
   let summary = "signed integer shift left";
 }
 
+def SplatOp : Std_Op<"splat", [NoSideEffect]> {
```
River707: This op should use the TCresVTEtIsSameAsOp<0, 0> trait

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -1082,6 +1095,42 @@ struct CondBranchOpLowering
...
+// The Splat operation is lowered to an insertelement + a shufflevector
+// operation. Splat to only 1-d vector result types are lowered.
+struct SplatOpLowering : public LLVMLegalizationPattern<SplatOp> {
+  using LLVMLegalizationPattern<SplatOp>::LLVMLegalizationPattern;
+
+  PatternMatchResult match(Operation *op) const override {
+    VectorType vectorType = cast<SplatOp>(op).getType().dyn_cast<VectorType>();
+    if (!vectorType || vectorType.getRank() != 1)
+      return matchFailure();
+    return matchSuccess();
```
River707: Just use matchAndRewrite instead

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: The result and the operand doesn't have the same element type here. The operand type here is the element type of the result.

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: Done.

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: Removed.

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: Removed.

lib/Dialect/StandardOps/Ops.cpp
bondhugula: Removed.

lib/Dialect/StandardOps/Ops.cpp
bondhugula: Except for checking for the operand type being the same as the result's element type?

test/IR/core-ops.mlir
bondhugula: Done.

lib/Dialect/StandardOps/Ops.cpp
bondhugula: You can actually have a splat of index type. Addresses can be stored in a vector and lowered.

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: Done.

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
bondhugula: Done.

lib/Dialect/StandardOps/Ops.cpp
bondhugula: Done.

include/mlir/Dialect/StandardOps/Ops.td
River707: TCresVTEtIsSameAsOp supports this exact use case: https://github.com/tensorflow/mlir/blob/c49b963db45a8425371083dc3ddef860bd28ad5c/include/mlir/IR/OpBase.td#L1517

lib/Dialect/StandardOps/Ops.cpp
River707: What I'm saying is, you can't create a SplatElementsAttr(technically a DenseElementsAttr) with an attribute of IndexType. It will assert.

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: But the doc comment over there says it checks if the operand and result have the same element type, which isn't the case here. Did you mean:
 TCOpIsBroadcastableToRes ?

River707: Ah yes, I see I was thinking of just using TCresVTEtIsSameAsOpBase. 
https://github.com/tensorflow/mlir/blob/be6746fc3b9e6bd0527cc961256e362f44130cd4/include/mlir/Dialect/VectorOps/VectorOps.td#L53

TCOpIsBroadcastableToRes won't work because both need to be shaped.

lib/Dialect/StandardOps/Ops.cpp
bondhugula: Thanks. We use 64-bit APInt to store attributes for the index type:
https://github.com/tensorflow/mlir/blob/c49b963db45a8425371083dc3ddef860bd28ad5c/lib/IR/Attributes.cpp#L268
It should be okay to add the missing case in DenseElementsAttr::get?

River707: We could, but if we go down this route we should really clarify how we store constants of IndexType internally. It implies that an 'index' has a maximum representation of i64, and must be treated as i64 when manipulating via constants. If someone writes code with their target in mind(i32 for example), index type won't work as they expect. DenseElementsAttr suffers more from this because we highly discourage operating on Attributes, because they are expensive to construct, and instead operate directly on APInt/uint32_t/uint64_t/etc.

River707: This is part of the reason why you are *not* allowed to use index in a vector or tensor. As such, index type should be removed from this op.
https://github.com/tensorflow/mlir/blob/master/g3doc/LangRef.md#index-type

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: Sorry, I still don't understand your comment. Those traits appear to be checking whether either their self types or elemental types match, which isn't what we need for splat. As an aside, the doc comments for both (TCresVTEtI..., TCOpIsBroad...) of these are inconsistent with their behavior and their names.

lib/Dialect/StandardOps/Ops.cpp
bondhugula: > We could, but if we go down this route we should really clarify how we store constants of IndexType internally. It implies that an 'index' has a maximum representation of i64, and must 

We already chose to store index type constants using 64-bit APInt internally. This was discussed at some point, and I think the thought process was that if a target didn't support 64-bit addresses, we would just truncate during target codegen. There were two other alternative design points that were discussed but this was the one that we went with at that time. The discussion on the choices and trade-offs was from @lattner

bondhugula: > This is part of the reason why you are _not_ allowed to use index in a vector or tensor. As such, index type should be removed from this op.
> https://github.com/tensorflow/mlir/blob/master/g3doc/LangRef.md#index-type

Thanks - I had forgotten about this. Then, we don't need to support index type here at all.

bondhugula: Done.

include/mlir/Dialect/StandardOps/Ops.td
River707: I'm trying to see if we can compose the traits together, i.e. selfOrElementType is too broad for what we really want. It allows shaped/non-shaped types on both sides, but that should be fine given that we constrain the operand and result types as well. That being said, we should likely add a trait that does exactly what we want given that we have the same verification for ExtractElementOp(Doesn't have to be this commit, but an issue/TODO would be good).

include/mlir/Dialect/StandardOps/Ops.td
River707: Seems like there is also an extra space in '[ OpBuilder'

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -238,13 +238,26 @@ class LLVMLegalizationPattern : public LLVMOpLowering {
   }
 
   // Get the array attribute named "position" containing the given list of
-  // integers as integer attribute elements.
+  // integers as integer attribute elements of index type.
+  static ArrayAttr getIndexArrayAttr(ConversionPatternRewriter &builder,
+                                     ArrayRef<int64_t> values) {
+    SmallVector<Attribute, 4> attrs;
+    attrs.reserve(values.size());
```
River707: This should already be covered by builder.getIndexArrayAttr(values)

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: Yes, it would be good to have a trait. But note that ExtractElementOp has the reverse requirement - the result there is an element of the operand, and the latter is the shaped type.

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -238,13 +238,26 @@ class LLVMLegalizationPattern : public LLVMOpLowering {
...
+    for (int64_t pos : values)
+      attrs.push_back(builder.getIntegerAttr(builder.getIndexType(), pos));
+    return builder.getArrayAttr(attrs);
+  }
+
+  // Get the array attribute named "position" containing the given list of
+  // integers as integer attribute elements of width "bitwidth".
   static ArrayAttr getIntegerArrayAttr(ConversionPatternRewriter &builder,
+                                       unsigned bitWidth,
```
River707: Is this function necessary anymore? It should be covered by builder.get* methods for most cases.

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -1082,6 +1095,41 @@ struct CondBranchOpLowering
...
+    if (!resultType || resultType.getRank() != 1)
+      return matchFailure();
+
+    auto splatOp = cast<SplatOp>(op);
+    // First insert it into an undef vector so we can shuffle it.
+    auto vectorType = lowering.convertType(splatOp.getType());
+    Value *undef = rewriter.create<LLVM::UndefOp>(op->getLoc(), vectorType);
+    auto zero = rewriter.create<LLVM::ConstantOp>(
+        op->getLoc(), lowering.convertType(rewriter.getIntegerType(32)),
+        rewriter.getIntegerAttr(rewriter.getIntegerType(32), 0));
```
River707: rewriter.getZeroAttr(i32Type)?

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -1082,6 +1095,41 @@ struct CondBranchOpLowering
...
+
+    auto v = rewriter.create<LLVM::InsertElementOp>(
+        op->getLoc(), vectorType, undef, splatOp.getOperand(), zero);
+
+    int64_t width = splatOp.getType().cast<VectorType>().getDimSize(0);
+    SmallVector<int64_t, 4> zeroValues(width, 0);
+
+    // Shuffle the value across the desired number of elements.
+    mlir::ArrayAttr zeroAttrs = getIntegerArrayAttr(rewriter, 32, zeroValues);
```
River707: rewriter.getI32ArrayAttr(zeroValues)

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -1082,6 +1095,41 @@ struct CondBranchOpLowering
...
+    Value *splatVector = rewriter.create<LLVM::ShuffleVectorOp>(
+        op->getLoc(), v, undef, zeroAttrs);
```
River707: rewriter.replaceOpWithNewOp

include/mlir/Dialect/StandardOps/Ops.td
River707: Yes, thanks for the correction. I was thinking InsertElement in my head.

bondhugula:
Thanks very much for the review! I've addressed all the remaining comments.

include/mlir/Dialect/StandardOps/Ops.td
bondhugula: Added a TODO at the verify method.

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
bondhugula: Not needed anymore.

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -1082,6 +1071,39 @@ struct CondBranchOpLowering
...
+  using LLVMLegalizationPattern<SplatOp>::LLVMLegalizationPattern;
+
+  PatternMatchResult
+  matchAndRewrite(Operation *op, ArrayRef<Value *> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    VectorType resultType = cast<SplatOp>(op).getType().dyn_cast<VectorType>();
+    if (!resultType || resultType.getRank() != 1)
+      return matchFailure();
+
+    auto splatOp = cast<SplatOp>(op);
```
River707: Can you hoist this up to avoid casting twice?

lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -1082,6 +1071,39 @@ struct CondBranchOpLowering
...
+        rewriter.getZeroAttr(rewriter.getIntegerType(32)));
+
+    auto v = rewriter.create<LLVM::InsertElementOp>(
+        op->getLoc(), vectorType, undef, splatOp.getOperand(), zero);
+
+    int64_t width = splatOp.getType().cast<VectorType>().getDimSize(0);
+    SmallVector<int32_t, 4> zeroValues(width, 0);
+
+    // Shuffle the value across the desired number of elements.
+    mlir::ArrayAttr zeroAttrs = rewriter.getI32ArrayAttr(zeroValues);
```
River707: Drop the mlir::.

test/Transforms/constant-fold.mlir
```diff
@@ -540,3 +540,15 @@ func @custom_insertion_position() {
...
+
+// CHECK-LABEL: func @splat_fold
+func @splat_fold() -> (vector<4xf32>, tensor<4xf32>) {
+  %c = constant 1.0 : f32
+  %v = splat %c : vector<4xf32>
+  %t = splat %c : tensor<4xf32>
+  return %v, %t : vector<4xf32>, tensor<4xf32>
+
+  // CHECK-NEXT: %cst = constant dense<1.000000e+00> : vector<4xf32>
+  // CHECK-NEXT: %cst{{.*}} = constant dense<1.000000e+00> : tensor<4xf32>
```
River707: Use FileCheck variables for the names here.

bondhugula:
Squashed and force pushed.

bondhugula:
> Squashed and force pushed.

Resolved conflicts with the recent sign/zero extension commits.

bondhugula:
An test case unrelated to this PR is failing in the internal build.

PR #151, opened 2019-09-26
Add a pair of hooks to DominanceInfo for MemToReg like pass

schweitzpgi:
These two hooks are to allow our implementation of the MemToReg like pass to transform the FIR dialect.

### DIFF
--- a/include/mlir/Analysis/Dominance.h
+++ b/include/mlir/Analysis/Dominance.h
@@ -91,7 +91,10 @@ public:
     return super::properlyDominates(a, b);
   }
 
+  /// Return the dominanceInfos from the Region containing block A.
   llvm::DomTreeNodeBase<Block> *getNode(Block *a);
+
+  /// Update the DFS numbers on all the dominanceInfos.
   void updateDFSNumbers();
 };
 
--- a/include/mlir/Analysis/Dominance.h
+++ b/include/mlir/Analysis/Dominance.h
@@ -91,10 +91,10 @@ public:
     return super::properlyDominates(a, b);
   }
 
-  /// Return the dominanceInfos from the Region containing block A.
-  llvm::DomTreeNodeBase<Block> *getNode(Block *a);
+  /// Return the dominance node from the Region containing block A.
+  DominanceInfoNode *getNode(Block *a);
 
-  /// Update the DFS numbers on all the dominanceInfos.
+  /// Update the internal DFS numbers for the dominance nodes.
   void updateDFSNumbers();
 };
 
--- a/include/mlir/Analysis/Dominance.h
+++ b/include/mlir/Analysis/Dominance.h
@@ -90,6 +90,9 @@ public:
   bool properlyDominates(Block *a, Block *b) {
     return super::properlyDominates(a, b);
   }
+
+  llvm::DomTreeNodeBase<Block> *getNode(Block *a);
+  void updateDFSNumbers();
 };
 
 /// A class for computing basic postdominance information.
--- a/lib/Analysis/Dominance.cpp
+++ b/lib/Analysis/Dominance.cpp
@@ -142,6 +142,18 @@ bool DominanceInfo::properlyDominates(Value *a, Operation *b) {
   return dominates(cast<BlockArgument>(a)->getOwner(), b->getBlock());
 }
 
+llvm::DomTreeNodeBase<Block> *DominanceInfo::getNode(Block *a) {
+  auto region = a->getParent();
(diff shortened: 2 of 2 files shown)


include/mlir/Analysis/Dominance.h
```diff
@@ -90,6 +90,9 @@ class DominanceInfo : public detail::DominanceInfoBase</*IsPostDom=*/false> {
   bool properlyDominates(Block *a, Block *b) {
     return super::properlyDominates(a, b);
   }
+
+  llvm::DomTreeNodeBase<Block> *getNode(Block *a);
+  void updateDFSNumbers();
```
joker-eph: These are public APIs, can you provide documentation for them?

schweitzpgi:
> Is this possible to get testing coverage for this in-tree?

Can you suggest the right way to test this interface? No test was provided here (although we exercise this by running mem2reg from the Fortran bits) because it wasn't totally clear what the best thing to do in the MLIR test environment.

include/mlir/Analysis/Dominance.h
schweitzpgi: Will do.

include/mlir/Analysis/Dominance.h
```diff
@@ -90,6 +90,12 @@ class DominanceInfo : public detail::DominanceInfoBase</*IsPostDom=*/false> {
   bool properlyDominates(Block *a, Block *b) {
     return super::properlyDominates(a, b);
   }
+
+  /// Return the dominanceInfos from the Region containing block A.
```
River707: dominanceInfos doesn't make sense here, can this change to 'dominance node'. This should also change the llvm::DomTreeNodeBase<Block> -> DominanceInfoNode

include/mlir/Analysis/Dominance.h
```diff
@@ -90,6 +90,12 @@ class DominanceInfo : public detail::DominanceInfoBase</*IsPostDom=*/false> {
...
+  llvm::DomTreeNodeBase<Block> *getNode(Block *a);
+
+  /// Update the DFS numbers on all the dominanceInfos.
```
River707: dominanceInfos doesn't make sense here either, can we just say 'Updated the internal DFS numbers for the dominance nodes'

lib/Analysis/Dominance.cpp
```diff
@@ -142,6 +142,18 @@ bool DominanceInfo::properlyDominates(Value *a, Operation *b) {
...
 }
 
+llvm::DomTreeNodeBase<Block> *DominanceInfo::getNode(Block *a) {
+  auto region = a->getParent();
+  assert(dominanceInfos.count(region) != 0);
+  return dominanceInfos[region]->getNode(a);
+}
+
+void DominanceInfo::updateDFSNumbers() {
+  for (auto &iter : dominanceInfos) {
```
River707: Please drop the trivial braces.

include/mlir/Analysis/Dominance.h
schweitzpgi: change pushed

include/mlir/Analysis/Dominance.h
schweitzpgi: change pushed

lib/Analysis/Dominance.cpp
schweitzpgi: done

lib/Analysis/Dominance.cpp
```diff
@@ -142,6 +142,17 @@ bool DominanceInfo::properlyDominates(Value *a, Operation *b) {
   return dominates(cast<BlockArgument>(a)->getOwner(), b->getBlock());
 }
 
+DominanceInfoNode *DominanceInfo::getNode(Block *a) {
+  auto region = a->getParent();
```
River707: nit: Region *region or auto *region

River707:
Orthogonally from this patch, how generic is the Mem2Reg pass for FIR? Would it be something that we could possibly contribute to upstream MLIR in the future(likely with some tweaks/generalizations)?

schweitzpgi:
> Orthogonally from this patch, how generic is the Mem2Reg pass for FIR? Would it be something that we could possibly contribute to upstream MLIR in the future(likely with some tweaks/generalizations)?

The code was taken from LLVM and suitably modified to work with FIR Operations, etc.  I'm confident it can be made generic with some templates and work with any dialect. (Perhaps with the appropriate Interface; i.e. a Memory form that uses local temporaries.)

PR #153, opened 2019-09-26
Integer right shift

ombre5733:
MLIR has already a std.shlis op but no lowering to LLVM.
This commit adds the lowering and complements std.shlis by std.shris and std.shriu. The latter ones are lowered to llvm.ashr and llvm.lshr.

### DIFF
--- a/include/mlir/Dialect/StandardOps/Ops.td
+++ b/include/mlir/Dialect/StandardOps/Ops.td
@@ -881,6 +881,14 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
   let summary = "signed integer shift left";
 }
 
+def ShrISOp : IntArithmeticOp<"shris"> {
+  let summary = "signed integer shift right";
+}
+
+def ShrIUOp : IntArithmeticOp<"shriu"> {
+  let summary = "unsigned integer shift right";
+}
+
 def SplatOp : Std_Op<"splat", [NoSideEffect]> {
   let summary = "splat or broadcast operation";
   let description = [{
--- a/include/mlir/Dialect/LLVMIR/LLVMOps.td
+++ b/include/mlir/Dialect/LLVMIR/LLVMOps.td
@@ -145,6 +145,9 @@ def LLVM_SRemOp : LLVM_ArithmeticOp<"srem", "CreateSRem">;
 def LLVM_AndOp : LLVM_ArithmeticOp<"and", "CreateAnd">;
 def LLVM_OrOp : LLVM_ArithmeticOp<"or", "CreateOr">;
 def LLVM_XOrOp : LLVM_ArithmeticOp<"xor", "CreateXor">;
+def LLVM_ShlOp : LLVM_ArithmeticOp<"shl", "CreateShl">;
+def LLVM_LShrOp : LLVM_ArithmeticOp<"lshr", "CreateLShr">;
+def LLVM_AShrOp : LLVM_ArithmeticOp<"ashr", "CreateAShr">;
 
 // Predicate for integer comparisons.
 def ICmpPredicateEQ  : I64EnumAttrCase<"eq", 0>;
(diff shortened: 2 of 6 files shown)


include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -881,6 +881,14 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
...
 }
 
+def ShrISOp : IntArithmeticOp<"shris"> {
+  let summary = "signed integer shift right";
+}
+
+def ShrIUOp : IntArithmeticOp<"shriu"> {
+  let summary = "unsigned integer shift right";
+}
+
```
joker-eph: I'm wondering where is this naming coming from? Is there some kind of standard?
I find these fairly hard to read if you're not using them daily (the LLVM ones as well by the way, but LLVM is more contained).
I wonder if we could read a custom printer/parser like:

```
%2 = bitshift(%0, %1)  {right, signed}  : i16
```

or more inventively:

```
%2 = signed<<(%0, %1) : i16
```

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -881,6 +881,14 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
   let summary = "signed integer shift left";
 }
 
+def ShrISOp : IntArithmeticOp<"shris"> {
+  let summary = "signed integer shift right";
+}
```
joker-eph: We should do a better job at providing extensive documentation here, as we intend to generate the documentation from this. See the doc for `sexti` above as an example

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -881,6 +881,14 @@ def ShlISOp : IntArithmeticOp<"shlis"> {
...
+
+def ShrIUOp : IntArithmeticOp<"shriu"> {
+  let summary = "unsigned integer shift right";
+}
```
joker-eph: We need to specify better the behavior of these operation for edge cases (wrapping, etc.). In this dialect we don't have a notion of poison just like LLVM, and this dialect is intended to be re-used in multiple context, which makes it harder to narrow down the semantics too much (so I am not sure what is the best thing to do here)

include/mlir/Dialect/StandardOps/Ops.td
ombre5733: There was already a `shlis` op in MLIR (it was added back when `addi` was introduced) so I just adopted that naming for the right shifts. Also this name seems to be consistent with `remis` and `diviu`.

That being said, I also like the `bitshift` syntax. We should just make sure that encoding the `signed`/`unsigned` behavior in an attribute is the right thing to do (because all other standard ops encode this in their name).

include/mlir/Dialect/StandardOps/Ops.td
ombre5733: I've written the doc for `sexti`. Glad you like it! :-) I'll write a similar paragraph also for the shifts.

include/mlir/Dialect/StandardOps/Ops.td
ombre5733: One more question: Should we have a signed left shift at all? Wouldn't a generic left shift be sufficient (also there is only one integer add op for both signed and unsigned)?

include/mlir/Dialect/StandardOps/Ops.td
ombre5733: Hm, good question. Should we constrain the shift amount to be less than the bit-width for example? Unfortunately we cannot verify this so we would need to enforce that via additional instructions.
Should we discuss this on the mailing list?

joker-eph: This is always welcome to be discussed on the mailing-list of course!

ombre5733:
Closing this in favor of #226.

PR #146, opened 2019-09-26
[ROCm] Adding ROCDL Dialect.

deven-amd:
This PR introduces the ROCDL Dialect (i.e. the ROCDL ops + the code to lower those ROCDL ops to LLVM intrinsics/functions). Think of ROCDL Dialect as analogous to the NVVM Dialect, but for AMD GPUs. This PR contains just the essentials needed to get a simple example up and running. We expect to make further additions to the ROCDL Dialect.

This is the first of 3 PRs we will be filing. Once this PR is merged it will be followed by
 * a PR to add a pass that lowers GPU Dialect to ROCDL Dialect
 * a PR to add a "mlir-rocm-runner" utility

Note:  This PR is a follow-up on PR #63. It builds on the work done by @whchung in that PR, and addresses most of the review comments in that PR.

---------------------------------

@whchung @joker-eph

PR #154, opened 2019-09-27
[ROCm] Adding pass to lower GPU Dialect to ROCDL Dialect.

deven-amd:
This is a follow-up to the PR #146  which introduced the ROCDL Dialect. This PR introduces a pass to lower GPU Dialect to the ROCDL Dialect. As with the previous PR, this one builds on the work done by @whchung, and addresses most of the review comments in the original PR #63 .

---------------------------------

@joker-eph @whchung


include/mlir/Conversion/GPUToROCDL/GPUToROCDLPass.h
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+// =============================================================================
+#ifndef MLIR_CONVERSION_GPUTOROCDL_GPUTOROCDLPASS_H_
+#define MLIR_CONVERSION_GPUTOROCDL_GPUTOROCDLPASS_H_
+
+#include <memory>
+
+namespace mlir {
+class LLVMTypeConverter;
+class OwningRewritePatternList;
```
River707: I don't think most of these forward declares are necessary.

lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include "mlir/Conversion/StandardToLLVM/ConvertStandardToLLVM.h"
+#include "mlir/Conversion/StandardToLLVM/ConvertStandardToLLVMPass.h"
+#include "mlir/Dialect/GPU/GPUDialect.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Dialect/LLVMIR/ROCDLDialect.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/StandardTypes.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassRegistry.h"
```
River707: PassRegistry isn't necessary.

lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                    ROCDL::BlockDimYOp, ROCDL::BlockDimZOp>,
+        GPUIndexIntrinsicOpLowering<gpu::BlockId, ROCDL::BlockIdXOp,
+                                    ROCDL::BlockIdYOp, ROCDL::BlockIdZOp>,
+        GPUIndexIntrinsicOpLowering<gpu::GridDim, ROCDL::GridDimXOp,
+                                    ROCDL::GridDimYOp, ROCDL::GridDimZOp>>(
+        converter);
+
+    ConversionTarget target(getContext());
+    target.addLegalDialect<LLVM::LLVMDialect>();
+    target.addLegalDialect<ROCDL::ROCDLDialect>();
```
River707: These methods chain: addLegalDialect<..., ...>();

test/Conversion/GPUToROCDL/gpu-to-rocdl.mlir
```diff
@@ -0,0 +1,37 @@
+// RUN: mlir-opt %s -lower-gpu-ops-to-rocdl-ops | FileCheck %s
+
+module attributes {gpu.kernel_module} {
+  // CHECK-LABEL: func @gpu_index_ops()
+  func @gpu_index_ops()
+      attributes { gpu.kernel } {
+    // CHECK: = rocdl.workitem.id.x : !llvm.i32
```
River707: Let's drop the '= ' on each of these.

deven-amd:
@River707 

Thank you for your feedback. I have pushed out a new commit which addresses the points you raised. Please re-review.

thanks

deven

lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+namespace {
+
+// Rewriting that replaces Op with XOp, YOp, or ZOp depending on the dimension
+// that Op operates on.  Op is assumed to return an `std.index` value and
+// XOp, YOp and ZOp are assumed to return an `llvm.i32` value.  Depending on
+// `indexBitwidth`, sign-extend or truncate the resulting value to match the
+// bitwidth expected by the consumers of the value.
+template <typename Op, typename XOp, typename YOp, typename ZOp>
+struct GPUIndexIntrinsicOpLowering : public LLVMOpLowering {
```
sherhut: This should be shared with the NVVM lowering by moving it to a shared implementation header file. The two targets are very similar, so I would hope we can share even more code over time.

deven-amd: @sherhut where should this shared implementation header file reside?

`include/mlir/Conversion/<?>/LowerGpuOpsCommon.h`

Once you decide on a name, I can submit a followup PR to create that dir/file and move the implementation of `struct GPUIndexIntrinsicOpLowering` to that file.  In the meantime, please take this PR (as it is gating the next PR we need to file for the `ROCM` backend for AMD GPUs)

thanks

deven

joker-eph: Thanks Deven, I'm fine if you do it in a follow-up CL. Thanks!

sherhut: As this is an implementation detail, I would place it in 

`lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h`

for now. We can always move it around if we decide there are too many GPU* directories. I just want to avoid that the code gets duplicated and then starts to diverge.

Issue #128, opened 2019-09-11
[spirv] Add a pass to decorate types with layout info


antiagainst:
If a composite type is used for global variables in `StorageBuffer`/`Uniform`/`PushConstant` storage class, then it must be explicitly [laid out](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_shadervalidation_a_validation_rules_for_shader_a_href_capability_capabilities_a). Right now lowering passes need to make sure the generated SPIR-V global variables follow the above rule by themselves. This is duplicating the work. We should consider to have a single pass to go through all global variables in a `spv.module` and change their types to have layout decorations. 

The layout rules can potentially be a CLI option to the pass. But for Vulkan ML, we only need [Vulkan Standard Buffer Layout](https://renderdoc.org/vkspec_chunked/chap14.html#interfaces-resources-layout).

antiagainst:
@denis0x0D, are you interested in this? :)

denis0x0D:
@antiagainst yes! Can I take it?
Thanks.

antiagainst:
Sure! Thanks for your help!

denis0x0D:
@antiagainst can you please comment on the following, do I get you right?
For each global variable in spv.module, if variable is not a composite type(spv.struct, spv.array, vector), we have to change its type to spv.struct. Something like this:
`spv.globalVariable @var1 bind(0, 0) : !spv.ptr<f32, StroageBuffer>` 
to 
`spv.globalVariable @var1 bind(0, 0) : !spv.ptr<!spv.struct<f32 [0]>, StorageBuffer>`

In this case global variable will be laid out be serializer, decorated with Offset.

antiagainst:
I think that's a separate issue. For this issue, we should just handle existing `!spv.struct` in StorageBuffer/Uniform/PushConstant storage class but without layout decorations. An example is turning

```
spv.globalVariable @var1 bind(0, 0) : !spv.ptr<!spv.struct<f32>, StorageBuffer>
```
into
```
spv.globalVariable @var1 bind(0, 0) : !spv.ptr<!spv.struct<f32 [0]>, StorageBuffer>
```

denis0x0D:
@antiagainst got it, thanks!

MaheshRavishankar:
[...]

PR #156, opened 2019-09-27
[spirv] Add pass to decorate types with layout info.

denis0x0D:
This PR adds a pass to decorate SPIR-V types with layout info https://github.com/tensorflow/mlir/issues/128
Current implementaton decorates type of spirv::GlobalVariableOp if type is !spv.ptr<!spv.struct<...>> in strorage class StorageBuffer/Uniform/PushConstants and does not have layout info. Also it changes the type for the direct user of spirv::GlobalVariableOp such as spv.AddressOfOp. 
It does not change the type for the indirect users such as Load, Store, FunctionCall and so on.

Thanks.

### DIFF
--- a/lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
+++ b/lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
@@ -1,0 +1,321 @@
+//===- DecorateSPIRVCompositeTypeLayoutPass.cpp - Decorate composite type -===//
+//
+// Copyright 2019 The MLIR Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+//
+// This file implements a pass to decorate the composite types used by
+// composite objects in the StorageBuffer, PhysicalStorageBuffer, Uniform, and
+// PushConstant storage classes with layout information. See SPIR-V spec
+// "2.16.2. Validation Rules for Shader Capabilities" for more details.
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Dialect/SPIRV/Passes.h"
+#include "mlir/Dialect/SPIRV/SPIRVDialect.h"
+#include "mlir/Dialect/SPIRV/SPIRVOps.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+using namespace mlir;
+
+/// According to the Vulkan spec "14.5.4. Offset and Stride Assignment":
+/// "There are different alignment requirements depending on the specific
+/// resources and on the features enabled on the device."
+///
+/// There are 3 types of alignment: scalar, base, extended.
+/// See the spec for details.
+///
+/// Note: Even if scalar alignment is supported, it is generally more
+/// performant to use the base alignment. So here the calculation is based on
+/// base alignment.
+///
+/// The memory layout must obey the following rules:
+/// 1. The Offset decoration of any member must be a multiple of its alignment.
+/// 2. Any ArrayStride or MatrixStride decoration must be a multiple of the
+/// alignment of the array or matrix as defined above.
+///
+/// According to the SPIR-V spec:
+/// "The ArrayStride, MatrixStride, and Offset decorations must be large
+/// enough to hold the size of the objects they affect (that is, specifying
+/// overlap is invalid)."
+namespace {
+class VulkanLayoutUtils {
+public:
+  using Alignment = uint64_t;
+
+  /// Returns a new type with layout info. Assigns the type size in bytes to the
+  /// `size`. Assigns the type alignment in bytes to the `alignment`.
+  static Type decorateType(spirv::StructType structType,
+                           spirv::StructType::LayoutInfo &size,
+                           Alignment &alignment);
+  /// Checks whether a type is legal in terms of Vulkan layout info
+  /// decoration. A type is dynamically illegal if it's a composite type in the
+  /// StorageBuffer, PhysicalStorageBuffer, Uniform, and PushConstant Storage
+  /// Classes without layout informtation.
+  static bool isLegalType(Type type);
+
+private:
+  static Type decorateType(Type type, spirv::StructType::LayoutInfo &size,
+                           Alignment &alignment);
+  static Type decorateType(VectorType compositeType,
+                           spirv::StructType::LayoutInfo &size,
+                           Alignment &aligment);
+  static Type decorateType(spirv::ArrayType compositeType,
+                           spirv::StructType::LayoutInfo &size,
+                           Alignment &aligment);
+  /// Calculates the alignment for the given scalar type.
+  static Alignment getScalarTypeAlignment(Type scalarType);
+};
+
+Type VulkanLayoutUtils::decorateType(spirv::StructType structType,
+                                     spirv::StructType::LayoutInfo &size,
+                                     VulkanLayoutUtils::Alignment &alignment) {
+  if (structType.getNumElements() == 0) {
+    return structType;
+  }
+
+  llvm::SmallVector<Type, 4> memberTypes;
+  llvm::SmallVector<spirv::StructType::LayoutInfo, 4> layoutInfo;
+  llvm::SmallVector<spirv::StructType::MemberDecorationInfo, 4>
+      memberDecorations;
+
+  spirv::StructType::LayoutInfo structMemberOffset = 0;
+  VulkanLayoutUtils::Alignment maxMemberAlignment = 1;
+
+  for (uint32_t i = 0, e = structType.getNumElements(); i < e; ++i) {
+    spirv::StructType::LayoutInfo memberSize = 0;
+    VulkanLayoutUtils::Alignment memberAlignment = 1;
+
+    auto memberType = VulkanLayoutUtils::decorateType(
+        structType.getElementType(i), memberSize, memberAlignment);
+    structMemberOffset = llvm::alignTo(structMemberOffset, memberAlignment);
+    memberTypes.push_back(memberType);
+    layoutInfo.push_back(structMemberOffset);
+    // According to the Vulkan spec:
+    // "A structure has a base alignment equal to the largest base alignment of
+    // any of its members."
+    structMemberOffset += memberSize;
+    maxMemberAlignment = std::max(maxMemberAlignment, memberAlignment);
+  }
+
+  // According to the Vulkan spec:
+  // "The Offset decoration of a member must not place it between the end of a
+  // structure or an array and the next multiple of the alignment of that
+  // structure or array."
+  size = llvm::alignTo(structMemberOffset, maxMemberAlignment);
+  alignment = maxMemberAlignment;
+  structType.getMemberDecorations(memberDecorations);
+  return spirv::StructType::get(memberTypes, layoutInfo, memberDecorations);
+}
+
+Type VulkanLayoutUtils::decorateType(Type type,
+                                     spirv::StructType::LayoutInfo &size,
+                                     VulkanLayoutUtils::Alignment &alignment) {
+  if (spirv::SPIRVDialect::isValidScalarType(type)) {
+    alignment = VulkanLayoutUtils::getScalarTypeAlignment(type);
+    // Vulkan spec does not specify any padding for a scalar type.
+    size = alignment;
+    return type;
+  }
+
+  switch (type.getKind()) {
+  case spirv::TypeKind::Struct:
+    return VulkanLayoutUtils::decorateType(type.cast<spirv::StructType>(), size,
+                                           alignment);
+  case spirv::TypeKind::Array:
+    return VulkanLayoutUtils::decorateType(type.cast<spirv::ArrayType>(), size,
+                                           alignment);
+  case StandardTypes::Vector:
+    return VulkanLayoutUtils::decorateType(type.cast<VectorType>(), size,
+                                           alignment);
+  default:
+    llvm_unreachable("unhandled SPIR-V type");
+  }
+}
+
+Type VulkanLayoutUtils::decorateType(VectorType vectorType,
+                                     spirv::StructType::LayoutInfo &size,
+                                     VulkanLayoutUtils::Alignment &alignment) {
+  const auto numElements = vectorType.getNumElements();
+  auto elementType = vectorType.getElementType();
+  spirv::StructType::LayoutInfo elementSize = 0;
+  VulkanLayoutUtils::Alignment elementAlignment = 1;
+
+  auto memberType = VulkanLayoutUtils::decorateType(elementType, elementSize,
+                                                    elementAlignment);
+  // According to the Vulkan spec:
+  // 1. "A two-component vector has a base alignment equal to twice its scalar
+  // alignment."
+  // 2. "A three- or four-component vector has a base alignment equal to four
+  // times its scalar alignment."
+  size = elementSize * numElements;
+  alignment = numElements == 2 ? elementAlignment * 2 : elementAlignment * 4;
+  return VectorType::get(numElements, memberType);
+}
+
+Type VulkanLayoutUtils::decorateType(spirv::ArrayType arrayType,
+                                     spirv::StructType::LayoutInfo &size,
+                                     VulkanLayoutUtils::Alignment &alignment) {
+  const auto numElements = arrayType.getNumElements();
+  auto elementType = arrayType.getElementType();
+  spirv::ArrayType::LayoutInfo elementSize = 0;
+  VulkanLayoutUtils::Alignment elementAlignment = 1;
+
+  auto memberType = VulkanLayoutUtils::decorateType(elementType, elementSize,
+                                                    elementAlignment);
+  // According to the Vulkan spec:
+  // "An array has a base alignment equal to the base alignment of its element
+  // type."
+  size = elementSize * numElements;
+  alignment = elementAlignment;
+  return spirv::ArrayType::get(memberType, numElements, elementSize);
+}
+
+VulkanLayoutUtils::Alignment
+VulkanLayoutUtils::getScalarTypeAlignment(Type scalarType) {
+  // According to the Vulkan spec:
+  // 1. "A scalar of size N has a scalar alignment of N."
+  // 2. "A scalar has a base alignment equal to its scalar alignment."
+  // 3. "A scalar, vector or matrix type has an extended alignment equal to its
+  // base alignment."
+  auto bitWidth = scalarType.getIntOrFloatBitWidth();
+  if (bitWidth == 1)
+    return 1;
+  return bitWidth / 8;
+}
+
+bool VulkanLayoutUtils::isLegalType(Type type) {
+  auto ptrType = type.dyn_cast<spirv::PointerType>();
+  if (!ptrType) {
+    return true;
+  }
+
+  auto storageClass = ptrType.getStorageClass();
+  auto structType = ptrType.getPointeeType().dyn_cast<spirv::StructType>();
+  if (!structType) {
+    return true;
+  }
+
+  switch (storageClass) {
+  case spirv::StorageClass::Uniform:
+  case spirv::StorageClass::StorageBuffer:
+  case spirv::StorageClass::PushConstant:
+  case spirv::StorageClass::PhysicalStorageBuffer:
+    return structType.hasLayout() || !structType.getNumElements();
+  default:
+    return true;
+  }
+}
+} // namespace
+
+namespace {
+class SPIRVGlobalVariableOpLayoutInfoDecoration
+    : public OpRewritePattern<spirv::GlobalVariableOp> {
+public:
+  using OpRewritePattern<spirv::GlobalVariableOp>::OpRewritePattern;
+
+  PatternMatchResult matchAndRewrite(spirv::GlobalVariableOp op,
+                                     PatternRewriter &rewriter) const override {
+    spirv::StructType::LayoutInfo structSize = 0;
+    VulkanLayoutUtils::Alignment structAlignment = 1;
+    SmallVector<NamedAttribute, 4> globalVarAttrs;
+
+    auto ptrType = op.type().cast<spirv::PointerType>();
+    auto structType = VulkanLayoutUtils::decorateType(
+        ptrType.getPointeeType().cast<spirv::StructType>(), structSize,
+        structAlignment);
+    auto decoratedType =
+        spirv::PointerType::get(structType, ptrType.getStorageClass());
+
+    // Save all named attributes except "type" attribute.
+    for (const auto &attr : op.getAttrs()) {
+      if (attr.first == "type") {
+        continue;
+      }
+      globalVarAttrs.push_back(attr);
+    }
+
+    rewriter.replaceOpWithNewOp<spirv::GlobalVariableOp>(
+        op, rewriter.getTypeAttr(decoratedType), globalVarAttrs);
+    return matchSuccess();
+  }
+};
+
+class SPIRVAddressOfOpLayoutInfoDecoration
+    : public OpRewritePattern<spirv::AddressOfOp> {
+public:
+  using OpRewritePattern<spirv::AddressOfOp>::OpRewritePattern;
+
+  PatternMatchResult matchAndRewrite(spirv::AddressOfOp op,
+                                     PatternRewriter &rewriter) const override {
+    auto spirvModule = op.getParentOfType<spirv::ModuleOp>();
+    auto varName = op.variable();
+    auto varOp = spirvModule.lookupSymbol<spirv::GlobalVariableOp>(varName);
+
+    rewriter.replaceOpWithNewOp<spirv::AddressOfOp>(
+        op, varOp.type(), rewriter.getSymbolRefAttr(varName));
+    return matchSuccess();
+  }
+};
+} // namespace
+
+static void populateSPIRVLayoutInfoPatterns(OwningRewritePatternList &patterns,
+                                            MLIRContext *ctx) {
+  patterns.insert<SPIRVGlobalVariableOpLayoutInfoDecoration,
+                  SPIRVAddressOfOpLayoutInfoDecoration>(ctx);
+}
+
+namespace {
+class DecorateSPIRVCompositeTypeLayoutPass
+    : public ModulePass<DecorateSPIRVCompositeTypeLayoutPass> {
+private:
+  void runOnModule() override;
+};
+
+void DecorateSPIRVCompositeTypeLayoutPass::runOnModule() {
+  auto module = getModule();
+  OwningRewritePatternList patterns;
+  populateSPIRVLayoutInfoPatterns(patterns, module.getContext());
+  ConversionTarget target(*(module.getContext()));
+  target.addLegalDialect<spirv::SPIRVDialect>();
+  target.addLegalOp<FuncOp>();
+  target.addDynamicallyLegalOp<spirv::GlobalVariableOp>(
+      [](spirv::GlobalVariableOp op) {
+        return VulkanLayoutUtils::isLegalType(op.type());
+      });
+
+  // Change the type for the direct users.
+  target.addDynamicallyLegalOp<spirv::AddressOfOp>([](spirv::AddressOfOp op) {
+    return VulkanLayoutUtils::isLegalType(op.pointer()->getType());
+  });
+
+  // TODO: Change the type for the indirect users such as spv.Load, spv.Store,
+  // spv.FunctionCall and so on.
+
+  for (auto spirvModule : module.getOps<spirv::ModuleOp>()) {
+    if (failed(applyFullConversion(spirvModule, target, patterns))) {
+      signalPassFailure();
+    }
+  }
+}
+} // namespace
+
+std::unique_ptr<OpPassBase<ModuleOp>>
+mlir::spirv::createDecorateSPIRVCompositeTypeLayoutPass() {
+  return std::make_unique<DecorateSPIRVCompositeTypeLayoutPass>();
+}
+
+static PassRegistration<DecorateSPIRVCompositeTypeLayoutPass>
+    pass("decorate-spirv-composite-type-layout",
+         "Decorate SPIR-V composite type with layout info");
--- a/include/mlir/Dialect/SPIRV/Passes.h
+++ b/include/mlir/Dialect/SPIRV/Passes.h
@@ -29,6 +29,9 @@ namespace spirv {
 
 std::unique_ptr<OpPassBase<mlir::ModuleOp>> createConvertStandardToSPIRVPass();
 
+std::unique_ptr<OpPassBase<mlir::ModuleOp>>
+createDecorateSPIRVCompositeTypeLayoutPass();
+
 } // namespace spirv
 } // namespace mlir
 
--- a/include/mlir/Dialect/SPIRV/SPIRVDialect.h
+++ b/include/mlir/Dialect/SPIRV/SPIRVDialect.h
@@ -38,6 +38,9 @@ public:
   /// Checks if the given `type` is valid in SPIR-V dialect.
   static bool isValidType(Type type);
 
+  /// Checks if the given `scalar type` is valid in SPIR-V dialect.
+  static bool isValidScalarType(Type type);
+
   /// Returns the attribute name to use when specifying decorations on results
   /// of operations.
   static std::string getAttributeName(Decoration decoration);
--- a/include/mlir/Dialect/SPIRV/SPIRVStructureOps.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVStructureOps.td
@@ -242,6 +242,14 @@ def SPV_GlobalVariableOp : SPV_Op<"globalVariable", [InModuleScope]> {
     OptionalAttr<SymbolRefAttr>:$initializer
   );
 
+  let builders = [
+    OpBuilder<"Builder *builder, OperationState &state, "
+      "TypeAttr type, ArrayRef<NamedAttribute> namedAttrs", [{
+      state.addAttribute("type", type);
+      state.addAttributes(namedAttrs);
+    }]>
+  ];
+
   let results = (outs);
 
   let hasOpcode = 0;
--- a/lib/Dialect/SPIRV/CMakeLists.txt
+++ b/lib/Dialect/SPIRV/CMakeLists.txt
@@ -20,3 +20,4 @@ target_link_libraries(MLIRSPIRV
   MLIRSupport)
 
 add_subdirectory(Serialization)
+add_subdirectory(Transforms)
--- a/lib/Dialect/SPIRV/SPIRVDialect.cpp
+++ b/lib/Dialect/SPIRV/SPIRVDialect.cpp
@@ -96,7 +96,7 @@ static bool isValidSPIRVIntType(IntegerType type) {
                             type.getWidth());
 }
 
-static bool isValidSPIRVScalarType(Type type) {
+bool SPIRVDialect::isValidScalarType(Type type) {
   if (type.isa<FloatType>()) {
     return !type.isBF16();
   }
@@ -107,7 +107,8 @@ static bool isValidSPIRVScalarType(Type type) {
 }
 
 static bool isValidSPIRVVectorType(VectorType type) {
-  return type.getRank() == 1 && isValidSPIRVScalarType(type.getElementType()) &&
+  return type.getRank() == 1 &&
+         SPIRVDialect::isValidScalarType(type.getElementType()) &&
          type.getNumElements() >= 2 && type.getNumElements() <= 4;
 }
 
@@ -117,7 +118,7 @@ bool SPIRVDialect::isValidType(Type type) {
       type.getKind() <= TypeKind::LAST_SPIRV_TYPE) {
     return true;
   }
-  if (isValidSPIRVScalarType(type)) {
+  if (SPIRVDialect::isValidScalarType(type)) {
     return true;
   }
   if (auto vectorType = type.dyn_cast<VectorType>()) {
--- a/lib/Dialect/SPIRV/Transforms/CMakeLists.txt
+++ b/lib/Dialect/SPIRV/Transforms/CMakeLists.txt
@@ -1,0 +1,8 @@
+add_llvm_library(MLIRSPIRVTransforms
+  DecorateSPIRVCompositeTypeLayoutPass.cpp
+  )
+
+target_link_libraries(MLIRSPIRVTransforms
+  MLIRPass
+  MLIRSPIRV
+  )
--- a/test/Dialect/SPIRV/Transforms/layout-decoration.mlir
+++ b/test/Dialect/SPIRV/Transforms/layout-decoration.mlir
@@ -1,0 +1,99 @@
+// RUN: mlir-opt -decorate-spirv-composite-type-layout -split-input-file -verify-diagnostics %s -o - | FileCheck %s
+
+spv.module "Logical" "GLSL450" {
+  // CHECK: spv.globalVariable @var0 bind(0, 1) : !spv.ptr<!spv.struct<i32 [0], !spv.struct<f32 [0], i32 [4]> [4], f32 [12]>, Uniform>
+  spv.globalVariable @var0 bind(0,1) : !spv.ptr<!spv.struct<i32, !spv.struct<f32, i32>, f32>, Uniform>
+
+  // CHECK: spv.globalVariable @var1 bind(0, 2) : !spv.ptr<!spv.struct<!spv.array<64 x i32 [4]> [0], f32 [256]>, StorageBuffer>
+  spv.globalVariable @var1 bind(0,2) : !spv.ptr<!spv.struct<!spv.array<64xi32>, f32>, StorageBuffer>
+
+  // CHECK: spv.globalVariable @var2 bind(1, 0) : !spv.ptr<!spv.struct<!spv.struct<!spv.array<64 x i32 [4]> [0], f32 [256]> [0], i32 [260]>, StorageBuffer>
+  spv.globalVariable @var2 bind(1,0) : !spv.ptr<!spv.struct<!spv.struct<!spv.array<64xi32>, f32>, i32>, StorageBuffer>
+
+  // CHECK: spv.globalVariable @var3 : !spv.ptr<!spv.struct<!spv.array<16 x !spv.struct<f32 [0], f32 [4], !spv.array<16 x f32 [4]> [8]> [72]> [0], f32 [1152]>, StorageBuffer>
+  spv.globalVariable @var3 : !spv.ptr<!spv.struct<!spv.array<16x!spv.struct<f32, f32, !spv.array<16xf32>>>, f32>, StorageBuffer>
+
+  // CHECK: spv.globalVariable @var4 bind(1, 2) : !spv.ptr<!spv.struct<!spv.struct<!spv.struct<i1 [0], i8 [1], i16 [2], i32 [4], i64 [8]> [0], f32 [16], i1 [20]> [0], i1 [24]>, StorageBuffer>
+  spv.globalVariable @var4 bind(1,2) : !spv.ptr<!spv.struct<!spv.struct<!spv.struct<i1, i8, i16, i32, i64>, f32, i1>, i1>, StorageBuffer>
+
+  // CHECK: spv.globalVariable @var5 bind(1, 3) : !spv.ptr<!spv.struct<!spv.array<256 x f32 [4]> [0]>, StorageBuffer>
+  spv.globalVariable @var5 bind(1,3) : !spv.ptr<!spv.struct<!spv.array<256xf32>>, StorageBuffer>
+
+  func @kernel() -> () {
+    %c0 = spv.constant 0 : i32
+    // CHECK: {{%.*}} = spv._address_of @var0 : !spv.ptr<!spv.struct<i32 [0], !spv.struct<f32 [0], i32 [4]> [4], f32 [12]>, Uniform>
+    %0 = spv._address_of @var0 : !spv.ptr<!spv.struct<i32, !spv.struct<f32, i32>, f32>, Uniform>
+    // CHECK:  {{%.*}} = spv.AccessChain {{%.*}}[{{%.*}}] : !spv.ptr<!spv.struct<i32 [0], !spv.struct<f32 [0], i32 [4]> [4], f32 [12]>, Uniform>
+    %1 = spv.AccessChain %0[%c0] : !spv.ptr<!spv.struct<i32, !spv.struct<f32, i32>, f32>, Uniform>
+    spv.Return
+  }
+}
+
+// -----
+
+spv.module "Logical" "GLSL450" {
+  // CHECK: spv.globalVariable @var0 : !spv.ptr<!spv.struct<!spv.struct<!spv.struct<!spv.struct<!spv.struct<i1 [0], i1 [1], f64 [8]> [0], i1 [16]> [0], i1 [24]> [0], i1 [32]> [0], i1 [40]>, Uniform>
+  spv.globalVariable @var0 : !spv.ptr<!spv.struct<!spv.struct<!spv.struct<!spv.struct<!spv.struct<i1, i1, f64>, i1>, i1>, i1>, i1>, Uniform>
+
+  // CHECK: spv.globalVariable @var1 : !spv.ptr<!spv.struct<!spv.struct<i16 [0], !spv.struct<i1 [0], f64 [8]> [8], f32 [24]> [0], f32 [32]>, Uniform>
+  spv.globalVariable @var1 : !spv.ptr<!spv.struct<!spv.struct<i16, !spv.struct<i1, f64>, f32>, f32>, Uniform>
+
+  // CHECK: spv.globalVariable @var2 : !spv.ptr<!spv.struct<!spv.struct<i16 [0], !spv.struct<i1 [0], !spv.array<16 x !spv.array<16 x i64 [8]> [128]> [8]> [8], f32 [2064]> [0], f32 [2072]>, Uniform>
+  spv.globalVariable @var2 : !spv.ptr<!spv.struct<!spv.struct<i16, !spv.struct<i1, !spv.array<16x!spv.array<16xi64>>>, f32>, f32>, Uniform>
+
+  // CHECK: spv.globalVariable @var3 : !spv.ptr<!spv.struct<!spv.struct<!spv.array<64 x i64 [8]> [0], i1 [512]> [0], i1 [520]>, Uniform>
+  spv.globalVariable @var3 : !spv.ptr<!spv.struct<!spv.struct<!spv.array<64xi64>, i1>, i1>, Uniform>
+
+  // CHECK: spv.globalVariable @var4 : !spv.ptr<!spv.struct<i1 [0], !spv.struct<i64 [0], i1 [8], i1 [9], i1 [10], i1 [11]> [8], i1 [24]>, Uniform>
+  spv.globalVariable @var4 : !spv.ptr<!spv.struct<i1, !spv.struct<i64, i1, i1, i1, i1>, i1>, Uniform>
+
+  // CHECK: spv.globalVariable @var5 : !spv.ptr<!spv.struct<i1 [0], !spv.struct<i1 [0], i1 [1], i1 [2], i1 [3], i64 [8]> [8], i1 [24]>, Uniform>
+  spv.globalVariable @var5 : !spv.ptr<!spv.struct<i1, !spv.struct<i1, i1, i1, i1, i64>, i1>, Uniform>
+
+  // CHECK: spv.globalVariable @var6 : !spv.ptr<!spv.struct<i1 [0], !spv.struct<i64 [0], i32 [8], i16 [12], i8 [14], i1 [15]> [8], i1 [24]>, Uniform>
+  spv.globalVariable @var6 : !spv.ptr<!spv.struct<i1, !spv.struct<i64, i32, i16, i8, i1>, i1>, Uniform>
+
+  // CHECK: spv.globalVariable @var7 : !spv.ptr<!spv.struct<i1 [0], !spv.struct<!spv.struct<i1 [0], i64 [8]> [0], i1 [16]> [8], i1 [32]>, Uniform>
+  spv.globalVariable @var7 : !spv.ptr<!spv.struct<i1, !spv.struct<!spv.struct<i1, i64>, i1>, i1>, Uniform>
+}
+
+// -----
+
+spv.module "Logical" "GLSL450" {
+  // CHECK: spv.globalVariable @var0 : !spv.ptr<!spv.struct<vector<2xi32> [0], f32 [8]>, StorageBuffer>
+  spv.globalVariable @var0 : !spv.ptr<!spv.struct<vector<2xi32>, f32>, StorageBuffer>
+
+  // CHECK: spv.globalVariable @var1 : !spv.ptr<!spv.struct<vector<3xi32> [0], f32 [12]>, StorageBuffer>
+  spv.globalVariable @var1 : !spv.ptr<!spv.struct<vector<3xi32>, f32>, StorageBuffer>
+
+  // CHECK: spv.globalVariable @var2 : !spv.ptr<!spv.struct<vector<4xi32> [0], f32 [16]>, StorageBuffer>
+  spv.globalVariable @var2 : !spv.ptr<!spv.struct<vector<4xi32>, f32>, StorageBuffer>
+}
+
+// -----
+
+spv.module "Logical" "GLSL450" {
+  // CHECK: spv.globalVariable @emptyStructAsMember : !spv.ptr<!spv.struct<!spv.struct<> [0]>, StorageBuffer>
+  spv.globalVariable @emptyStructAsMember : !spv.ptr<!spv.struct<!spv.struct<>>, StorageBuffer>
+
+  // CHECK: spv.globalVariable @arrayType : !spv.ptr<!spv.array<4 x !spv.array<4 x f32>>, StorageBuffer>
+  spv.globalVariable @arrayType : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, StorageBuffer>
+
+  // CHECK: spv.globalVariable @InputStorage : !spv.ptr<!spv.struct<!spv.array<256 x f32>>, Input>
+  spv.globalVariable @InputStorage : !spv.ptr<!spv.struct<!spv.array<256xf32>>, Input>
+
+  // CHECK: spv.globalVariable @customLayout : !spv.ptr<!spv.struct<f32 [256], i32 [512]>, Uniform>
+  spv.globalVariable @customLayout : !spv.ptr<!spv.struct<f32 [256], i32 [512]>, Uniform>
+
+  // CHECK:  spv.globalVariable @emptyStruct : !spv.ptr<!spv.struct<>, Uniform>
+  spv.globalVariable @emptyStruct : !spv.ptr<!spv.struct<>, Uniform>
+}
+
+// -----
+
+spv.module "Logical" "GLSL450" {
+  // CHECK: spv.globalVariable @var0 : !spv.ptr<!spv.struct<i32 [0]>, PushConstant>
+  spv.globalVariable @var0 : !spv.ptr<!spv.struct<i32>, PushConstant>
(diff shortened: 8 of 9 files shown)


lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+  /// type kind.
+  static Type getArrayOrVectorType(spirv::CompositeType compositeType,
+                                   Type elementType, uint64_t arrayStride = 0);
+  /// Calculate the size in bytes of the primitive type.
+  static unsigned getPrimitiveTypeSizeInBytes(Type primitiveType);
+  /// Check that type is primitive in terms of SPIR-V dialect.
+  static bool isPrimitiveType(Type type);
+};
+
+Type SPIRVLayoutInfoUtils::decorateType(spirv::StructType structType,
```
MaheshRavishankar: This method needs to account for alignment of the data types as well. For example if you have an spv.struct<f32, f64>, then you need to
1) Align the entire struct to the largest alignment of any of its members
2) Add padding so that individual members are aligned.

I have to look at the SPEC carefully, but definitely needs taking care of alignment here.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+  // "The ArrayStride, MatrixStride, and Offset decorations must be large
+  // enough to hold the size of the objects they affect (that is, specifying
+  // overlap is invalid)."
+  auto bitWidth = primitiveType.getIntOrFloatBitWidth();
+  if (bitWidth == 1 || bitWidth == 8 || bitWidth == 16)
+    return 4;
+  return bitWidth / 8;
+}
+
+bool SPIRVLayoutInfoUtils::isLegalType(Type type) {
```
MaheshRavishankar: Could you add some comments of what is legal and what is not. Even if just a copy of paste from relevant portion of the spec, thats fine.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+  }
+}
+
+unsigned SPIRVLayoutInfoUtils::getPrimitiveTypeSizeInBytes(Type primitiveType) {
+  // According to the SPIR-V spec:
+  // "The ArrayStride, MatrixStride, and Offset decorations must be large
+  // enough to hold the size of the objects they affect (that is, specifying
+  // overlap is invalid)."
+  auto bitWidth = primitiveType.getIntOrFloatBitWidth();
+  if (bitWidth == 1 || bitWidth == 8 || bitWidth == 16)
```
MaheshRavishankar: Why do this for bitwidth 8 and 16? Couldn't they use 1 byte and 2 bytes respectively. THe SPEC snippet mentioned here doesn't seem to show why this choice is made.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+namespace {
+class SPIRVGlobalVariableOpLayoutInfoDecoration
+    : public OpRewritePattern<spirv::GlobalVariableOp> {
+public:
+  using OpRewritePattern<spirv::GlobalVariableOp>::OpRewritePattern;
+
+  PatternMatchResult matchAndRewrite(spirv::GlobalVariableOp op,
+                                     PatternRewriter &rewriter) const override {
+    // Initial offset is zero.
+    spirv::StructType::LayoutInfo baseOffset = 0;
```
MaheshRavishankar: If the struct already has the layout information, nothing should be done (match should fail)

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+class SPIRVAddressOfOpLayoutInfoDecoration
+    : public OpRewritePattern<spirv::AddressOfOp> {
+public:
+  using OpRewritePattern<spirv::AddressOfOp>::OpRewritePattern;
+
+  PatternMatchResult matchAndRewrite(spirv::AddressOfOp op,
+                                     PatternRewriter &rewriter) const override {
+    auto spirvModule = op.getParentOfType<spirv::ModuleOp>();
+    if (!spirvModule) {
```
MaheshRavishankar: I dont think you need to check this (or this check needs to be part of the verifier of addressOfOp

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+      return matchFailure();
+    }
+    auto varName = op.variable();
+    auto varOp = spirvModule.lookupSymbol<spirv::GlobalVariableOp>(varName);
+    if (!varOp) {
```
MaheshRavishankar: This doesnt need to be checked either.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+        return SPIRVLayoutInfoUtils::isLegalType(op.type());
+      });
+  // Change the type for the direct users.
+  target.addDynamicallyLegalOp<spirv::AddressOfOp>([](spirv::AddressOfOp op) {
+    return SPIRVLayoutInfoUtils::isLegalType(op.pointer()->getType());
+  });
+  // TODO: Change the type for the indirect users such as spv.Load, spv.Store,
+  // spv.FunctionCall and so on.
+  for (auto spirvModule : module.getOps<spirv::ModuleOp>()) {
+    if (failed(applyPartialConversion(spirvModule, target, patterns))) {
```
MaheshRavishankar: applyPartialConversion is OK, but maybe better to do applyFullConversion. It is better to know that not all global variables conversion was possible.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
denis0x0D: @MaheshRavishankar oh, I've missed the alignment, thanks for mention this, you are right! I'll fix it.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
denis0x0D: @MaheshRavishankar the spec just requires to have Offset decorations large
 enough to hold the size of the objects they affect, I was missing the spec rule for alignment, initial idea was to align to 32-bit, but it's wrong.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
denis0x0D: You are right, but the check is under isLegalType, in terms of legalization I assume that type is illegal if it's a !spv.ptr<!spv.struct<...> in strorage class StorageBuffer/Uniform/PushConstant and does not have layout info, otherwise it could be illegal type with layout info, which is wrong. So, If type has layout info the pattern is not applicable here, I've provided a test with custom layout info. Do I miss something? Please fix me, if I'm wrong.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+    auto memberType = SPIRVLayoutInfoUtils::decorateType(
+        structType.getElementType(i), structOffset);
+
+    memberTypes.push_back(memberType);
+    layoutInfo.push_back(fieldOffset);
+    fieldOffset = structOffset;
+  }
+
+  offset += structOffset;
+  return spirv::StructType::get(memberTypes, layoutInfo);
```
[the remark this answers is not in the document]
denis0x0D: @MaheshRavishankar thanks for mention this!

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+  }
+  if (storageClass != spirv::StorageClass::Uniform &&
+           storageClass != spirv::StorageClass::StorageBuffer &&
+           storageClass != spirv::StorageClass::PushConstant) {
+    return true;
+  }
+  return structType.hasLayout();
+}
+
+bool SPIRVLayoutInfoUtils::isPrimitiveType(Type primitiveType) {
```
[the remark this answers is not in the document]
denis0x0D: @MaheshRavishankar thanks!

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,282 @@
...
+  if (!ptrType) {
+    return true;
+  }
+  auto storageClass = ptrType.getStorageClass();
+  auto structType = ptrType.getPointeeType().dyn_cast<spirv::StructType>();
+  if (!structType) {
+    return true;
+  }
+  if (storageClass != spirv::StorageClass::Uniform &&
+           storageClass != spirv::StorageClass::StorageBuffer &&
```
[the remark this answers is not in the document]
denis0x0D: @MaheshRavishankar thanks!

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
MaheshRavishankar: I am not sure when the `isLegal` function is invoked during the conversion. You might be right. But do add a test where the layout of the struct is different from what is assigned by default in this pass, and that shouldn't change. If that test passes, then I guess this is fine.

joker-eph:
It seemed that in the bug @antiagainst was in the opinion that layout should always be present and it is the responsibility of the lowering to make sure it happens? (and so there would be a verifier for these to be there)

MaheshRavishankar:
> It seemed that in the bug @antiagainst was in the opinion that layout should always be present and it is the responsibility of the lowering to make sure it happens? (and so there would be a verifier for these to be there)

Reading the bug, the statement is that instead of having the burden on every lowering, there could be a pass that could assign layout.
AFAICS, when the layout assignment happens is not directly related to verification. The verification can still check that the layout is specified. Lowering infrastructures just need to make sure this pass is run before the verification is run. IMO it is probably not worth having this in the verification itself, but rather as a check in the serialization. Though, having it in the verification is also OK.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
denis0x0D: isLegalType function is invoked:
```
  target.addDynamicallyLegalOp<spirv::GlobalVariableOp>(
      [](spirv::GlobalVariableOp op) {
        return SPIRVLayoutInfoUtils::isLegalType(op.type());
      });

```


My base assumptions for this pass are:
1. SPIR-V dialect is legal and therefore all SPIR-V ops are legal.
2. FuncOp is legal, because SPIR-V dialect reuses it from standard ops.
3. spirv:GlobalVariableOp is dynamically legal, it's illegal if:
a. Has special type. b. Has special strorage class c. Does not have layout info.
After pattern is applied for illegal op it becomes legal.

The struct with custom layout is legal, because it has layout.
Here is a test:  
```
// CHECK: spv.globalVariable @customLayout : !spv.ptr<!spv.struct<f32 [64], i32 [128]>, Uniform>
  spv.globalVariable @customLayout : !spv.ptr<!spv.struct<f32 [64], i32 [128]>, Uniform>


```
by the way, I've just started to learn MLIR Legalization machinery, and it could be another and better way to apply a convertion for those ops, please fix me in this case.

MaheshRavishankar: Ok, thanks for the clarification. Your understanding is how I understand as well. Since there is already a test, this seems OK to me. Please ignore my concern on this point!

joker-eph:
I'm am still confused: the comment in the bug I'm referring to is:

> During the lowering, we also need to generate properly laid out spv.struct types for global resources. This issue focuses on the later. (Although I think the layout calculation rule can be shared. :)

Is this what you're referring to? It seems to indicate that the layout will happen during the lowering, but this is different from having a separate pass. I understand this as the need to primitives/utility functions that can be involved during lowering to apply the layout rules.

We fundamentally can't have the lowering emitting invalid IR and a pass that operates on invalid IR.

MaheshRavishankar:
> I'm am still confused: the comment in the bug I'm referring to is:
> 
> > During the lowering, we also need to generate properly laid out spv.struct types for global resources. This issue focuses on the later. (Although I think the layout calculation rule can be shared. :)
> 
> Is this what you're referring to? It seems to indicate that the layout will happen during the lowering, but this is different from having a separate pass. I understand this as the need to primitives/utility functions that can be involved during lowering to apply the layout rules.
> 
> We fundamentally can't have the lowering emitting invalid IR and a pass that operates on invalid IR.

First of all, I dont think I would classify it as illegal IR if the struct type is not laid out. It has to be laid before serialization of the SPIR-V module, but I dont think we should treat the IR where struct types are not laid out after lowering (i.e. some dialect -> SPIR-V dialect) as illegal. It just has to happen before the serialization. So the pass in this CL could be run "just before" the serialization step.
If we do want to consider IR that doesnt have the struct laid out as illegal IR, the lowering (i.e. some dialect -> SPIR-V lowering) need not explicitly worry about the layout. Then before the lowering returns, it can invoke this pass to lay out the struct. So the lowering would create "legal" IR after return. The verifier can be enhanced to check that the structs are indeed laid out.

I am not sure I follow what the issue being raised here is. The pass is still useful to have. You can call it a "legalization pass" and the decision of when to invoke it is a separate concern

joker-eph:
> First of all, I dont think I would classify it as illegal IR if the struct type is not laid out

Well OK, but then that's a different story now. You originally wrote in the bug "I agree, it should be part of verifier too", and you wrote here "The verifier can be enhanced to check that the structs are indeed laid out", are you using "verifier" with a different meaning than the [MLIR verifier](https://github.com/tensorflow/mlir/blob/master/tools/mlir-opt/mlir-opt.cpp#L55)?

> I am not sure I follow what the issue being raised here is.

The issue is one of invariant and IR design and make sure things are properly considered documented (which they may be, but it isn't clear to me from reading the bug, and I may have missed the documentation in the repo?).
From what I read in the bug (and your explanation here: " It just has to happen before the serialization"), it is not valid in the SPIR-V spec to not have the layout.
If so, we may be saying that in MLIR it is legal to not have it, but that means that no transformation or analysis that operate on SPIR-V will be able to assume that a layout is present and everywhere where it matter you will have to check if it is present. This may or may not be intended, but it probably deserves to be considered and spelled out.
Deviation from the spec like this may not be intuitive for folks writing passes on SPIR-V dialect, so we should just be careful (in the doc and the API design) to make is hard to mis-use.

> If we do want to consider IR that doesnt have the struct laid out as illegal IR, the lowering [...] can invoke this pass to lay out the struct. So the lowering would create "legal" IR after return. The verifier can be enhanced to check that the structs are indeed laid out.

That may be an OK workflow, but this should not be exposed as a *pass*: it is never appropriate for a pass to take as input illegal IR by definition.

I need to dig a bit more into the layout question here, I only had a superficial look on this so far.

MaheshRavishankar:
@joker-eph : Fair point, I am not being very clear here. Let me restate

1) The SPIR-V dialect is meant to represent the SPIR-V spec, but its not a perfect mirror. There already many deviations to make it an amenable to optimization passes. You are right in saying we should have a clear spec of what is legal in SPIR-V dialect, and the verifier should check those.
2) At serialization time we have obviously make sure the serialized binary matches the SPIR-V spec.  The serialization should catch any deviations, but the SPIR-V dialect spec should be such that a module with valid SPIR-V spec should always be serializable.

Coming to the layout question here, one way to view this is that *all* spirv::StructTypes will get some default layout. I don't think we have explicitly stated this anywhere, but it seems like using a C-style layout would be the way forward, and thats what is being done in the pass. So by default lowering passes don't need to explicitly specify the layout if they want to use the default (but they are free to specify a custom layout for a struct). Before serialization though, we need to materialize the default layout. This pass is doing that. Requiring that all lowering always specify the layout will probably result in a lot of redundancy since most common case would be what is being done here.

Given the above, I was wrong in saying that we should check that the layout is specified in the verifier. If the approach above is OK, then we dont need to. The serializer can report an error if the layout isn't specified for a struct.

Disclaimer: This is my understanding. @antiagainst might have more insight.

joker-eph:
> but the SPIR-V dialect spec should be such that a module with valid SPIR-V spec should always be serializable.

My understanding was that there was a clear goal for the SPIR-V dialect to have a "straightforward" translation to/from SPIR-V binary format. 
While I agree with "deviations to make it an amenable to optimization passes" (I argued in this direction in the past), this is different from having a valid SPIR-V IR that can't be directly serialized.

antiagainst:
Sorry my previous reply on the bug wasn't clear enough. Let me expand it with more details. :)

A SPIR-V binary module contains not only the logic of the program (called shader/kernel in graphics/copute term), but also instructions for aiding the "interface" between the program and the application (or runtime in ML term). Examples for the former include `OpFunction`, `OpIAdd`, etc. Examples for the later include capabilities/extensions needed by this module (specified via `OpCapability`, `OpExtension`), layout information for resources (specified as `OpDecorate` on struct/array types), entry points (specified via `OpEntryPoint`, `OpExecutionMode`). Instructions for the later are at the very beginning of a binary module so that when a Vulkan driver trying to compile this module into its own ISA, it can reject early if the interface is incorrect instead of going through all the logic. Let's call the former as logic instructions and the later as interface instructions for the purpose of the discussion here. (There are also debug instructions not falling into either category, but let's ignore them for now.)

For logic instructions the verification is clear and unconditional. An `OpIAdd` can only be used adding two scalar or vector of integer values; anything other than that is wrong. But for interface instructions, it depends on the execution environment: Vulkan may require one way, and OpenCL may require another. Note that the requirement of "Composite objects in the StorageBuffer, PhysicalStorageBuffer, Uniform, and PushConstant Storage Classes must be explicitly laid out" is in SPIR-V spec "2.16.2. Validation Rules for Shader Capabilities". So it should only be enforced when a SPIR-V module requires `Shader` capability, which is the case for Vulkan. It's valid to have a module without those annotations if we don't know the execution environment or it's one that does not support `Shader` capability. Also, these interface instructions are typically "duplicated" information from the main logic (like required extension or capability, which we can deduce by going over all instructions in a module) or do not have a unique way (for example any `void(void)` function in the module can potentially be an entry point).

So for a pass, we should have the assumption that all ops corresponding to logic instructions are valid. But for information corresponding to interface instructions, I think most passes don't really care about them. (General passes like const folding, CSE, or SPIR-V specific ones like freezing spec constant won't behave differently w.r.t. weather we target Vulkan 1.0 or Vulkan 1.1 or OpenCL.)  To make a rough analogy with LLVM, the interface instructions are more like metadata from IR's perspective but are required for the interface with the application. I don't think we want to burden all of these passes to update the required capabilities or annotating layouts when generating new ops, like we have no guarantee regarding metadata in LLVM passes. It's better to add them with dedicated transformations, which will require external input regarding the application (e.g., what execution environment we are targeting and what layout rules we want to use). We don't have the support for different target execution environments right now; but we surely need and will add that for supporting like Vulkan 1.0 vs Vulkan 1.1.

joker-eph:
Thanks Lei for the great explanation!

> To make a rough analogy with LLVM, the interface instructions are more like metadata from IR's perspective but are required for the interface with the application. 

LLVM metadata can always be dropped, is this the case here?
LLVM has module flag which can be semantically meaningful and are likely closer to "interface" as here.

>I don't think we want to burden all of these passes to update the required capabilities or annotating layouts when generating new ops, like we have no guarantee regarding metadata in LLVM passes.

If you present the capabilities and/or the layout are purely redundant with the IR, I'd argue to strip them entirely on import and add them on export.
If you can't strip them, why? Are they meaningful then? If they are then the corollary seems like passes would need to maintain these informations no?

MaheshRavishankar:
To unblock this work, can we reach a consensus. If having a pass is problematic, then maybe convert all the layout assignment logic here into a utility function. For now we can invoke this in the serializer to make sure the generated binary is spec compliant?

MaheshRavishankar:
@denis0x0D : Looked through your changes. Seem OK to me. Will wait for response about whether having it as a pass or utility function from @antiagainst , @joker-eph to approve.

antiagainst:
> LLVM metadata can always be dropped, is this the case here?

It cannot. So I said this is a "rough analogy" to LLVM. :)

> LLVM has module flag which can be semantically meaningful and are likely closer to "interface" as here.

That is good to know! Thanks for pointing this out!

> If you present the capabilities and/or the layout are purely redundant with the IR, I'd argue to strip them entirely on import and add them on export.
> If you can't strip them, why? Are they meaningful then? If they are then the corollary seems like passes would need to maintain these informations no?

I think it depends on what we are talking about. 

The layout information is decorated onto the type (both in the spec and in the SPIR-V dialect). If we have two structs with identical members but different layout, they are two different types inside SPIR-V. If we strip the information during import, we are losing this difference. Besides, there can exist [multiple layout schemes](https://github.com/Microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#memory-layout-rules) and for different storage classes we can have different rules. If we strip them out during import and then re-export, we can generate modules that do not match with the original layout.

For layout I think it does not matter for passes since the layout (and other decorations on types) is baked into the type itself. So passes can operate on `spv.struct`s with or without layout info similarly. I think we just need a dedicated pass to turn "layoutless" types into ones with layout (it will need CLI options for which layout scheme to use from the developer) at the final stage when we know about the execution environment so that we generate valid modules according to that execution environment. 

For required extensions and capabilities, I was actually debating over it but settled down on the current approach: allow setting the required extensions/capabilities as attributes on `spv.module`. The reasons behind are:
* It models SPIR-V spec more directly given this is what the spec requires.
* It makes it possible for us to import a real-world SPIR-V blob and do verification on it and print nice error messages regarding capability/extension issues (which is quite common).
* It makes a runtime's life easier to investigate the required extensions/capabilities and decide whether the current hardware supports a specific `spv.module`. The runtime does not need to have knowledge over what capabilities/extensions are required for each opcode/operand or run a dedicated pass to generate that. (I was thinking that someday after we have a binary format for MLIR in general some runtime or game engine might ship a `spv.module` directly and use it as the single source of truth for graphical shaders to generate others like DXIL, GLSL, etc.)

Regarding the interaction with passes, I think most passes don't care either. We just need a dedicated pass at the final stage to go over all opcodes and operands to collect the needed extensions/capabilities and attach to the module.

antiagainst:
> Will wait for response about whether having it as a pass or utility function

I think making this as a pass makes sense for me. It's not that we are having wrong IRs in the middle; I think it as we are going from generic to specific and attaching more information along the way. (You can also view this as "lowering" I think.) Validations at different levels kicks in at different stages. :)

denis0x0D:
IMHO, it's better to have a pass then hook in some other places, because it would be easy for me to assign my own memory layout (in case I GPU vendor and know which layout gives the best performance) and therefore easy to maintain, I'll not have all that pain while syncing my copy with trunk version.
Thanks.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,329 @@
...
+/// 3.The ArrayStride, MatrixStride, and Offset decorations must be large
+/// enough to hold the size of the objects they affect (that is, specifying
+/// overlap is invalid).
+class SPIRVLayoutInfoUtils {
+public:
+  using Alignment = uint64_t;
+
+  /// Return a new spirv::StructType with layout info, increment a base size by
+  /// proccessed type size in bytes and populate an alignment.
+  static Type decorateType(spirv::StructType structType,
```
antiagainst: So these functions, with all the match calculations inside, can be quite hard to understand. What about explaining the meaning of `baseSize` and `baseAlignment` in comments?

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,329 @@
...
+  // According to the SPIR-V spec:
+  // 1."A two-component vector has a base alignment equal to twice its scalar
+  // alignment."
+  // 2."A three- or four-component vector has a base alignment equal to four
+  // times its scalar alignment."
+  auto vectorTypeAlignment =
+      numElements == 2 ? elementAlignment * 2 : elementAlignment * 4;
+
+  auto vectorSize = elementSize * numElements;
+  baseSize += llvm::alignTo(vectorSize, vectorTypeAlignment);
```
antiagainst: So instead of letting each type to figure out how it should fit in the struct type by updating the `baseSize` and `baseAlignment` for the struct type, what about letting them focus on themselves? I mean for each of these functions, we still have the `size` and `alignment` parameters to "return" size and alignment, but make it just the size and alignment of the *current* type? It would avoid us to have all these `alignTo` and `max` calls in each case (we just need them in struct type). it's also easier to reason about alignment and padding if they are all in struct type, not scattered everywhere.

test/Dialect/SPIRV/Transforms/global-decoration.mlir
```diff
@@ -0,0 +1,81 @@
+// RUN: mlir-opt -decorate-with-layoutinfo-spirv -split-input-file -verify-diagnostics %s -o - | FileCheck %s
```
antiagainst: Name it as layout-decoration.mlir to be clear?

test/Dialect/SPIRV/Transforms/global-decoration.mlir
```diff
@@ -0,0 +1,81 @@
...
+// -----
+
+spv.module "Logical" "GLSL450" {
+  // CHECK: spv.globalVariable @var0 : !spv.ptr<!spv.struct<!spv.struct<!spv.struct<!spv.struct<!spv.struct<i1 [0], i1 [1], f64 [2]> [0], i1 [16]> [0], i1 [24]> [0], i1 [32]> [0], i1 [40]>, Uniform>
+  spv.globalVariable @var0 : !spv.ptr<!spv.struct<!spv.struct<!spv.struct<!spv.struct<!spv.struct<i1, i1, f64>, i1>, i1>, i1>, i1>, Uniform>
+
+  // CHECK: spv.globalVariable @var1 : !spv.ptr<!spv.struct<!spv.struct<i16 [0], !spv.struct<i1 [0], f64 [1]> [2], f32 [18]> [0], f32 [24]>, Uniform>
+  spv.globalVariable @var1 : !spv.ptr<!spv.struct<!spv.struct<i16, !spv.struct<i1, f64>, f32>, f32>, Uniform>
+
+  // CHECK: spv.globalVariable @var2 : !spv.ptr<!spv.struct<!spv.struct<i16 [0], !spv.struct<i1 [0], !spv.array<16 x !spv.array<16 x i64 [8]> [128]> [1]> [2], f32 [2058]> [0], f32 [2064]>, Uniform>
```
antiagainst: Shouldn't `!spv.array<16 x !spv.array<16 x i64 [8]> [128]>` be at offset 8?

Also `!spv.struct<i1 [0], !spv.array<16 x !spv.array<16 x i64 [8]> [128]> [1]>` should be offset at 8?

denis0x0D: @antiagainst 
`!spv.array<16 x !spv.array<16 x i64 [8]> [128]>`
here is an array stride for array.

!spv.struct<i1 [0], !spv.array<16 x !spv.array<16 x i64 [8]> [128]> [1]>

```
spv.struct {
i1, {offset = 0}
!spv.array<...> {offset = 1}
}
```
I think the offset for the second member from the beginning of the struct is 1 byte? Do I miss something here?

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
```diff
@@ -0,0 +1,329 @@
...
+  // Change the type for the direct users.
+  target.addDynamicallyLegalOp<spirv::AddressOfOp>([](spirv::AddressOfOp op) {
+    return SPIRVLayoutInfoUtils::isLegalType(op.pointer()->getType());
+  });
+
+  // TODO: Change the type for the indirect users such as spv.Load, spv.Store,
+  // spv.FunctionCall and so on.
+
+  for (auto spirvModule : module.getOps<spirv::ModuleOp>()) {
+    // The Logical addressing model means pointers are abstract, having no
```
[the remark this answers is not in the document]
denis0x0D: @antiagainst please fix me if I'm wrong, as far as I understood the memory layout could depend on addressing model Physical32/64, Logical. So Logical does not allow to have a pointer inside a composite object, all pointers a logical, in physical addressing model we could have a pointer inside composite object and therefore the size of the pointer depends on the addressing model.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
denis0x0D: @antiagainst as far as I understood, please fix me if I'm wrong, the spec allows to have nested structs,
something like this:
 ` !spv.ptr<!spv.struct<!spv.array<16x!spv.struct<f32, f64, !spv.array<16xf32>>>, f32>, StorageBuffer>`

in this case top level struct size must be aligned to 8 bytes, because:
>  An array has a base alignment equal to the base alignment of its element type.

> A structure has a base alignment equal to the largest base alignment of any of its members.

So, we have to go like DFS here, and update base alignment for the parent type to the largest alignment of its childs and in the same way replace types without layout to types with layout? Do I miss something?

antiagainst: Yes the spec allows nested structs and we should recursively handle it. I meant that we don't need let child types to update for their parent type. Instead, they should just return their alignment and size and the parent type can do proper handling each child type. 

I think examples will make what I mean clear. You can check how layout is done in DXC at here: 

https://github.com/microsoft/DirectXShaderCompiler/blob/3e302950dd507e9c433705111ac0de1b143d6d1b/tools/clang/lib/SPIRV/AlignmentSizeCalculator.cpp#L66

Let me know if it's still not clear. Happy to clarify. :)

test/Dialect/SPIRV/Transforms/global-decoration.mlir
antiagainst: An array has a base alignment equal to the base alignment of its element type. Here the array is containing `i64`, which means the array's base alignment is 8. So it should be aligned to 8, right?

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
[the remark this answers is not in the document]
antiagainst: That's right. But it should not be the concern of this pass I think. We should have validations for the Vulkan execution environment (or more specifically, Logical addressing model) to enforce that. :)

test/Dialect/SPIRV/Transforms/global-decoration.mlir
denis0x0D: oh, this is my fault, I've misunderstood it, I was thinking about composite size alignment and not about the start address inside the struct. Thanks a lot for mention this!

joker-eph:
> To unblock this work, can we reach a consensus. If having a pass is problematic,

I only saw this problematic if this was meant to be a verifier failure, which it isn't apparently, so this voids any concerns I had with having it as a pass.

joker-eph:
> > If you present the capabilities and/or the layout are purely redundant with the IR, I'd argue to strip them entirely on import and add them on export.
> > If you can't strip them, why? Are they meaningful then? If they are then the corollary seems like passes would need to maintain these informations no?
> 
> I think it depends on what we are talking about.
> 
> The layout information is decorated onto the type (both in the spec and in the SPIR-V dialect). If we have two structs with identical members but different layout, they are two different types inside SPIR-V.

I already see a problem with the wording here, you wrote that "the layout information is decorated onto the type", from which I take that that layout is not part of the type ; but then added that "two structs with identical members but different layout [..] are two different types". 

> If we strip them out during import and then re-export, we can generate modules that do not match with the original layout.

If you add layout on "layout less" types, then you are also not preserving the ability to generate modules that match the original layout.


> Besides, there can exist [multiple layout schemes](https://github.com/Microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#memory-layout-rules) and for different storage classes we can have different rules. If we strip them out during import and then re-export, we can generate modules that do not match with the original layout.

This link is interesting, it indicates that the *frontend* has to pick the layout when lowering to SPIR-V if I understand correctly. How can this be compatible with this pass then?

> 
> For layout I think it does not matter for passes since the layout (and other decorations on types) is baked into the type itself. So passes can operate on `spv.struct`s with or without layout info similarly.

This is not clear to me. Canonicalization would need to account for the layout and properly propagate it / take it into account when rewriting operation / changing type (mem2reg kind of transformations for instance).


> I think we just need a dedicated pass to turn "layoutless" types into ones with layout

You haven't made it clear why is "layout less" even a thing at this point? I think this is probably one critical piece I am missing right now.

> (it will need CLI options for which layout scheme to use from the developer) at the final stage when we know about the execution environment so that we generate valid modules according to that execution environment.

It isn't clear to me that the layout isn't a contract with the host at some point, and that you can just do this...

denis0x0D:
@antiagainst @MaheshRavishankar thanks a lot for review, updated according to the comments.

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
denis0x0D: @antiagainst thanks for the explanation, I hope I understood it right :)

lib/Dialect/SPIRV/Transforms/DecorateWithLayoutInfoSPIRVPass.cpp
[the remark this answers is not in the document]
denis0x0D: @antiagainst thanks for the explanation, the check is removed.

antiagainst:
@joker-eph:

> I already see a problem with the wording here, you wrote that "the layout information is decorated onto the type", from which I take that that layout is not part of the type ; but then added that "two structs with identical members but different layout [..] are two different types".

I think the terminology might be confusing here. Sorry about that!

Decoration is a SPIR-V spec term. Using struct type as an example, in the binary format, the main type is defined using `OpTypeStruct`. The layout information is decorated to the result id generated from `OpTypeStruct` with `OpDecorate`/`OpMemberDecorate`:

```spirv
OpDecorate %0 Block
OpMemberDecorate %0 0 0
OpMemberDecorate %0 1 4
...
%int = OpTypeInt 32 0
%0 = OpTypeStruct %int %int
...
```
These decorations are semantic impacting. Although represented using different instructions, they should be considered as part of the type. We cannot load a value from a pointer to the above type as a struct without layout decorations. Structs with different layout decorations are different types.

In the SPIR-V dialect, such decorations are [baked into](https://github.com/tensorflow/mlir/blob/ba60cccff7f1fc559862c64da42a591cb97e9312/lib/Dialect/SPIRV/SPIRVTypes.cpp#L450-L453) the `spirv::StructType`: we have fields for them in the `StructTypeStorage`. The above is represented as

```mlir
!spv.struct<i32 [0], i32 [4]>
```

> You haven't made it clear why is "layout less" even a thing at this point? I think this is probably one critical piece I am missing right now.

Let me rephrase and provide more context to make it clear. :)

SPIR-V is intended to serve multiple APIs; Vulkan and OpenCL are two of them.

For Vulkan, the `Shader` capability is required. The `Shader` capability requires "Composite objects in the StorageBuffer, PhysicalStorageBuffer, Uniform, and PushConstant Storage Classes must be explicitly laid out." ([Spec 2.16.2](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_shadervalidation_a_validation_rules_for_shader_a_href_capability_capabilities_a)). For OpenCL, the `Kernel` capability is required. The `Kernel` capability does **not** have the above requirement, i.e., composite types in those storage classes are not required to be laid out.

So, "layoutless" is the way for OpenCL. Passes for OpenCL-flavored SPIR-V operate on struct types without layout information. For Vulkan, the purpose of these layout information decoration (SPIR-V spec term again) is to match with the data layout in memory, which is set up and populated using C++ API. These information might be useful for driver compilers given that they will further compile SPIR-V into architecture-specific ISA with physical pointers, but they should not affect how transformations inside MLIR work, which is on logical addressing model. They are just additional information put in the SPIR-V module for further consumers (driver compilers) to consider when working together with the application code (written in C++). In a sense, it's like saying one function in the SPIR-V module is an entry point with name `foo` so the driver compiler can properly expose it when compiling the shader for compute pipeline. We just need to add such information at the end of the conversion flow inside MLIR (going from a high-level dialect).  Normal passes should not care about it; they can just operate on a "layoutless" version.

Does the above make sense?

> This is not clear to me. Canonicalization would need to account for the layout and properly propagate it / take it into account when rewriting operation / changing type (mem2reg kind of transformations for instance).

Is this resolved with the explanation in the above?

> This link is interesting, it indicates that the frontend has to pick the layout when lowering to SPIR-V if I understand correctly. How can this be compatible with this pass then?

From a compiler's perspective, I don't think it's the frontend that should pick the layout; it's the SPIR-V backend. (The link I posted previously is for the SPIR-V backend inside DXC.)

How data is laid out in memory is external knowledge to the compiler. The compiler won't know at all if without input from the developer: the developer may want to use Vulkan rules to lay out the data, or may want to use DirectX rules (for interop with DirectX), or OpenGL rules (for interop with OpenGL), or whatever. These different rule set calculation and decoration should be provided as libraries so the developer can choose and use to match the layout they want to use on API/runtime side. This pass can be such a library for this purpose. Yes right now we only support Vulkan rules because that's what we are interested currently, but in the future I think this pass will be extended to support more rule sets. 

> It isn't clear to me that the layout isn't a contract with the host at some point, and that you can just do this...

Hopefully the above provides more context. Yes it's a contract with the host, so we need to get input from the developer and put the **external** information inside the module so further SPIR-V consumers have access to those information. Am I answering your question here?

> If you add layout on "layout less" types, then you are also not preserving the ability to generate modules that match the original layout.

I'm not sure I get your question here fully. To target Vulkan, "layoutless" is more of an intermediate state. We need to add layout information at the end of the SPIR-V lowering. If we are importing a real-world SPIR-V blob, then those layout should already be inside the module and we should preserve them.

antiagainst:
Sorry, clicked the wrong button...

lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
```diff
@@ -0,0 +1,318 @@
...
+    return structType;
+  }
+
+  llvm::SmallVector<Type, 4> memberTypes;
+  llvm::SmallVector<spirv::StructType::LayoutInfo, 4> layoutInfo;
+  llvm::SmallVector<spirv::StructType::MemberDecorationInfo, 4>
+      memberDecorations;
+
+  spirv::StructType::LayoutInfo structMemberOffset = 0;
+  spirv::StructType::LayoutInfo memberSize = 0;
```
antiagainst: Define `memberSize` and `memberAlignment` inside the for loop so we don't need to reset them at the end?

lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
```diff
@@ -0,0 +1,318 @@
...
+    // According to the Vulkan spec:
+    // "A structure has a base alignment equal to the largest base alignment of
+    // any of its members."
+    structMemberOffset += memberSize;
+    maxMemberAlignment = std::max(maxMemberAlignment, memberAlignment);
+    memberAlignment = 1;
+    memberSize = 0;
+  }
+
+  size = llvm::alignTo(structMemberOffset, maxMemberAlignment);
```
antiagainst: Is this correct? Why we are adjusting the size with alignment? (We need that when placing the struct inside another struct, but that's adjusting the offset instead of the struct size itself and it should be handled by the outer struct's member layout logic, right?)

lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
```diff
@@ -0,0 +1,318 @@
...
+  // According to the Vulkan spec:
+  // 1. "A two-component vector has a base alignment equal to twice its scalar
+  // alignment."
+  // 2. "A three- or four-component vector has a base alignment equal to four
+  // times its scalar alignment."
+  auto vectorTypeAlignment =
+      numElements == 2 ? elementAlignment * 2 : elementAlignment * 4;
+
+  auto vectorSize = elementSize * numElements;
+  size = llvm::alignTo(vectorSize, vectorTypeAlignment);
```
antiagainst: Here too. This method is for vector type itself and return the real size for the vector. It should not worry about padding to match the alignment. That should be the responsibility of an outer type. (And the outer type may act differently regarding padding.)

lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
denis0x0D: @antiagainst thanks for review, I've looked at this rule https://github.com/microsoft/DirectXShaderCompiler/blob/3e302950dd507e9c433705111ac0de1b143d6d1b/tools/clang/lib/SPIRV/AlignmentSizeCalculator.cpp#L324

but I might be wrong with an interpretation, please fix me in this case.

If we do not align struct size to the max alignment of its member, we will get:
` !spv.ptr<!spv.struct<!spv.struct<!spv.struct<i1 [0], i64 [8]> [0], i1 [16]> [0], i1 [17]>, Uniform>`

with alignment for struct size:
` !spv.ptr<!spv.struct<!spv.struct<!spv.struct<i1 [0], i64 [8]> [0], i1 [16]> [0], i1 [24]>, Uniform>`

It's similiar to c++ rules I think, but I'm not sure that rules are applicable for spir-v alignment.
```
struct A {
  bool a;
  uint64_t b;
};

struct B {
  A a;
  bool b;
};

struct C {
  B b;
  bool c;
};

sizeof (A) == 16
sizeof (B) == 24
sizoef (C) == 32
```

Thanks.

MaheshRavishankar: Agree with @denis0x0D here. The only way to make sure that you satisfy the alignment properties of all elements of a struct, is to make sure that the struct itself is aligned to the maximum alignment of any element of the struct.

denis0x0D:
rebased on master to avoid "conflicting file" issue.

lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
antiagainst: Ah yes, sorry forgot about that rule. :) Could you quote the spec here? "The Offset decoration of a member must not place it between the end of a structure or an array and the next multiple of the alignment of that structure or array."
 
@MaheshRavishankar: it's a slightly different issue here. The struct member alignment is satisfied at Line 98. Here we are talking about the element following a struct inside another struct.

lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
```diff
@@ -0,0 +1,316 @@
...
+/// resources and on the features enabled on the device."
+///
+/// There are 3 types of alignment: scalar, base, extended.
+/// See the spec for details.
+///
+/// Note: Even if scalar alignment is supported, it is generally more
+/// performant to use the base alignment. So here the calculation is based on
+/// base alignment.
+///
+/// The memory layout must obey the following rules:
```
antiagainst: These rules are not from the current Vulkan spec. Could you use the most recent version?

https://renderdoc.org/vkspec_chunked/chap14.html#interfaces-resources-layout

lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
```diff
@@ -0,0 +1,316 @@
...
+    memberTypes.push_back(memberType);
+    layoutInfo.push_back(structMemberOffset);
+    // According to the Vulkan spec:
+    // "A structure has a base alignment equal to the largest base alignment of
+    // any of its members."
+    structMemberOffset += memberSize;
+    maxMemberAlignment = std::max(maxMemberAlignment, memberAlignment);
+  }
+
+  size = llvm::alignTo(structMemberOffset, maxMemberAlignment);
```
antiagainst: Could you quote the spec here? "The Offset decoration of a member must not place it between the end of a structure or an array and the next multiple of the alignment of that structure or array."

lib/Dialect/SPIRV/Transforms/DecorateSPIRVCompositeTypeLayoutPass.cpp
denis0x0D: @antiagainst oh, sorry, its a mix from Vulkan spec and SPIR-V spec.

antiagainst:
One final nit, could you update and elaborate the commit message a bit? Thanks!

denis0x0D:
@antiagainst updated, thanks!

denis0x0D:
@antiagainst rebased on master to follow up the commit https://github.com/tensorflow/mlir/commit/cea968877b68b0bca7d7e40cbae2e6cac741aec4

joker-eph:
> > You haven't made it clear why is "layout less" even a thing at this point? I think this is probably one critical piece I am missing right now.
> 
> Let me rephrase and provide more context to make it clear. :)
> 
> SPIR-V is intended to serve multiple APIs; Vulkan and OpenCL are two of them.
> 
> For Vulkan, the `Shader` capability is required. The `Shader` capability requires "Composite objects in the StorageBuffer, PhysicalStorageBuffer, Uniform, and PushConstant Storage Classes must be explicitly laid out." ([Spec 2.16.2](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_shadervalidation_a_validation_rules_for_shader_a_href_capability_capabilities_a)). For OpenCL, the `Kernel` capability is required. The `Kernel` capability does **not** have the above requirement, i.e., composite types in those storage classes are not required to be laid out.

OK, but at this point the world seems fairly static: if you have the kernel capability you don't need layouts, otherwise you do.
We're not yet to a point where we need to add a "default" one generically in the SPIR-V dialect itself.

I'm also curious if this is only valid in OpenCL for structure that don't communicate with the host and are only used on the device? Otherwise it isn't clear to me how do you get the data back on the host.

> but they should not affect how transformations inside MLIR work, which is on logical addressing model.

It isn't clear to me that this is entirely true: are you saying that I can't emit SPIR-V code that includes some sort of casting that would rely on the layout being preserved?
In C/C++, when struct are layout by clang into LLVM, the code is definitely not operating symbolically (in the simpler case: you can cast a pointer to a struct into an i8* and index into any member using the layout).

> They are just additional information put in the SPIR-V module for further consumers (driver compilers) to consider when working together with the application code (written in C++).

Aren't they *contract* with the application code?


> > This is not clear to me. Canonicalization would need to account for the layout and properly propagate it / take it into account when rewriting operation / changing type (mem2reg kind of transformations for instance).
> 
> Is this resolved with the explanation in the above?

Not really: in LLVM the layout is used and taken into account by the passes (using the DataLayout). 
Years ago LLVM was trying to be agnostic and have passes be able to operate without datalayout, but in practice it didn't work out well (when you get into details of assumptions analysis have to make).


> > This link is interesting, it indicates that the frontend has to pick the layout when lowering to SPIR-V if I understand correctly. How can this be compatible with this pass then?
> 
> From a compiler's perspective, I don't think it's the frontend that should pick the layout; it's the SPIR-V backend. (The link I posted previously is for the SPIR-V backend inside DXC.)

This does not make sense to me, so I probably misunderstand what you wrote above.
If the layout is a contract with the application code, then this contract has to be expressed by the application and cannot be picked a GPU driver independently of the application, otherwise in case of mismatch the host couldn't read the data.
That is unless struct aren't exposed to the application code directly and 


> How data is laid out in memory is external knowledge to the compiler. The compiler won't know at all if without input from the developer: the developer may want to use Vulkan rules to lay out the data, or may want to use DirectX rules (for interop with DirectX), or OpenGL rules (for interop with OpenGL), or whatever. 

This also seems to me to contradict the "I don't think it's the frontend that should pick the layout; it's the SPIR-V backend". Unless a SPIR-V kernel is supposed to be portable across Vulkan/OpenGL/OpenCL and only at runtime when you load SPIR-V into the driver you pick your model? (I doubt it considering how much specificity for each use-case is baked into the IR).

> > It isn't clear to me that the layout isn't a contract with the host at some point, and that you can just do this...
> 
> Hopefully the above provides more context. Yes it's a contract with the host, so we need to get input from the developer and put the **external** information inside the module so further SPIR-V consumers have access to those information. Am I answering your question here?

You're partially answering, but this is leaving me very confused about the whole story: it just does not line up for me at the moment.

> > If you add layout on "layout less" types, then you are also not preserving the ability to generate modules that match the original layout.
> 
> I'm not sure I get your question here fully. To target Vulkan, "layoutless" is more of an intermediate state. We need to add layout information at the end of the SPIR-V lowering. If we are importing a real-world SPIR-V blob, then those layout should already be inside the module and we should preserve them.

What I was saying is that if you import a SPIR-V module that is "layoutless" and you always add layout as part of your flow, then your not preserving the "layoutless" aspect of it.

antiagainst:
> OK, but at this point the world seems fairly static: if you have the kernel capability you don't need layouts, otherwise you do.
We're not yet to a point where we need to add a "default" one generically in the SPIR-V dialect itself.

I'm not sure I get your idea here. Having layout information absent does not mean we force a "default" layout; it's simply we don't have such information.

> I'm also curious if this is only valid in OpenCL for structure that don't communicate with the host and are only used on the device? Otherwise it isn't clear to me how do you get the data back on the host.

I'm less familiar with the OpenCL-flavored SPIR-V. My understanding is that OpenCL has relatively well defined layout [rule](https://www.khronos.org/registry/OpenCL/sdk/1.1/docs/man/xhtml/attributes-types.html) (following C/C++) and that can be represented with a single decoration, e.g. [`CPacked`](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_decoration_a_decoration), on the struct type. So there is no need to explicitly specify the offset of each struct member. On graphics side because of historical reasons and interop between different APIs, we have multiple rule sets and lots of flexibility here so need to explicitly specify everything.

> It isn't clear to me that this is entirely true: are you saying that I can't emit SPIR-V code that includes some sort of casting that would rely on the layout being preserved?
In C/C++, when struct are layout by clang into LLVM, the code is definitely not operating symbolically (in the simpler case: you can cast a pointer to a struct into an i8* and index into any member using the layout).

You can do pointer arithmetics to index into a struct/array member (which requires layout information) for OpenCL-flavored SPIR-V.  But Vulkan supports logical addressing model; it has much restricted pointer usage. You cannot cast a struct/array pointer to an integer and then do pointer arithmetics to get its member. (You can find the list of casting ops in [3.32.11. Conversion Instructions](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_conversion_a_conversion_instructions). Note that those related to pointer <-> integer casting requires `Kernel` or `Addresses` capability, which is for OpenCL.)  The way to index into a struct is use `OpAccessChain`, which is still logical addressing that layout does not matter (at SPIR-V level).

> Aren't they contract with the application code?

Yes, they are contract with the application. So they are useful for further SPIR-V consumers (the driver compiler) when trying to "link" with the application. :) I don't quite see what's the issue with my original wording? 

> Not really: in LLVM the layout is used and taken into account by the passes (using the DataLayout).
Years ago LLVM was trying to be agnostic and have passes be able to operate without datalayout, but in practice it didn't work out well (when you get into details of assumptions analysis have to make).

Oh, I think I understand your concerns better now by knowing the LLVM history on this. Yeah, C/C++ compilers cannot avoid the layout issue because it's rooted in the language. I guess this will be more of an issue if we want to fully support OpenCL, which has a language more akin to C/C++. For Vulkan, shader languages are fairly restricted and as said before, you cannot be very creative on pointers so it's less a concern.

However, I still see the benefits of this functionality being a pass. While we are focusing on Vulkan right now, it can serve both Vulkan and OpenCL. Being a pass means it can be chosen to apply at a suitable time. For OpenCL maybe we can apply it at the very beginning so following OpenCL specific passes can have concrete layout to operate on. For Vulkan, we just need it at the very last to make sure these "contract information for application" from the developer is carried into the final module. 

> This does not make sense to me, so I probably misunderstand what you wrote above.
If the layout is a contract with the application code, then this contract has to be expressed by the application and cannot be picked a GPU driver independently of the application, otherwise in case of mismatch the host couldn't read the data.
That is unless struct aren't exposed to the application code directly and

Yes it's a contract between the shader program and the application and yes they need to match. (Actually lots of issues/crashes are because they don't match. ;-P) But we are **not** letting the GPU driver to pick a layout independently of the application. Instead, it's the developer, who writes both the program and the application, provides the information to the shader compiler to get the information embedded in the final module and then the driver pick this information up. From the shader program and its compiler's perspective, how data is put in memory on the application side is completely unknown; only the developer knows  that information. So we are requiring the developer to specify that information via a CLI option (or a pass creation parameter or whatever equivalent).

> This also seems to me to contradict the "I don't think it's the frontend that should pick the layout; it's the SPIR-V backend". Unless a SPIR-V kernel is supposed to be portable across Vulkan/OpenGL/OpenCL and only at runtime when you load SPIR-V into the driver you pick your model? (I doubt it considering how much specificity for each use-case is baked into the IR).

Maybe I'm using some terminology wrongly and results in confusion... My previous comment was regarding DXC implementation specifically. The layout selection happens in the SPIR-V CodeGen there. But I guess yes you can also specify the layout directly in the source code so that qualifies as selecting the layout in a "frontend". (Although if talking about frontend I think more about compiler frontend that turns a source code into an AST, instead of the source code itself.)

> You're partially answering, but this is leaving me very confused about the whole story: it just does not line up for me at the moment.

I'm sorry for the confusion! I'm actually also a bit confused about what's your deepest concerns after we've clarified that the input to this pass should not be considered as invalid. We should have some time to chat over this. :) 

> What I was saying is that if you import a SPIR-V module that is "layoutless" and you always add layout as part of your flow, then your not preserving the "layoutless" aspect of it.

Existing SPIR-V blobs for Vulkan must be laid out, so such case (importing a "layoutless" module) should not exist.

joker-eph:
> > OK, but at this point the world seems fairly static: if you have the kernel capability you don't need layouts, otherwise you do.
> > We're not yet to a point where we need to add a "default" one generically in the SPIR-V dialect itself.
> 
> I'm not sure I get your idea here. Having layout information absent does not mean we force a "default" layout; it's simply we don't have such information.

I am saying that if the capability requires a layout, then you should already have this information in the IR, and you shouldn't need a pass that adds it after.
If your capability does not require a layout, then you don't need a pass to add it.

> 
> > Aren't they contract with the application code?
> 
> Yes, they are contract with the application. So they are useful for further SPIR-V consumers (the driver compiler) when trying to "link" with the application. :) I don't quite see what's the issue with my original wording?

I see a contradiction between saying that we can add a layout after the fact and the claim that it is a contract.  The fact that it is a contract with the host application implies to me that the lowering from the GPU dialect to SPIR-V dialect *must* provide a layout.

> > This does not make sense to me, so I probably misunderstand what you wrote above.
> > If the layout is a contract with the application code, then this contract has to be expressed by the application and cannot be picked a GPU driver independently of the application, otherwise in case of mismatch the host couldn't read the data.
> > That is unless struct aren't exposed to the application code directly and
> 
> Yes it's a contract between the shader program and the application and yes they need to match. 

I sniped most of what you wrote, but I'm still reading a contradiction in all the parts I omitted and this last sentence.

> (Actually lots of issues/crashes are because they don't match. ;-P) But we are **not** letting the GPU driver to pick a layout independently of the application. Instead, it's the developer, who writes both the program and the application, provides the information to the shader compiler to get the information embedded in the final module and then the driver pick this information up. From the shader program and its compiler's perspective, how data is put in memory on the application side is completely unknown; only the developer knows that information. So we are requiring the developer to specify that information via a CLI option (or a pass creation parameter or whatever equivalent).

Right, so again: this informations comes from the higher level frontend/framework during lowering, and it shouldn't be added after the fact, since it is a contract with "something else" (the application code) which isn't visible from the SPIR-V module.
I still don't see the use-case for this pass right now.


> 
> > This also seems to me to contradict the "I don't think it's the frontend that should pick the layout; it's the SPIR-V backend". Unless a SPIR-V kernel is supposed to be portable across Vulkan/OpenGL/OpenCL and only at runtime when you load SPIR-V into the driver you pick your model? (I doubt it considering how much specificity for each use-case is baked into the IR).
> 
> Maybe I'm using some terminology wrongly and results in confusion... My previous comment was regarding DXC implementation specifically. The layout selection happens in the SPIR-V CodeGen there. But I guess yes you can also specify the layout directly in the source code so that qualifies as selecting the layout in a "frontend". (Although if talking about frontend I think more about compiler frontend that turns a source code into an AST, instead of the source code itself.)

By frontend here, I mean anything that generates the SPIR-V dialect (just like when you come from a higher-level, targeting something like SPIR-V is a backend): multi-layer is hard :)


> > What I was saying is that if you import a SPIR-V module that is "layoutless" and you always add layout as part of your flow, then your not preserving the "layoutless" aspect of it.
> 
> Existing SPIR-V blobs for Vulkan must be laid out, so such case (importing a "layoutless" module) should not exist.

OK, so I think at least one thing is clear: this excludes that this pass has any use when the input is a SPIR-V binary.

antiagainst:
More on this. :) After the above comments and an initial offline chat, I think the questions still remaining are:

1) Where does this layout information come from.
2) What's the use case for this pass.

Here are my thoughts on these two questions:

For 1), the layout information comes from an external source. By external, I mean external to the SPIR-V module itself. We are not inferring the layout information by using the SPIR-V module alone; it is given from an external source.

Because of progressive implementation, the current implementation of this pass does not reveal that, which I guess is causing part of the confusion. It should have an CLI option to let developers specify which layout rule to use in the future. Or as discussed and suggested offline by @joker-eph, the information can also be an attribute attached to some high-level dialect coming as input to SPIR-V lowering, assuming there is an "driver" (like IREE) coordinating both the kernel and runtime side. Then the driver can decide which layout it want to use on runtime side and put that information onto the function to be compiled into SPIR-V module.

Still, no matter which approach to take here (CLI options or function attributes), they are both external inputs to this pass.

For 2), we need to think about a few workflows.

If importing an existing SPIR-V blob, as said previously, it should already be laid out. So we don't need this pass here.

If going from some high-level dialect to SPIR-V for ML, then it falls into the case I said for 1). If we have a function attribute or something from the high-level dialect, then great we can use that. Here we still have a use case for this pass: it can consume that attribute and decorate with explicit layout. The benefits I see is we have better separated concerns and more clear boundaries. The lowering pass can just focus on turning a function into a `spv.module` and let this pass worry about all the layout calculation details. This pass is just another step in the lowering procedure and it's also reusable for other purposes like the following. 

If we are using the SPIR-V dialect inside a shading language (like HLSL) compiler, then the input shading language may not contain layout rule. (This is the case as of today in HLSL. It is a language for DirectX but because it's widely used we also want to adopt it for Vulkan. So unsurprisingly, Vulkan specific annotations are missing in the source code sometimes. Even say in the future we add it, vendors are not always free to use because those annotations will be new and not accepted by existing HLSL compilers like `fxc.exe` so causing trouble for targeting multiple APIs.) Then we have no where to get such information; we can only require the developer to specify it via a CLI option. Then a pass like this is needed to calculate and attach the explicit layout according to the rule.

joker-eph:
Since I still don't understand the flow here, I think we should raise this to the mailing-list and clarify. I'll start an email thread there.

Thanks for all the information though.

joker-eph:
> If we are using the SPIR-V dialect inside a shading language (like HLSL) compiler, then the input shading language may not contain layout rule.

To me this is the role of the HLSL compiler itself when in the IR it create the SPIR-V type to provide the layout in a way that is compatible with the execution environment (just like Clang layout C struct based on the target ABI, LLVM does not do it later)

PR #157, opened 2019-09-27
fix simplify-affine-structures bug

bondhugula:
- normalizeMemRefs was being called online while walking over ops - the former erases/replaces ops

### DIFF
--- a/lib/Transforms/SimplifyAffineStructures.cpp
+++ b/lib/Transforms/SimplifyAffineStructures.cpp
@@ -102,9 +102,16 @@ void SimplifyAffineStructures::runOnFunction() {
     }
   });
 
-  // Turn memrefs' non-identity layouts maps into ones with identity.
-  func.walk([](AllocOp op) { normalizeMemRef(op); });
+  // Turn memrefs' non-identity layouts maps into ones with identity. Collect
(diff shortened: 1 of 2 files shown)


lib/Transforms/SimplifyAffineStructures.cpp
```diff
@@ -102,8 +102,14 @@ void SimplifyAffineStructures::runOnFunction() {
...
   });
 
-  // Turn memrefs' non-identity layouts maps into ones with identity.
-  func.walk([](AllocOp op) { normalizeMemRef(op); });
+  // Turn memrefs' non-identity layouts maps into ones with identity. Collect
+  // alloc ops first and then process since normalizeMemRef replaces/erases ops
+  // during memref rewriting.
+  SmallVector<AllocOp, 4> allocOps;
+  func.walk([&](AllocOp op) { allocOps.push_back(op); });
+  for (auto allocOp : llvm::make_early_inc_range(allocOps)) {
```
joker-eph: You don't need `make_early_inc_range` here as you're not modifying the `allocOps` vector itself.

bondhugula: Fixed, thanks.

PR #160, opened 2019-09-28
[spirv] Add array length check.

denis0x0D:
According to the SPIR-V spec:
"Length is the number of elements in the array. It must be at least 1."

Thanks!

### DIFF
--- a/lib/Dialect/SPIRV/SPIRVTypes.cpp
+++ b/lib/Dialect/SPIRV/SPIRVTypes.cpp
@@ -56,12 +56,14 @@ struct spirv::detail::ArrayTypeStorage : public TypeStorage {
 };
 
 ArrayType ArrayType::get(Type elementType, unsigned elementCount) {
+  assert(elementCount && "ArrayType needs at least one element");
   return Base::get(elementType.getContext(), TypeKind::Array, elementType,
                    elementCount, 0);
 }
 
 ArrayType ArrayType::get(Type elementType, unsigned elementCount,
                          ArrayType::LayoutInfo layoutInfo) {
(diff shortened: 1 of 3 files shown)


lib/Dialect/SPIRV/SPIRVTypes.cpp
```diff
@@ -56,12 +56,14 @@ struct spirv::detail::ArrayTypeStorage : public TypeStorage {
 };
 
 ArrayType ArrayType::get(Type elementType, unsigned elementCount) {
+  assert(elementCount && "ArrayType needs at least one element type");
```
antiagainst: "... at least one element" ?

lib/Dialect/SPIRV/SPIRVTypes.cpp
```diff
@@ -56,12 +56,14 @@ struct spirv::detail::ArrayTypeStorage : public TypeStorage {
...
   return Base::get(elementType.getContext(), TypeKind::Array, elementType,
                    elementCount, 0);
 }
 
 ArrayType ArrayType::get(Type elementType, unsigned elementCount,
                          ArrayType::LayoutInfo layoutInfo) {
+  assert(elementCount && "ArrayType needs at least one element type");
```
antiagainst: Same here

lib/Dialect/SPIRV/SPIRVTypes.cpp
denis0x0D: Thanks, fixed! I was looking on struct type assert :)

`  assert(!memberTypes.empty() && "Struct needs at least one member type");`

PR #161, opened 2019-09-29
Fix syntax of 'call' and 'splat' ops

bondhugula:
- fix missing return value syntax on call / splat ops
- reflow cond_br / store op syntax


joker-eph:
We have some doc duplication with the ODS definition here, we need to spend time reconciling and remove this doc in favor of auto-generated one.

PR #163, opened 2019-10-02
Moving the GPUIndexIntrinsicOpLowering template to a common location

deven-amd:
The GPUIndexIntrinsicOpLowering template is currently is used by the code in both the GPUToNVVM and GPUToROCDL dirs. Moving it to a common location to remove code duplication.

This is a followup to the request made in PR #154 

------------------------

@joker-eph @sherhut . Not quite sure whether I got this right...please review.

### DIFF
--- a/lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h
+++ b/lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h
@@ -1,0 +1,90 @@
+//===- IndexIntrinsicsOpLowering.h - GPU IndexOps Lowering class *- C++ -*-===//
+//
+// Copyright 2019 The MLIR Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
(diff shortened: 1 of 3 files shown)


lib/Conversion/GPUToNVVM/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+include_directories(../GPUCommon)
```
sherhut: It would be preferable to avoid `include_directories` to make it easier to integrate this with bazil builds. Instead, please use relative paths in the `#include` directive.

lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h
```diff
@@ -0,0 +1,88 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+#ifndef MLIR_CONVERSION_GPUCOMMON_INDEXINTRINSICSOPLOWERING_H_
+#define MLIR_CONVERSION_GPUCOMMON_INDEXINTRINSICSOPLOWERING_H_
+
+namespace mlir {
+
```
sherhut: It would be nice to make this self contained. An include of `ConvertStandardToLLVM.h` should do the trick I think.

lib/Conversion/GPUToNVVM/CMakeLists.txt
deven-amd: pushed out a new commit with the requested change

lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h
deven-amd: pushed out a new commit with the requested change

ftynse:
@deven-amd could you please rebase?

PR #166, opened 2019-10-03
Interfaces

sana-damani:
Inliner currently does not work:
1. doesn't handle generic calls
2. adding new pass results in build break

Also need feedback on how to test shape inference.


sana-damani:
Pull request opened on accident. Closing.

PR #169, opened 2019-10-07
[ROCm] Adding the ROCM backend implementation for GPU dialect + the mlir-rocm-runner utility

deven-amd:
This PR  adds
 * the ROCM backend for the GPU Dialect
 * the mlir-rocm-runner utiliy (that can "run" mlir codes with GPU dialect on AMD GPUs)

The ROCM backend support for AMD GPUs is similar in concept and implementation to the CUDA backend support for NV GPUs.

--------------

@joker-eph @whchung


lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,408 @@
...
+              const std::string &rocdlDir) {
+
+  // AMDGPU version-neutral bitcodes.
+  static std::vector<std::string> *rocdlFilenames =
+      new std::vector<std::string>({"hc.amdgcn.bc", "opencl.amdgcn.bc",
+                                    "ocml.amdgcn.bc", "ockl.amdgcn.bc",
+                                    "oclc_finite_only_off.amdgcn.bc",
+                                    "oclc_daz_opt_off.amdgcn.bc",
+                                    "oclc_correctly_rounded_sqrt_on.amdgcn.bc",
+                                    "oclc_unsafe_math_off.amdgcn.bc"});
```
whchung: @deven-amd need to add `oclc_wavefrontsize64_on.amdgcn.bc`. refer to:

https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/pull/590/files

deven-amd: done

include/mlir/Conversion/GPUToROCM/GPUToROCMPass.h
```diff
@@ -0,0 +1,102 @@
...
+static constexpr const char *kHSACOGetterSuffix = "_hsaco";
+static constexpr const char *kHSACOStorageSuffix = "_hsaco_cst";
+
+/// enum to represent the AMD GPU versions supported by the ROCM backend
+enum class AMDGPUVersion { GFX900 };
+
+/// enum to represent the HSA Code Object versions supported by the ROCM backend
+enum class HSACOVersion { V3 };
+
+/// Configurable parameters for geenrating the HSACO blobs from GPU Kernels
```
joker-eph: "generating"

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+/// A pass converting tagged kernel functions to HSA Code Object blobs.
+///
+/// If tagged as a kernel module, each contained function is translated to ROCDL
+/// IR, which is then compiled using the llvm AMDGPU backend to generate the GPU
+/// binary code (i.e. the HSACO file). The HSACO binary blob is attached as an
+/// attribute to the function and the function body is erased.
+class GpuKernelToHSACOPass : public ModulePass<GpuKernelToHSACOPass> {
+public:
+  GpuKernelToHSACOPass(rocm::HSACOGeneratorConfig hsacoGeneratorConfig =
+                           rocm::HSACOGeneratorConfig(/*isTestMode*/ true))
```
River707: /*...=*/

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+    // This is a module containing a GPU kernel function, we have work to do!
+
+    // Make sure the AMDGPU target is initialized.
+    LLVMInitializeAMDGPUTarget();
+    LLVMInitializeAMDGPUTargetInfo();
+    LLVMInitializeAMDGPUTargetMC();
+    LLVMInitializeAMDGPUAsmPrinter();
+
+    auto llvmModule = translateModuleToROCDLIR(getModule());
+    if (!llvmModule) {
```
River707: Drop trivial braces when possible.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+    for (auto function : getModule().getOps<FuncOp>()) {
+      if (!gpu::GPUDialect::isKernel(function)) {
+        continue;
+      }
+
+      if (failed(translateGpuKernelToHSACOAnnotation(*llvmModule, function)))
+        signalPassFailure();
+    }
+
+    // dump the MLIR
```
River707: Can we have a more descriptive comment here? Also, why have this in the first place? We have pass manager instrumentation for printing the IR after passes, and environment variables are generally discouraged for this type of thing.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+  rocm::HSACOGeneratorConfig config;
+};
+
+// get the "-mcpu" option string corresponding to the given AMDGPU version enum
+std::string GetMcpuOptionString(rocm::AMDGPUVersion v) {
+  switch (v) {
+  case rocm::AMDGPUVersion::GFX900:
+    return "gfx900";
+    break;
```
River707: The 'break's aren't necessary on any of these.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+}
+
+// Gets the ROCm-Device-Libs filenames for a particular AMDGPU version.
+static std::vector<std::string>
+GetROCDLPaths(const rocm::AMDGPUVersion amdgpuVersion,
+              const std::string &rocdlDir) {
+
+  // AMDGPU version-neutral bitcodes.
+  static std::vector<std::string> *rocdlFilenames =
+      new std::vector<std::string>({"hc.amdgcn.bc", "opencl.amdgcn.bc",
```
River707: Let's avoid the use of static std::vectors and use something like an array of constexpr StringLiteral:
https://github.com/llvm-mirror/llvm/blob/0402afb4c1b9450302a66ed04566194db86190d9/lib/Target/AMDGPU/AMDGPUAnnotateKernelFeatures.cpp#L201

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+  llvm::SmallString<128> appendedPath;
+  llvm::sys::path::append(appendedPath, rocdlDir,
+                          GetBitcodeFilename(amdgpuVersion));
+  result.push_back(appendedPath.c_str());
+
+  return std::move(result);
+}
+
+std::unique_ptr<llvm::Module>
+LoadBitcodeModule(const std::string &filename, llvm::LLVMContext &llvmContext) {
```
River707: MLIR/LLVM style is camelCase function names.

River707: These functions should also be declared static in the global namespace, only classes/structs go in anonymous namespaces.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+  llvm::SMDiagnostic diagnostic;
+  std::unique_ptr<llvm::Module> bitcodeModule(
+      llvm::parseIRFile(llvm::StringRef(filename.data(), filename.size()),
+                        diagnostic, llvmContext));
+
+  if (bitcodeModule == nullptr) {
+    llvm::errs() << diagnostic.getFilename().str() << ":"
+                 << diagnostic.getLineNo() << ":" << diagnostic.getColumnNo()
```
River707: Can we reencode these with a proper MLIR location and emit them as diagnostics? Possibly as a note attached to a diagnostic for the current module being converted.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+
+// Links the given llvm module with the given bitcode modules.
+void LinkWithBitcodeModules(
+    llvm::Module &llvmModule,
+    const std::vector<std::string> &bitcodeModulePaths) {
+
+  llvm::Linker linker(llvmModule);
+
+  for (auto &filename : bitcodeModulePaths) {
+
```
River707: nit: Drop the newline here.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  return bitcodeModule;
+}
+
+// Links the given llvm module with the given bitcode modules.
+void LinkWithBitcodeModules(
+    llvm::Module &llvmModule,
+    const std::vector<std::string> &bitcodeModulePaths) {
+
```
River707: nit: Drop the newline here.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+} // anonymous namespace
+
+// Emits the given module to HSA Code Object. targetMachine is an initialized
+// TargetMachine for the AMDGPU target.
+OwnedHSACO
+GpuKernelToHSACOPass::emitModuleToHSACO(llvm::Module &llvmModule,
+                                        llvm::TargetMachine &targetMachine) {
+
+  static char tempdirTemplate[] = "/tmp/amdgpu_mlir-XXXXXX";
```
River707: Can you use the llvm filesystem machinery for generating a temporary file? This looks non-compliant across different platforms.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,409 @@
...
+      new llvm::raw_fd_ostream(irPath, ec, llvm::sys::fs::F_None));
+  llvmModule.print(*irFileStream, nullptr);
+  irFileStream->flush();
+
+  //// emit GCN ISA binary
+  llvm::legacy::PassManager codegenPasses;
+  llvm::SmallVector<char, 0> stream;
+  llvm::raw_svector_ostream pstream(stream);
+  std::unique_ptr<llvm::raw_fd_ostream> isabinFileStream(
+      new llvm::raw_fd_ostream(isabinPath, ec, llvm::sys::fs::F_Text));
```
River707: Is the dynamic memory allocation really necessary here?

River707:
> Hi,
> This is a huge commit. Is it possible for you to split this into multiple, more bite sized, changes? That would help the review, and also make sure that each individual component is well tested and gets the proper review it deserves.
> 
> Thanks!

Just noticed that you have nice split commits, whoops.

lib/Conversion/GPUToROCM/GenerateHSACOAccessors.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include "mlir/IR/Module.h"
+#include "mlir/IR/StandardTypes.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassRegistry.h"
+
+#include "llvm/ADT/STLExtras.h"
+
+#include <iostream>
+
+namespace mlir {
```
River707: In .cpp files, we follow the convention of 'using namespace mlir', i.e. no 'namespace mlir {' in .cpp files.

lib/Conversion/GPUToROCM/GenerateHSACOAccessors.cpp
```diff
@@ -0,0 +1,135 @@
...
+      for (auto func : module.getOps<FuncOp>()) {
+        if (StringAttr hsacoBlob =
+                func.getAttrOfType<StringAttr>(rocm::kHSACOAnnotation))
+          generate(func, hsacoBlob);
+      }
+      module.erase();
+    }
+
+    // dump the MLIR
+    if (getenv("MLIR_ROCM_DUMP_AFTER_GEN_HSACO_GETTER")) {
```
River707: Same comment as the other file.

deven-amd: this was meant for personal debug use only and not meant to make it into the upstream commit. 
will remove it (and other such instances) the next push

lib/Conversion/GPUToROCM/GenerateHSACOAccessors.cpp
deven-amd: will remove it in the next push

include/mlir/Conversion/GPUToROCM/GPUToROCMPass.h
deven-amd: will fix it in the next push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: this will be removed in the next push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: will remove them in the next push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: will remove them in the next push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: will remove them in the next push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: will remove them in the next push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: will fix in the next push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: will fix this in the next push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: have fixed this in the new push

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
deven-amd: Can we leave this as is for the time being?
This part of the code is expected to be in flux, and I need to read up on how to properly MLIR diagnostics. We can fix this in a follow up PR (there are a few other places that will require a similar change).

thanks

test/Target/rocdl.mlir
```diff
@@ -14,18 +14,18 @@ func @rocdl_special_regs() -> !llvm.i32 {
...
+  // CHECK: call i64 @__ockl_get_local_size(i32 1)
+  %8 = rocdl.workgroup.dim.y : !llvm.i64
+  // CHECK: call i64 @__ockl_get_local_size(i32 2)
+  %9 = rocdl.workgroup.dim.z : !llvm.i64
+  // CHECK: call i64 @__ockl_get_global_size(i32 0)
+  %10 = rocdl.grid.dim.x : !llvm.i64
+  // CHECK: call i64 @__ockl_get_global_size(i32 1)
+  %11 = rocdl.grid.dim.y : !llvm.i64
+  // CHECK: call i64 @__ockl_get_global_size(i32 2)
+  %12 = rocdl.grid.dim.z : !llvm.i64
```
joker-eph: This seems like the bug fix from ConvertToROCDLIR.cpp, please split it out in its own PR, it seems to me to be independent from the new features

joker-eph:
> This is a huge commit. Is it possible for you to split this into multiple, more bite sized, changes? That would help the review, and also make sure that each individual component is well tested and gets the proper review it deserves.

+1: it seems to me that adding the runner is independent from adding the lowering passes.

sherhut:
I was not sure whether this is ready for review or whether I should wait for the push. Is this ready now and should I review?

deven-amd:
@joker-eph 

I will break this PR into one for each commit, as per your request.

The PR for the first commit (converting return type from i32 to i64) can stand on its own. (Filed PR #171)

The PRs for the other 4 commits (in this PR) will need to be filed sequentially since the later commits assume/require the presence of the former commits.

I will file the PR for the first of those four commits (ConvertKernelFuncToHSACO) after I have addressed all your code review feedback.

Closing out this PR

include/mlir/Conversion/GPUToROCM/GPUToROCMPass.h
```diff
@@ -0,0 +1,102 @@
...
+  /// the directory containing the ROCDL bitcode libraries
+  std::string rocdlDir;
+
+  /// the path the ld.lld linker to use when generating the HSACO
+  std::string linkerPath;
+};
+
+} // namespace rocm
+
+using OwnedHSACO = std::unique_ptr<std::vector<char>>;
```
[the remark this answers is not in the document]
deven-amd: Two reasons for this.

1. The binary blob owned by this variable could potentially be large, and hence pass-by-pointer instead of pass-by-value (Note that the "HSACO" string is only returned in the "test" scenario)

2. Consistency with the CUDA side (where we OwnedCubin, which is also unique_ptr wrapper on a vector of chars)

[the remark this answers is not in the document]
joker-eph: > The binary blob owned by this variable could potentially be large, and hence pass-by-pointer instead of pass-by-value

A vector is very small (on the scale of 3 pointers / 24 bytes), you don't have to copy the heap allocated buffer.

> Consistency with the CUDA side (where we OwnedCubin, which is also unique_ptr wrapper on a vector of chars)

 That's a better reason, I should probably look into this.

joker-eph:
> The PRs for the other 4 commits (in this PR) will need to be filed sequentially since the later commits assume/require the presence of the former commits.

If you have the 4 commits, then you can file the 4 PRs at once, assuming you push the first commit to a branch, the 2nd to another branch (rebased on the first one of course), etc.
The diff for the second PR will include the first one, but that'll clear itself when the first one merges.

include/mlir/Conversion/GPUToROCM/GPUToROCMPass.h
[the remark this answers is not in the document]
deven-amd: doesn't the  `std::vector<T>` own the memory used by its contents, and hence copying that vector could (sans compiler optimizations) also result in the need to allocate memory for its contents?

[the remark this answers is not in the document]
joker-eph: If you pass it by value, you need to std::move it (same as the unique_ptr) to avoid the copy.
Just returning it by value already implies the move.

PR #171, opened 2019-10-08
[ROCm] Changing the return type for the device function calls from i32 to i64

deven-amd:
Also making corresponding changes in the unit test for the same

-----------------------

@joker-eph @whchung


joker-eph:
Thanks!

joker-eph:
(nit: your title is accurately describing the "what". In general, I like to have a contextual description in the commit message to also explain the "why": https://chris.beams.io/posts/git-commit/ ; but don't bother here I've already pulled the changes in)

PR #172, opened 2019-10-08
Use correct spelling in functions for quantization

kiszk:
This PR renames these two functions using correct spelling
```
getDefaultMaxinumForInteger() -> getDefaultMaximumForInteger()
getDefaultMininumForInteger() -> getDefaultMinimumForInteger()
```

These functions are called from [another repository](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/lite/transforms/load_quantization_recipe.cc#L78-L85). After merging this PR, I will create a PR at [another repository](https://github.com/tensorflow/tensorflow).


joker-eph:
We will take care of updating the uses in TensorFlow when pulling this inside google.

PR #175, opened 2019-10-09
minor spelling tweaks under bindings, examples, g3doc,  test, unittests, and utils

kiszk:
This PR addresses spelling tweaks under bindings, examples, g3doc,  test, unittests, and utils directories.   
Another PR will address spelling tweaks in other directories.

PR #177, opened 2019-10-09
minor spelling tweaks under lib

kiszk:
This PR addresses spelling tweaks under lib directory.   
This PR belongs to the same group as #175.

### DIFF
--- a/lib/Dialect/SPIRV/Serialization/Deserializer.cpp
+++ b/lib/Dialect/SPIRV/Serialization/Deserializer.cpp
@@ -15,7 +15,7 @@
 // limitations under the License.
 // =============================================================================
 //
-// This file defines the SPIR-V binary to MLIR SPIR-V module deseralization.
+// This file defines the SPIR-V binary to MLIR SPIR-V module deserialization.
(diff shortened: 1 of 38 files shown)


lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -564,7 +564,7 @@ LogicalResult Deserializer::processExtInstImport(ArrayRef<uint32_t> words) {
   if (words.size() < 2) {
     return emitError(unknownLoc,
                      "OpExtInstImport must have a result <id> and a literal "
-                     "string for the extensed instruction set name");
+                     "string for the extented instruction set name");
```
joker-eph: `extended` seems the right word here?

kiszk: Oh, you are right. I will address it in another PR.

kiszk: Ah, thank you for fixing it.

PR #181, opened 2019-10-10
[ROCm] Adding pass to generate the HSACO binary blob from the GPU kernel function

deven-amd:
This is the first of four PRs to add support for the ROCm backend for AMD GPUs. This PR adds the pass to generate the HSA Code objects corresponding the GPU kernels. This PR builds on the work done by @whchung in PR #63

This PR is a follow-up to request in PR #169. It addresses all but two of the code-review comments made in that PR for the `ConvertKernelFuncToHSACO.cpp` file.

1. https://github.com/tensorflow/mlir/pull/169#discussion_r332618425

> Why are you wrapping the vector in a unique_ptr instead of just using vector by value?
> 
> There are a bunch of places where the code could be simplified, for instance:
> 
> ```
>    const char data[] = "HSACO";
>    return std::make_unique<std::vector<char>>(data, data + sizeof(data) - 1);
> ```
> 
> would likely become:
> 
> ```
>   return "HSACO";
> ```
Two reasons for this.

  * The binary blob owned by this variable could potentially be large, and hence pass-by-pointer instead of pass-by-value (Note that the "HSACO" string is only returned in the "test" scenario)

  * Consistency with the CUDA side (where we OwnedCubin, which is also unique_ptr wrapper on a vector of chars)

2. https://github.com/tensorflow/mlir/pull/169#discussion_r332617848

> Do we have to go through files? Isn't `lld` usable as a library?

At least for the time being, it is easier to go through files. Dumping out the .ll file is helpful for debugging purposes, so that is something we would like to keep regardless. Also I am still trying to figure out how to invoke `lld` as an API call, within the mlir setup :(

----------------------------------------------------------------------

@joker-eph @whchung


lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,401 @@
...
+    return {};
+
+  // Lower LLVM module to HSA code object
+  return emitModuleToHSACO(llvmModule, function, *targetMachine);
+}
+
+LogicalResult GpuKernelToHSACOPass::translateGpuKernelToHSACOAnnotation(
+    llvm::Module &llvmModule, FuncOp function) {
+
+  auto hsaco = convertModuleToHSACO(llvmModule, function);
```
joker-eph: Nit: don't use `auto` when the type isn't obvious from the context.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,401 @@
...
+                                        FuncOp function,
+                                        llvm::TargetMachine &targetMachine) {
+  llvm::SmallString<128> tempdirName;
+  if (llvm::sys::fs::createUniqueDirectory("/tmp/amdgpu_mlir", tempdirName)) {
+    llvm::errs() << "Failed to create tempdir for generating HSACO\n";
+    return {};
+  }
+
+  // prepare filenames for all stages of compilation:
+  // IR, ISA, binary ISA, and HSACO
```
joker-eph: Please add a TODO and file a bug to replace this with using lld as a library for the link.

include/mlir/Conversion/GPUToROCM/GPUToROCMPass.h
```diff
@@ -0,0 +1,89 @@
...
+enum class HSACOVersion { V3 };
+
+/// Configurable parameters for generating the HSACO blobs from GPU Kernels
+struct HSACOGeneratorConfig {
+
+  /// Constructor - sets the default values for the configurable parameters
+  HSACOGeneratorConfig(bool isTestMode)
+      : testMode(isTestMode), amdgpuVersion(AMDGPUVersion::GFX900),
+        hsacoVersion(HSACOVersion::V3), rocdlDir("/opt/rocm/lib/"),
+        linkerPath("/opt/rocm/hcc/bin/ld.lld") {}
```
joker-eph: I suspect default values like this one should be generated by the cmake build script

CMakeLists.txt
```diff
@@ -44,6 +44,13 @@ endif()
 
 set(MLIR_CUDA_RUNNER_ENABLED 0 CACHE BOOL "Enable building the mlir CUDA runner")
 
+# Build the ROCM conversions if the AMDGPU backend is available
+if ("AMDGPU" IN_LIST LLVM_TARGETS_TO_BUILD)
+  set(MLIR_ROCM_CONVERSIONS_ENABLED 1)
+else()
+  set(MLIR_ROCM_CONVERSIONS_ENABLED 0)
+endif()
```
joker-eph: This isn't correct I believe: at the moment you have a dependency on the environment (lld, ROCm runtime bitcode libraries, etc.)

As mentioned elsewhere: there should be a step detecting the environment (and we should probably put this behind an opt-in CMake flag by the way) and generating a rocm_config.h that can be included in the code.

deven-amd: I have address this in the CMakeLists.txt file in the GPUToROCM dir

deven-amd (commented):
@joker-eph , I have pushed out a new rebased commit that addresses the feedback you provided. 

Please re-review....thanks

deven

deven-amd:
@joker-eph , gentle ping

joker-eph:
> @joker-eph , gentle ping

Sorry, I missed the update (and LLVM Dev Meeting happened). I'm catching up on everything and will get to it this week. Ping on Monday if you don't see an update!

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,407 @@
...
+  llvm::legacy::PassManager codegenPasses;
+  llvm::SmallVector<char, 0> stream;
+  llvm::raw_svector_ostream pstream(stream);
+  llvm::raw_fd_ostream isabinFileStream(isabinPath, ec, llvm::sys::fs::F_Text);
+  llvmModule.setDataLayout(targetMachine.createDataLayout());
+  targetMachine.addPassesToEmitFile(codegenPasses, isabinFileStream, nullptr,
+                                    llvm::TargetMachine::CGFT_ObjectFile);
+  codegenPasses.run(llvmModule);
+  isabinFileStream.flush();
+
```
joker-eph: I would try to keep maximum coverage for the test, since the pass is only available when the AMDGPU backend is built in, we should be able to get until there for the basic test. We could just reload the binary file `isabinFilename` print it as hexadecimal and return.

Something like:

```
if (testMode) {
  auto hsacoFileOrError = llvm::MemoryBuffer::getFileAsStream(isabinFilename);
  if ((ec = hsacoFileOrError.getError()))
    return {};
  std::unique_ptr<llvm::MemoryBuffer> hsacoFile = std::move(hsacoFileOrError.get());
  SmallVector<char, 1> buffer;
  {
    llvm::raw_svector_ostream os(buffer);
    for (auto char : hascoFile.getBuffer())
      os.write_hex(char);
  }
  return std::make_unique<std::vector<char>>(buffer.begin(), buffer.end());
}
```


The test would be able to check that we actually correctly generate an ELF binary (by checking the ELF magic in hexadecimal) and cover everything but the linker when ROCM is not available.

lib/Conversion/GPUToROCM/ConvertKernelFuncToHSACO.cpp
```diff
@@ -0,0 +1,407 @@
...
+}
+
+std::unique_ptr<OpPassBase<ModuleOp>> mlir::createConvertGPUKernelToHSACOPass(
+    rocm::HSACOGeneratorConfig hsacoGeneratorConfig) {
+  return std::make_unique<GpuKernelToHSACOPass>(hsacoGeneratorConfig);
+}
+
+static PassRegistration<GpuKernelToHSACOPass>
+    pass("test-kernel-to-hsaco",
+         "Convert all kernel functions to HSA code object blobs");
```
joker-eph: With a macro that define if ROCM is installed or not, we can then conditionally register another version of the pass "rocm-hsaco-lowering" or something like which actually run the linker as well, and add a test for this.

PR #184, opened 2019-10-11
add LLVM IR dialect hooks for FP128 and X86_FP80 types

schweitzpgi:
Add the quad precision FP128 and the X86 extended floating point types to the set of LLVM IR dialect floating-point type builders.

### DIFF
--- a/lib/Dialect/LLVMIR/IR/LLVMDialect.cpp
+++ b/lib/Dialect/LLVMIR/IR/LLVMDialect.cpp
@@ -1219,7 +1219,8 @@ LLVMDialect::LLVMDialect(MLIRContext *context)
   impl->floatTy = LLVMType::get(context, llvm::Type::getFloatTy(llvmContext));
   impl->halfTy = LLVMType::get(context, llvm::Type::getHalfTy(llvmContext));
   impl->fp128Ty = LLVMType::get(context, llvm::Type::getFP128Ty(llvmContext));
-  impl->x86_fp80Ty = LLVMType::get(context, llvm::Type::getX86_FP80Ty(llvmContext));
+  impl->x86_fp80Ty = LLVMType::get(context,
+                                   llvm::Type::getX86_FP80Ty(llvmContext));
   /// Other Types.
   impl->voidTy = LLVMType::get(context, llvm::Type::getVoidTy(llvmContext));
 }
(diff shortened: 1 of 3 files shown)


lib/Dialect/LLVMIR/IR/LLVMDialect.cpp
```diff
@@ -1367,6 +1367,16 @@ LLVMType LLVMType::getFloatTy(LLVMDialect *dialect) {
 LLVMType LLVMType::getHalfTy(LLVMDialect *dialect) {
   return dialect->impl->halfTy;
 }
+LLVMType LLVMType::getQuadTy(LLVMDialect *dialect) {
+  return getLocked(dialect, [=] {
```
River707: These types are thread-safe to query from the LLVMContext, so we can cache them as we do the others.

River707: Let's also use the same method name as the LLVM::get methods if possible.

schweitzpgi: OK, I wasn't sure if it was better to cache them or not.

River707: I added caching for most of the ones that makes sense(mostly mirroring what happens within LLVMContextImpl internally). We mainly try to avoid calling getLocked here if we can given that it will lock two mutexs: 1 for the LLVM module, and 1 internally in the MLIR context.

schweitzpgi: Thank you for the explanation.

lib/Dialect/LLVMIR/IR/LLVMDialect.cpp
```diff
@@ -1218,6 +1218,8 @@ LLVMDialect::LLVMDialect(MLIRContext *context)
   impl->doubleTy = LLVMType::get(context, llvm::Type::getDoubleTy(llvmContext));
   impl->floatTy = LLVMType::get(context, llvm::Type::getFloatTy(llvmContext));
   impl->halfTy = LLVMType::get(context, llvm::Type::getHalfTy(llvmContext));
+  impl->fp128Ty = LLVMType::get(context, llvm::Type::getFP128Ty(llvmContext));
+  impl->x86_fp80Ty = LLVMType::get(context, llvm::Type::getX86_FP80Ty(llvmContext));
```
River707: Can you wrap this around 80 characters? It is failing clang format.

schweitzpgi: Oops. Fixed.

PR #187, opened 2019-10-15
Update Chapter 3 to demonstrate pattern match and rewrite optimizations

joker-eph:
This is using Table-driven Declarative Rewrite Rules (DRR), the previous
version of the tutorial only showed the C++ patterns.

### DIFF
--- a/examples/toy/Ch3/CMakeLists.txt
+++ b/examples/toy/Ch3/CMakeLists.txt
@@ -2,16 +2,35 @@ set(LLVM_LINK_COMPONENTS
   Support
   )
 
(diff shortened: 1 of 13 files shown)


joker-eph:
@sana-damani  there is a warning:

```
projects/mlir/examples/toy/Ch3/ToyCombine.inc:128:6: warning: unused function 'populateWithGenerated' [-Wunused-function]
```

River707:
The warning is because we always generate the `populateWithGenerated` function, but in this case we don't want to use it. We want to register the canonicalization patterns directly with the op.

PR #193, opened 2019-10-17
Use a SmallVector instead of an ArrayRef to materialize a temporary local array

joker-eph:
This pattern is error prone and unfortunately none of the sanitizer is catching
it at the moment.

### DIFF
--- a/lib/Dialect/GPU/IR/GPUDialect.cpp
+++ b/lib/Dialect/GPU/IR/GPUDialect.cpp
@@ -142,7 +142,7 @@ static LogicalResult verifyAllReduce(gpu::AllReduce allReduce) {
     return allReduce.emitError(
         "expected either an op attribute or a non-empty body");
   if (allReduce.op()) {
-    ArrayRef<StringRef> supportedOps{"add", "mul"};
+    SmallVector<StringRef, 2> supportedOps{"add", "mul"};
     if (!llvm::is_contained(supportedOps, *allReduce.op()))
       return allReduce.emitError("op \"") << *allReduce.op() << "\" is invalid";
   }


bondhugula:
> This pattern is error prone and unfortunately none of the sanitizer is catching
> it at the moment.

Building with GCC does yield a warning for such a pattern. For eg. I did get:
```
/home/uday/llvm-project/llvm/projects/mlir/lib/Dialect/GPU/IR/GPUDialect.cpp:145:50:   required from here
/home/uday/llvm-project/llvm/include/llvm/ADT/ArrayRef.h:101:37: warning: initializing ‘llvm::ArrayRef<llvm::StringRef>::Data’ from ‘std::initializer_list<llvm::StringRef>::begin’ does not extend the lifetime of the underlying array [-Winit-list-lifetime]
  101 |     : Data(Vec.begin() == Vec.end() ? (T*)nullptr : Vec.begin()),
/home/uday/llvm-project/llvm/include/llvm/ADT/ArrayRef.h: In instantiation of ‘llvm::ArrayRef<T>::ArrayRef(const std::initializer_list<_Tp>&) [with T = mlir::Type]’:
```
Building with GCC in general could expose all other such uses. 

In the past, when I used a pattern like this once, it did lead to a failure in the automatic checks on MLIR's github, and @ftynse fixed it. So, either something has changed in your tests, or this is something that's not easily reproducible.

ftynse:
It may be tricky to reproduce, like many stack-related UBs. ArrayRef will end up storing a pointer to a temporary, but there's no guarantee something will reuse that address before we read from the ArrayRef (which would lead to a segfault).

I don't know enough of the sanitizer structure to understand why it doesn't warn about such cases. However, creating a stack ArrayRef object from an initializer list is something we can add as a diagnostic to clang-tidy.

joker-eph:
We are building with GCC also, I suspect were don't have a Release-no-assert config though, maybe adding would have caught this issue? @jpienaar ?

Issue #165, opened 2019-10-03
[spirv] Add canonicalization patterns for ops


antiagainst:
We don't have much canonicalization patterns for SPIR-V ops thus far. It would be nice to gradually build them up. Interesting ones include but not limited to:

* Combine chained `spv.AccessChain`s into one `spv.AccessChain`
* Convert `spv.selection` to `spv.Select` if possible
* Constant fold arithmetic / logical / cast / etc. ops
* Other peephole optimizations like converting chained `spv.Bitcast` ops into one, turning `spv.LogicalNot (spv.IEqual ...)` into `spv.INotEqual`, etc.
* ...

antiagainst:
Ping @denis0x0D to see if you are potentially interested in helping. :)

denis0x0D:
@antiagainst it would be great for me to take some of these tasks, thanks a lot!

denis0x0D:
@antiagainst @MaheshRavishankar is it ok, if I take 
 

- Convert spv.selection to spv.Select if possible

antiagainst:
Sure, please feel free to take it! Thanks for the help as always, Denis! :)

denis0x0D:
@antiagainst @MaheshRavishankar can you please specify regarding to "convert spv.selection to spv.Select", how actually deep analysis should be? 

For example:

```
func @selection(%cond: i1) -> () {
...
  %zero = spv.constant 0: i32
  %one = spv.constant 1: i32
  %two = spv.constant 2: i32
  %x = spv.Variable init(%zero) : !spv.ptr<i32, Function>
  spv.selection {
   spv.BranchConditional %cond, ^then, ^else
  ^then:
    spv.Store "Function" %x, %one : i32
    spv.Branch ^merge
  ^else:
    spv.Store "Function" %x, %two : i32
    spv.Branch ^merge
  ^merge:
    spv._merge
  }
 ...
}
```
this code could be simplified as follows:

```
func @selection(%arg0: i1) -> () {
...
    %0 = spv.constant 0 : i32
    %1 = spv.constant 1 : i32
    %2 = spv.constant 2 : i32
    %3 = spv.Variable init(%0) : !spv.ptr<i32, Function>
    %4 = spv.Select %arg0, %1, %2 : i1, i32
    spv.Store "Function" %3, %4 : i32
  ...
}
```
this pattern is simple to detect and therefore to simplify the code, but there could situations that require more deep analysis for example
[...]

PR #198, opened 2019-10-18
[spirv] AccessChain canonicalization.

denis0x0D:
This patch is addressing  https://github.com/tensorflow/mlir/issues/165 

> Combine chained `spirv::AccessChainOp` operations into one
> `spirv::AccessChainOp` operation.

@antiagainst @MaheshRavishankar can you please take a look? Thanks!

### DIFF
--- a/lib/Dialect/SPIRV/SPIRVOps.cpp
+++ b/lib/Dialect/SPIRV/SPIRVOps.cpp
@@ -544,6 +544,41 @@ static LogicalResult verify(spirv::AccessChainOp accessChainOp) {
   return success();
 }
 
+namespace {
+
+// Combine chained `spirv::AccessChainOp` operations into one
+// `spirv::AccessChainOp` operation.
+struct CombineChainedAccessChain
+    : public OpRewritePattern<spirv::AccessChainOp> {
+  using OpRewritePattern<spirv::AccessChainOp>::OpRewritePattern;
+
+  PatternMatchResult matchAndRewrite(spirv::AccessChainOp accessChainOp,
+                                     PatternRewriter &rewriter) const override {
+    auto parentAccessChainOp = dyn_cast_or_null<spirv::AccessChainOp>(
+        accessChainOp.base_ptr()->getDefiningOp());
+
+    if (!parentAccessChainOp) {
+      return matchFailure();
+    }
+
+    // Combine indices.
+    SmallVector<Value *, 4> indices(parentAccessChainOp.indices());
+    indices.append(accessChainOp.indices().begin(),
+                   accessChainOp.indices().end());
+
+    rewriter.replaceOpWithNewOp<spirv::AccessChainOp>(
+        accessChainOp, parentAccessChainOp.base_ptr(), indices);
+
+    return matchSuccess();
+  }
+};
+} // namespace
+
+void spirv::AccessChainOp::getCanonicalizationPatterns(
+    OwningRewritePatternList &results, MLIRContext *context) {
+  results.insert<CombineChainedAccessChain>(context);
+}
+
 //===----------------------------------------------------------------------===//
 // spv._address_of
 //===----------------------------------------------------------------------===//
@@ -1976,7 +2011,8 @@ namespace {
 //                       | merge block |
 //                       +-------------+
 //
-struct SelectionOpCanonicalizer : public OpRewritePattern<spirv::SelectionOp> {
+struct ConvertSelectionOpToSelect
+    : public OpRewritePattern<spirv::SelectionOp> {
   using OpRewritePattern<spirv::SelectionOp>::OpRewritePattern;
 
   PatternMatchResult matchAndRewrite(spirv::SelectionOp selectionOp,
@@ -2023,7 +2059,7 @@ struct SelectionOpCanonicalizer : public OpRewritePattern<spirv::SelectionOp> {
                                     selectOp.getResult(), storeOpAttributes);
 
     // `spv.selection` is not needed anymore.
-    rewriter.replaceOp(op, llvm::None);
+    rewriter.eraseOp(op);
     return matchSuccess();
   }
 
@@ -2071,7 +2107,7 @@ private:
   }
 };
 
-PatternMatchResult SelectionOpCanonicalizer::canCanonicalizeSelection(
+PatternMatchResult ConvertSelectionOpToSelect::canCanonicalizeSelection(
     Block *trueBlock, Block *falseBlock, Block *mergeBlock) const {
   // Each block must consists of 2 operations.
   if ((std::distance(trueBlock->begin(), trueBlock->end()) != 2) ||
@@ -2110,7 +2146,7 @@ PatternMatchResult SelectionOpCanonicalizer::canCanonicalizeSelection(
 
 void spirv::SelectionOp::getCanonicalizationPatterns(
     OwningRewritePatternList &results, MLIRContext *context) {
-  results.insert<SelectionOpCanonicalizer>(context);
+  results.insert<ConvertSelectionOpToSelect>(context);
 }
 
 //===----------------------------------------------------------------------===//
--- a/test/Dialect/SPIRV/canonicalize.mlir
+++ b/test/Dialect/SPIRV/canonicalize.mlir
@@ -1,6 +1,64 @@
 // RUN: mlir-opt %s -split-input-file -canonicalize | FileCheck %s
 
 //===----------------------------------------------------------------------===//
+// spv.AccsessChain
+//===----------------------------------------------------------------------===//
+
+func @combine_full_access_chain() -> f32 {
+  // CHECK: %[[INDEX:.*]] = spv.constant 0
+  // CHECK-NEXT: %[[VAR:.*]] = spv.Variable
+  // CHECK-NEXT: %[[PTR:.*]] = spv.AccessChain %[[VAR]][%[[INDEX]], %[[INDEX]], %[[INDEX]]]
+  // CHECK-NEXT: spv.Load "Function" %[[PTR]]
+  %c0 = spv.constant 0: i32
+  %0 = spv.Variable : !spv.ptr<!spv.struct<!spv.array<4x!spv.array<4xf32>>, !spv.array<4xi32>>, Function>
+  %1 = spv.AccessChain %0[%c0] : !spv.ptr<!spv.struct<!spv.array<4x!spv.array<4xf32>>, !spv.array<4xi32>>, Function>
+  %2 = spv.AccessChain %1[%c0, %c0] : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
+  %3 = spv.Load "Function" %2 : f32
+  spv.ReturnValue %3 : f32
+}
+
+// -----
+
+func @combine_access_chain_multi_use() -> !spv.array<4xf32> {
+  // CHECK: %[[INDEX:.*]] = spv.constant 0
+  // CHECK-NEXT: %[[VAR:.*]] = spv.Variable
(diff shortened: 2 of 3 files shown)


lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -544,6 +544,62 @@ static LogicalResult verify(spirv::AccessChainOp accessChainOp) {
...
+        !canCanonicalizeChainedAccesses(accessChainOp, nextAccessChainOp)) {
+      return matchFailure();
+    }
+    // Combine indices.
+    llvm::SmallVector<Value *, 4> indices(accessChainOp.indices());
+    indices.insert(indices.end(), nextAccessChainOp.indices().begin(),
+                   nextAccessChainOp.indices().end());
+
+    // Replace `nextAccessChainOp` with newly created `spirv::AccessChainOp`.
+    rewriter.replaceOpWithNewOp<spirv::AccessChainOp>(
```
antiagainst: Hmm, normally an op's canonicalization should only change the op itself. So instead of looking at and replacing the **next** access chain op, what about just checking whether the current access chain op's base pointer to see whether it comes from an access chain op, if so, replace the **current** access chain op by using the previous access chain op's base pointer and all the indices and leave the previous access chain op untouched? That is, turning

```
%0 = spv.AccessChain %ptr, %idx0, %idx1
%1 = spv.AccessChain %0, %idx2, %idx3
```
to
```
%0 = spv.AccessChain %ptr, %idx0, %idx1
%1 = spv.AccessChain %ptr, %idx0, %idx1, idx2, %idx3
```

The previous access chain op may have multiple users and that's fine. If it has no additional users then the framework will handle the removal of it directly.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -544,6 +544,62 @@ static LogicalResult verify(spirv::AccessChainOp accessChainOp) {
   return success();
 }
 
+namespace {
+
+// Combine chained `spirv::AccessChainOp` operations into one
+// `spirv::AccessChainOp` operation. The current approach is based on analyzing
+// only two operations at once: the current `spirv::AccessChainOp` and an
+// operation immediately following the current operation.
+struct AccessChainOpCanonicalizer
```
antiagainst: Can we call this as `CombineChainedAccessChain` to be clear? The existing name is too general and indicates we have all canonicalization patterns inside. Rather, we'd like to create another class for another canonicalization transformation.

Also related, can we rename `SelectionOpCanonicalizer` to `ConvertSelectionToSelect`?

test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -1,5 +1,111 @@
 // RUN: mlir-opt %s -split-input-file -canonicalize | FileCheck %s
 
+//===----------------------------------------------------------------------===//
```
antiagainst: Nice tests! :)

test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -1,5 +1,111 @@
...
+// spv.AccsessChain
+//===----------------------------------------------------------------------===//
+
+func @access_chain_canonicalization_full_0() -> f32 {
+  // CHECK: %[[INDEX_0:.*]] = spv.constant 0 : i32
+  // CHECK-NEXT: %[[COMPOSITE:.*]] = spv.Variable : !spv.ptr<!spv.struct<!spv.array<4 x !spv.array<4 x f32>>, !spv.array<4 x i32>>, Function>
```
antiagainst: We can simplify the `CHECK`s here. No need to check all the details. What about something like

```
// CHECK: %[[INDEX:.*]] = spv.constant 0
// CHECK: %[[VAR:.*]] = spv.Variable
// CHECK: %[[PTR:.*]] = spv.AccessChain %[[VAR]][%[[INDEX]], %[[INDEX]], %[[INDEX]]]
// CHECK: spv.Load "Function" %[[PTR]]
```
The type consistency is guaranteed by each op's validation itself and we should be fine to rely on that.

Similarly for all the test cases.

test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -1,5 +1,111 @@
 // RUN: mlir-opt %s -split-input-file -canonicalize | FileCheck %s
 
+//===----------------------------------------------------------------------===//
+// spv.AccsessChain
+//===----------------------------------------------------------------------===//
+
+func @access_chain_canonicalization_full_0() -> f32 {
```
antiagainst: Nit: `combine_full_access_chain`? A good name can serve as documentation. :)

test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -1,5 +1,111 @@
...
+}
+
+// -----
+
+// The result of the second `spv.AccessChain` operation has two uses and therefore cannot be combined with the third.
+func @access_chain_canonicalization_partial() -> !spv.array<4xf32> {
+  // CHECK: %[[INDEX_0:.*]] = spv.constant 0 : i32
+  // CHECK-NEXT: %[[COMPOSITE:.*]] = spv.Variable : !spv.ptr<!spv.struct<!spv.array<4 x !spv.array<4 x f32>>, !spv.array<4 x i32>>, Function>
+  // CHECK-NEXT: %[[COMPONENT_0:.*]] = spv.AccessChain %[[COMPOSITE]][%[[INDEX_0]], %[[INDEX_0]]] : !spv.ptr<!spv.struct<!spv.array<4 x !spv.array<4 x f32>>, !spv.array<4 x i32>>, Function>
+  // CHECK-NEXT: %[[COMPONENT_1:.*]] = spv.AccessChain %[[COMPONENT_0]][%[[INDEX_0]]] : !spv.ptr<!spv.array<4 x f32>, Function>
```
antiagainst: We should also be able to turn this into having two indices if looking at the previous access chain op instead of the next one.

test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -1,5 +1,111 @@
...
+  %1 = spv.AccessChain %0[%c0] : !spv.ptr<!spv.struct<!spv.array<4x!spv.array<4xf32>>, !spv.array<4xi32>>, Function>
+  %2 = spv.AccessChain %1[%c0] : !spv.ptr<!spv.array<4x!spv.array<4xf32>>, Function>
+  %3 = spv.AccessChain %2[%c0] : !spv.ptr<!spv.array<4xf32>, Function>
+  %5 = spv.Load "Function" %3 : f32
+  spv.ReturnValue %5 : f32
+}
+
+// -----
+
+func @access_chain_canonicalization_full_1() -> i32 {
```
antiagainst: I think we can merge this one into the previous by turning one of the access chain op to contain more than one indices.

test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -1,5 +1,111 @@
...
+  %1 = spv.AccessChain %0[%c0, %c1] : !spv.ptr<!spv.struct<!spv.array<4x!spv.struct<!spv.array<4xi32>>>>, Function>
+  %2 = spv.AccessChain %1[%c0, %c1] : !spv.ptr<!spv.struct<!spv.array<4xi32>>, Function>
+  %3 = spv.Load "Function" %2 : i32
+  spv.ReturnValue %3 : i32
+}
+
+// -----
+
+// The result of the second `spv.AccessChain` operation has two uses and therefore cannot be combined with the third.
+func @access_chain_canonicalization_partial() -> !spv.array<4xf32> {
```
antiagainst: combine_access_chain_multi_use?

test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -1,5 +1,111 @@
...
+  %3 = spv.AccessChain %2[%c0] : !spv.ptr<!spv.array<4xf32>, Function>
+  %4 = spv.Load "Function" %2 : !spv.array<4xf32>
+  %5 = spv.Load "Function" %3 : f32
+  spv.ReturnValue %4: !spv.array<4xf32>
+}
+
+// -----
+
+// Each result of the first two `spv.AccessChain` operations has more than one use.
+func @cannot_canonicalize_access_chain_0() -> f32 {
```
antiagainst: Similarly this will be similar to the previous one and can be removed if looking at the previous access chain op instead of the next one.

test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -1,5 +1,111 @@
...
+  %4 = spv.Load "Function" %1 : !spv.array<4x!spv.array<4xf32>>
+  %5 = spv.Load "Function" %2 : !spv.array<4xf32>
+  %6 = spv.Load "Function" %3 : f32
+  spv.ReturnValue %6: f32
+}
+
+// -----
+
+// Not a chained accesses.
+func @cannot_canonicalize_access_chain_1() -> !spv.array<4xi32> {
```
antiagainst: dont_combine_access_chain_without_common_base ?

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: @antiagainst thanks for point me on this!

denis0x0D:
@antiagainst thanks a lot for review! Updated.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -544,6 +544,41 @@ static LogicalResult verify(spirv::AccessChainOp accessChainOp) {
...
+    auto parentAccessChainOp = dyn_cast<spirv::AccessChainOp>(
+        accessChainOp.base_ptr()->getDefiningOp());
+
+    if (!parentAccessChainOp) {
+      return matchFailure();
+    }
+
+    // Combine indices.
+    llvm::SmallVector<Value *, 4> indices(parentAccessChainOp.indices());
+    indices.insert(indices.end(), accessChainOp.indices().begin(),
```
River707: Drop the llvm:: from the SmallVector. The insert can also be changed to append.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -544,6 +544,41 @@ static LogicalResult verify(spirv::AccessChainOp accessChainOp) {
...
+
+// Combine chained `spirv::AccessChainOp` operations into one
+// `spirv::AccessChainOp` operation.
+struct CombineChainedAccessChain
+    : public OpRewritePattern<spirv::AccessChainOp> {
+  using OpRewritePattern<spirv::AccessChainOp>::OpRewritePattern;
+
+  PatternMatchResult matchAndRewrite(spirv::AccessChainOp accessChainOp,
+                                     PatternRewriter &rewriter) const override {
+    auto parentAccessChainOp = dyn_cast<spirv::AccessChainOp>(
```
antiagainst: Use `dyn_cast_or_null`here. Base pointer can come from a block arugment where `getDefiningOp` will return nullptr.

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: thanks, updated!

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: thanks, updated

antiagainst:
@denis0x0D, could you squash the commits into one?

denis0x0D:
@antiagainst thanks a lot for review! Updated.

PR #202, opened 2019-10-21
More affine expr simplifications for floordiv and mod

bondhugula:
Add one more simplification for floordiv and mod affine expressions.
Examples:
 (2*d0 + 1) floordiv 2 is simplified to d0
 (8*d0 + 4*d1 + d2) floordiv 4 simplified to 4*d0 + d1 + d2 floordiv 4.
 etc.

 Similarly, (4*d1 + 1) mod 2 is simplified to 1,
            (2*d0 + 8*d1) mod 8 simplified to 2*d0 mod 8.

Change getLargestKnownDivisor to return int64_t to be consistent and
to avoid casting at call sites (since the result is mixed in expressions
of int64_t/index type).

### DIFF
--- a/include/mlir/IR/AffineExpr.h
+++ b/include/mlir/IR/AffineExpr.h
@@ -114,8 +114,9 @@ public:
   /// floordiv, ceildiv, and mod is only allowed w.r.t constants.
   bool isPureAffine() const;
 
-  /// Returns the greatest known integral divisor of this affine expression.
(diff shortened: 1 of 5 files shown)


bondhugula:
Resolved conflicts here; please let me know if there's anything else.

bondhugula:
Looks like this was completely missed after being approved and has been pending for about a month!

River707:
Looks like no-one added the pull tag. Sorry about that.

Issue #197, opened 2019-10-18
Operation interfaces: can't name the interface method same as free function on op


bondhugula:
With the way tablegen based operation interfaces are generated
https://github.com/tensorflow/mlir/blob/master/g3doc/OpDefinitions.md#operation-interfaces
it doesn't appear to be possible to use the same name for an interface method and a free function that it calls taking the concrete op as input. The latter would resolve to an internal op interface generated method from implicit conversion of the concrete op to Operation *, and thus lead to an infinite recursion. Normally, one would like to use the same name.  For example, consider:
```
InterfaceMethod<"/*insert doc here*/",
      "unsigned", "computeSomeProperty", (ins), [{
        return computeSomeProperty(op);
    }]>,
```
computeSomeProperty(op) will resolve to an internally generated method:
```
unsigned computeSomeProperty(Operation *tablegen_opaque_op) final 
```
instead of the intended mlir::computeSomeProperty(ConcreteOpType opType);

antiagainst:
Isn't this a C++ symbol resolution issue? Does using fully qualified symbol name help?

bondhugula:
> Isn't this a C++ symbol resolution issue? Does using fully qualified symbol name help?

The different ops that use the op interface typically reside in different dialects and sometimes in different namespaces (for eg. mlir::loop::ForOp and mlir::AffineForOp). There isn't a single fully qualified symbol name.

River707:
This is c++ visibility rules, it has nothing to do with interfaces. What you are encountering is essentially:

unsigned computeSomeProperty(Operation *opaque_op) {
  auto op = cast<ConcreteOp>(opaque_op);
  return computeSomeProperty(op);
}

You will have to qualify the body as you in c++. Why can't you just use `mlir::computeSomeProperty` if that is what you want? The place that the call is resolved is where ever the interface header is included(just like any other template).

bondhugula:
> This is c++ visibility rules, it has nothing to do with interfaces. What you are encountering is essentially:
> 
> unsigned computeSomeProperty(Operation *opaque_op) {
> auto op = cast(opaque_op);
> return computeSomeProperty(op);
> }
> 
> You will have to qualify the body as you in c++. Why can't you just use `mlir::computeSomeProperty` if that is what you want? The place that the call is resolved is where ever the interface header is included(just like any other template).

[...]

Issue #194, opened 2019-10-17
Consider trip count in loop invariant code motion


sherhut:
The current loop-invariant code motion pass does not consider trip counts when optimizing loops. This should be added to avoid hoisting code out of loops that actually never execute and hence adding additional computation.

The loop-like interface should probably grow a method to query whether a loop has a statically known trip count. For AffineFor, this can be answered using the mlir::getConstantTripCount(AffineForOp forOp) helper.

bondhugula:
I'll be happy to take this. @River707 @joker-eph Could you please assign to me?

River707:
Why would we even need to worry about loops that would never execute? Shouldn't these just be DCE'd?

bondhugula:
> Why would we even need to worry about loops that would never execute? Shouldn't these just be DCE'd?

Because:
1) at compile time, it may not be known whether they execute at least once;
2) at the time LICM is run, it may not be known whether they execute at least once, i.e., the fact that they are dead may arise only after further transformations. 

As such, we can't hoist unless the lower bound on the trip count is known to be at least one.

River707:
Yep, thanks. I was thinking of something completely different.

PR #203, opened 2019-10-21
LICM pass zero trip count loop handling; zero trip count loop removal in

bondhugula:
-simplify-affine-structures

- addresses #194

- change name of tablegen auto-generated internal helper for op
  interfaces to avoid potential conflicts with methods of same
  name in dialect namespaces. (addresses #197)

### DIFF
--- a/include/mlir/Transforms/LoopLikeInterface.h
+++ b/include/mlir/Transforms/LoopLikeInterface.h
@@ -22,6 +22,7 @@
 #ifndef MLIR_TRANSFORMS_LOOPLIKEINTERFACE_H_
 #define MLIR_TRANSFORMS_LOOPLIKEINTERFACE_H_
 
+#include "mlir/Analysis/LoopAnalysis.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/Support/LogicalResult.h"
 #include "llvm/ADT/ArrayRef.h"
--- a/include/mlir/Transforms/LoopLikeInterface.td
+++ b/include/mlir/Transforms/LoopLikeInterface.td
@@ -56,6 +56,10 @@ def LoopLikeOpInterface : OpInterface<"LoopLikeOpInterface"> {
       }],
       "LogicalResult", "moveOutOfLoop", (ins "ArrayRef<Operation *>":$ops)
     >,
+    InterfaceMethod<"Get the trip count if it is a constant.",
+      "llvm::Optional<uint64_t>", "getConstantTripCount", (ins), [{
+      return getConstantTripCount(op);
+    }]>,
   ];
 }
 
--- a/lib/Dialect/LoopOps/LoopOps.cpp
+++ b/lib/Dialect/LoopOps/LoopOps.cpp
@@ -140,7 +140,7 @@ bool ForOp::isDefinedOutsideOfLoop(Value *value) {
 
 LogicalResult ForOp::moveOutOfLoop(ArrayRef<Operation *> ops) {
   for (auto *op : ops)
-    op->moveBefore(this->getOperation());
+    op->moveBefore(*this);
   return success();
 }
 
@@ -151,6 +151,31 @@ ForOp mlir::loop::getForInductionVarOwner(Value *val) {
   assert(ivArg->getOwner() && "unlinked block argument");
   auto *containingInst = ivArg->getOwner()->getParentOp();
   return dyn_cast_or_null<ForOp>(containingInst);
+}
+
+Optional<uint64_t> mlir::loop::getConstantTripCount(ForOp forOp) {
+  Value *lb = forOp.lowerBound();
+  Value *ub = forOp.upperBound();
+
+  if (lb == ub)
+    return 0;
+
+  IntegerAttr lbCst, ubCst, step;
+  if (!matchPattern(lb, m_Constant(&lbCst)) ||
+      !matchPattern(ub, m_Constant(&ubCst)))
+    return llvm::None;
+
+  int64_t lbConst = lbCst.getValue().getSExtValue();
+  int64_t ubConst = ubCst.getValue().getSExtValue();
+  if (ubConst - lbConst <= 0)
+    return 0;
+
+  if (!matchPattern(forOp.step(), m_Constant(&step)))
+    return llvm::None;
+
+  // Step is guaranteed to be positive.
+  int64_t stepConst = step.getValue().getSExtValue();
+  return llvm::divideCeil(ubConst - lbConst, stepConst);
 }
 
 //===----------------------------------------------------------------------===//
--- a/lib/Transforms/LoopInvariantCodeMotion.cpp
+++ b/lib/Transforms/LoopInvariantCodeMotion.cpp
@@ -62,7 +62,7 @@ static bool canBeHoisted(Operation *op,
   auto thisOpIsSideEffecting = sideEffecting;
   if (thisOpIsSideEffecting != SideEffecting::Never) {
     thisOpIsSideEffecting = interface.isSideEffecting(op);
-    // If the op always has sideeffects, we cannot hoist.
+    // If the op always has side effects, we cannot hoist.
     if (thisOpIsSideEffecting == SideEffecting::Always)
       return false;
   }
@@ -70,9 +70,7 @@ static bool canBeHoisted(Operation *op,
   // can be hoisted.
   for (auto &region : op->getRegions()) {
     for (auto &block : region.getBlocks()) {
-      for (auto &innerOp : block) {
-        if (innerOp.isKnownTerminator())
-          continue;
+      for (auto &innerOp : block.without_terminator()) {
         if (!canBeHoisted(&innerOp, definedOutside, thisOpIsSideEffecting,
                           interface))
           return false;
@@ -112,7 +110,7 @@ static LogicalResult moveLoopInvariantCode(LoopLikeOpInterface looplike,
     }
   }
 
-  // For all instructions that we found to be invariant, move outside of the
+  // For all operations that we found to be invariant, move outside of the
   // loop.
   auto result = looplike.moveOutOfLoop(opsToMove);
   LLVM_DEBUG(looplike.print(llvm::dbgs() << "Modified loop\n"));
@@ -126,12 +124,16 @@ void LoopInvariantCodeMotion::runOnOperation() {
   // Walk through all loops in a function in innermost-loop-first order. This
   // way, we first LICM from the inner loop, and place the ops in
   // the outer loop, which in turn can be further LICM'ed.
-  getOperation()->walk([&](Operation *op) {
-    if (auto looplike = dyn_cast<LoopLikeOpInterface>(op)) {
-      LLVM_DEBUG(op->print(llvm::dbgs() << "\nOriginal loop\n"));
-      if (failed(moveLoopInvariantCode(looplike, interface)))
-        signalPassFailure();
-    }
+  getOperation()->walk([&](LoopLikeOpInterface loopLikeOp) {
+    // Skip zero trip count loops. For unknown trip counts, we still move
+    // invariant code since it is side-effect free, and in general profitable.
+    // TODO: when necessary, we could only move when the trip count is
+    // guaranteed to be at least one.
+    if (loopLikeOp.getConstantTripCount() == uint64_t(0))
+      return;
+    LLVM_DEBUG(loopLikeOp.print(llvm::dbgs() << "\nOriginal loop\n"));
+    if (failed(moveLoopInvariantCode(loopLikeOp, interface)))
+      signalPassFailure();
   });
 }
 
@@ -146,4 +148,4 @@ std::unique_ptr<Pass> mlir::createLoopInvariantCodeMotionPass() {
 
 static PassRegistration<LoopInvariantCodeMotion>
     pass("loop-invariant-code-motion",
-         "Hoist loop invariant instructions outside of the loop");
+         "Hoist loop invariant operations outside of the loop");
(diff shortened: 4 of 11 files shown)


lib/Dialect/LoopOps/LoopOps.cpp
```diff
@@ -153,6 +153,27 @@ ForOp mlir::loop::getForInductionVarOwner(Value *val) {
...
+  if (!matchPattern(lb, m_Constant(&lbCst)))
+    return llvm::None;
+  if (!matchPattern(ub, m_Constant(&ubCst)))
+    return llvm::None;
+
+  int64_t lbConst = lbCst.getValue().getSExtValue();
+  int64_t ubConst = ubCst.getValue().getSExtValue();
+
+  if (ubConst > lbConst)
+    return ubConst - lbConst;
```
River707: You also need to check the step, it isn't guaranteed to be one.

lib/Transforms/SimplifyAffineStructures.cpp
```diff
@@ -110,8 +111,21 @@ void SimplifyAffineStructures::runOnFunction() {
...
+
+  // Remove zero trip count loops.
+  // TODO: templatize walk methods on operation interfaces.
+  // TODO: this could be moved to a more appropriate place.
+  func.walk([&](Operation *op) {
+    auto loopOp = dyn_cast<LoopLikeOpInterface>(op);
+    if (!loopOp)
+      return;
+    auto tripCount = loopOp.getConstantTripCount();
+    if (tripCount.hasValue() && tripCount.getValue() == 0)
```
River707: nit: On these compares you don't have to check for 'hasValue'. You can just compare directly: tripCount == 0

tools/mlir-tblgen/OpInterfacesGen.cpp
```diff
@@ -41,7 +41,10 @@ using mlir::tblgen::OpInterfaceMethod;
 // beginning of the argument list.
 static void emitMethodNameAndArgs(const OpInterfaceMethod &method,
                                   raw_ostream &os, bool addOperationArg) {
-  os << method.getName() << '(';
+  // Whenever an operation argument is added, suffix helper method name with an
+  // underscore to avoid conflicts with free functions of same name on the
+  // concrete ops using this interface.
+  os << method.getName() << (addOperationArg? "_(" : "(");
```
River707: Needs formatting.

River707 (changes requested):

include/mlir/Transforms/LoopLikeInterface.td
```diff
@@ -56,6 +56,12 @@ def LoopLikeOpInterface : OpInterface<"LoopLikeOpInterface"> {
       }],
       "LogicalResult", "moveOutOfLoop", (ins "ArrayRef<Operation *>":$ops)
     >,
+    InterfaceMethod<[{
+        Get the trip count if it is a constant.
```
River707: nit: Can you just use a single line string here?

lib/Dialect/LoopOps/LoopOps.cpp
```diff
@@ -153,6 +153,32 @@ ForOp mlir::loop::getForInductionVarOwner(Value *val) {
...
+  int64_t lbConst = lbCst.getValue().getSExtValue();
+  int64_t ubConst = ubCst.getValue().getSExtValue();
+  if (ubConst - lbConst <= 0)
+    return 0;
+
+  if (!matchPattern(forOp.step(), m_Constant(&step)))
+    return llvm::None;
+
+  // Step is guaranteed to be positive.
+  int64_t stepConst = step.getValue().getSExtValue();
```
River707: Is this verified anywhere?

lib/Transforms/SimplifyAffineStructures.cpp
```diff
@@ -110,8 +111,21 @@ void SimplifyAffineStructures::runOnFunction() {
   for (auto allocOp : allocOps) {
     normalizeMemRef(allocOp);
   }
+
+  // Remove zero trip count loops.
+  // TODO: templatize walk methods on operation interfaces.
+  // TODO: this could be moved to a more appropriate place.
+  func.walk([&](Operation *op) {
```
River707: walk([&](LoopLikeOpInterface loopOp) {
  ...
}

?

test/Transforms/loop-invariant-code-motion.mlir
```diff
@@ -199,10 +199,54 @@ func @invariant_affine_nested_if_else() {
...
+  %cf7 = constant 7.0 : f32
+  %N = constant 0 : index
+
+  affine.for %arg0 = 0 to %N {
+    affine.for %arg1 = 0 to 10 {
+      %v0 = addf %cf7, %cf7 : f32
+    }
+  }
+  // CHECK: %0 = alloc() : memref<10xf32>
+  // CHECK-NEXT: %cst = constant 7.000000e+00 : f32
```
River707: Can we avoid hardcoding the SSA value names?

lib/Dialect/LoopOps/LoopOps.cpp
```diff
@@ -153,6 +153,32 @@ ForOp mlir::loop::getForInductionVarOwner(Value *val) {
...
+Optional<uint64_t> mlir::loop::getConstantTripCount(ForOp forOp) {
+  Value *lb = forOp.lowerBound();
+  Value *ub = forOp.upperBound();
+
+  if (lb == ub)
+    return 0;
+
+  IntegerAttr lbCst, ubCst, step;
+  if (!matchPattern(lb, m_Constant(&lbCst)))
+    return llvm::None;
```
River707: if (!... || !...) ?

River707 (changes requested):

lib/Dialect/LoopOps/LoopOps.cpp
bondhugula: Yes, in verify(ForOp).

include/mlir/Transforms/LoopLikeInterface.td
```diff
@@ -56,6 +56,10 @@ def LoopLikeOpInterface : OpInterface<"LoopLikeOpInterface"> {
       }],
       "LogicalResult", "moveOutOfLoop", (ins "ArrayRef<Operation *>":$ops)
     >,
+    InterfaceMethod<[{ Get the trip count if it is a constant. }],
```
River707: It's just a string block so you can use 
"..."

lib/Transforms/SimplifyAffineStructures.cpp
```diff
@@ -110,8 +111,17 @@ void SimplifyAffineStructures::runOnFunction() {
   for (auto allocOp : allocOps) {
     normalizeMemRef(allocOp);
   }
+
+  // Remove zero trip count loops.
+  // TODO: this could be moved to a more appropriate place.
+  func.walk([&](LoopLikeOpInterface loopOp) {
+    auto tripCount = loopOp.getConstantTripCount();
+    if (tripCount == 0UL)
```
River707: Just inline the tripCount variable?

River707 (approved):

lib/Transforms/LoopInvariantCodeMotion.cpp
```diff
@@ -128,6 +126,13 @@ void LoopInvariantCodeMotion::runOnOperation() {
   // the outer loop, which in turn can be further LICM'ed.
   getOperation()->walk([&](Operation *op) {
```
River707: We could also change this to:

walk([&](LoopLikeOpInterface loopLike) {
  ...
}

lib/Transforms/LoopInvariantCodeMotion.cpp
```diff
@@ -126,12 +124,16 @@ void LoopInvariantCodeMotion::runOnOperation() {
...
-      LLVM_DEBUG(op->print(llvm::dbgs() << "\nOriginal loop\n"));
-      if (failed(moveLoopInvariantCode(looplike, interface)))
-        signalPassFailure();
-    }
+  getOperation()->walk([&](LoopLikeOpInterface loopLikeOp) {
+    // Skip zero trip count loops. For unknown trip counts, we still move
+    // invariant code since it is side-effect free, and in general profitable.
+    // TODO: when necessary, we could only move when the trip count is
+    // guaranteed to be at least one.
+    if (loopLikeOp.getConstantTripCount() == 0UL)
```
River707: This doesn't play well depending on the typedef for size_t. Can you change both of the `0UL`s to be `size_t(0)` instead?

bondhugula: Done. (I assume you meant uint64_t(0)).

River707: yep thanks, I don't know why I typed size_t.

include/mlir/Transforms/LoopLikeInterface.h
```diff
@@ -22,6 +22,7 @@
 #ifndef MLIR_TRANSFORMS_LOOPLIKEINTERFACE_H_
 #define MLIR_TRANSFORMS_LOOPLIKEINTERFACE_H_
 
+#include "mlir/Analysis/LoopAnalysis.h"
```
joker-eph: This is a problem in terms of dependencies (Bazel is catching some cycles for us here).

The problem is that while the LoopOps.h which defines the LoopOps dialect is correctly providing a self-contained version of getConstantTripCount, the Affine dialect is not: instead it relies on the `getConstantTripCount` provided by the Analysis library.
Having the affine dialect itself be dependent on the Analysis library, while the Analysis library is also depending on Affine does not seem right.

I suspect all the Affine specific Analysis should be moved into the affine dialect to avoid the cycle.

joker-eph (changes requested):

include/mlir/Transforms/LoopLikeInterface.h
bondhugula: This issue was discussed on this thread: https://groups.google.com/a/tensorflow.org/forum/?utm_medium=email&utm_source=footer#!msg/mlir/6DV_IbyfnQ8/0Ub44e0eDgAJ

While moving the Analysis into the dialect is one solution, the plan from the start has been to keep ops separate from analysis and transforms, and not have a dependence from the dialect ops to analysis/transforms. The issue here is that LoopInterface.h is needed both for "op support" and "analysis support". In this scenario which is a first I think, there is a free function being generated via the operation interface. If we had a separate header for it, we wouldn't have that being included from the affine dialect. Note that the affine dialect is not using getConstantTripCount anywhere inside (only those that already depend on Analysis use it). It's actually odd / out of line that something from "mlir/Transforms/" is being included into include/mlir/Dialect/AffineOps/AffineOps.h! (see here:
https://github.com/tensorflow/mlir/blob/2bd4e0786ccb652c5c97fde037aafbe17e7ea186/include/mlir/Dialect/AffineOps/AffineOps.h#L31 )

So, the larger issue is that: since operation interfaces could be used to generate reusable op get/set methods as well as analysis/transform methods, shouldn't we be segregating their headers? If we want to go with merging affine analysis and ops, at some point, the issue of merging standardops and analysis/transforms is also likely to come up.

joker-eph: I think in general we'll have something like:

- lib/Transforms & lib/Analysis : should contain core logic, using only interfaces. These should not depend on dialects like Affine.
- lib/Dialect/XYZ/Transforms & lib/Dialect/XYZ/Analysis : dialect specific passes. For instance we have already lib/Dialect/Linalg/Analysis/ and lib/Dialect/Linalg/Transforms/
- include/mlir/Interfaces/...  : self-contained interface headers intended to implemented/overridden in dialects and used everywhere.

(I haven't thought much further, there might be another reasonable and scalable option)

bondhugula: Yes, but this doesn't solve the dependence cycle issue we have here unless you have:

include/mlir/Interfaces/Ops/...
include/mlir/Interfaces/Analysis/
include/mlir/Interfaces/Transforms/

(Because the cycle in context is created due to dialect ops depending on Interfaces and Interfaces depending on Analysis. Having the above separation would mean a Dialect/XYZ/Ops would only depend on InterfaceOps and InterfaceAnalysis can depend on Analysis/ or any Dialect/XYZ/Analysis. )

joker-eph: > the cycle in context is created due to dialect ops depending on Interfaces and Interfaces depending on Analysis

Can you clarify how is this interface depending on Analysis?

I don't see a direct dependency other than the Affine situation. For instance the use of this interface in the loop dialect does not pull anything from Analysis as far as I can tell.

bondhugula: > > the cycle in context is created due to dialect ops depending on Interfaces and Interfaces depending on Analysis
> 
> Can you clarify how is this interface depending on Analysis?

The interface method getConstantTripCount depends on the analysis method mlir::getConstantTripCount for AffineForOp, but for loop::ForOp depends on ForOp::getConstantTripCount, which is simpler (doesn't need analysis) and is thus an op method.

> 
> I don't see a direct dependency other than the Affine situation. For instance the use of this interface in the loop dialect does not pull anything from Analysis as far as I can tell.

That's because: for loop::forOp, mlir::loop::getConstantTripCount doesn't rely on Analysis and lives on loop::ForOp itself. It's a method of loop::ForOp defined in Dialect/LoopOps/LoopOps.cpp unlike the free function mlir::getConstantTripCount(AffineForOp) in mlir/Analysis/LoopAnalysis.cpp.

joker-eph: I agree with what you wrote, but my conclusion is that the interface itself does not depends on Analysis. Only the affine dialect does.

PR #204, opened 2019-10-21
fix asm printer for affine expr (#201)

### LINKED ISSUE #201: affine expression print bug - parenthesis ellision
```
$ cat /tmp/bug.mlir 
#map = (d0)[s0] -> ((-s0 floordiv 4) + 1024)
func @foo(memref<10xf32, #map>)
```
The above is printed with an extra left parenthesis.
```
$ mlir-opt /tmp/bug.mlir
#map0 = (d0)[s0] -> ((-s0 floordiv 4 + 1024)
module {
  func @foo(memref<10xf32, #map0>)
}
```

```
$ mlir-opt /tmp/bug.mlir  | mlir-opt
<stdin>:4:1: error: expected ',' or ')'
module {
^
```

bondhugula:
- fixes #201 and another affine expr print bug

### DIFF
--- a/test/IR/affine-map.mlir
+++ b/test/IR/affine-map.mlir
@@ -174,6 +174,9 @@
 // CHECK: #map{{[0-9]+}} = (d0) -> (d0 - (d0 + 1))
 #map53 = (d0) -> (d0 - (d0 + 1))
 
+// CHECK: #map{{[0-9]+}} = (d0)[s0] -> ((-s0) floordiv 4, d0 floordiv -1)
+#map54 = (d0)[s0] -> (-s0 floordiv 4, d0 floordiv -1)
+
 // Single identity maps are removed.
 // CHECK: func @f0(memref<2x4xi8, 1>)
 func @f0(memref<2x4xi8, #map0, 1>)
@@ -343,3 +346,6 @@ func @f52(memref<1xi8, #map52>)
 
 // CHECK: func @f53(memref<1xi8, #map{{[0-9]+}}>)
 func @f53(memref<1xi8, #map53>)
+
+// CHECK: func @f54(memref<10xi32, #map{{[0-9]+}}>)
+func @f54(memref<10xi32, #map54>)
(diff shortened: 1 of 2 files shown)


test/IR/affine-map.mlir
```diff
@@ -174,6 +174,9 @@
 // CHECK: #map{{[0-9]+}} = (d0) -> (d0 - (d0 + 1))
 #map53 = (d0) -> (d0 - (d0 + 1))
 
+// CHECK: #map{{[0-9]+}} = (d0)[s0] -> ((-s0) floordiv 4, d0 floordiv -1)
+#map54 = (d0)[s0] -> (-s0 floordiv 4, d0 floordiv -1)
```
ftynse: The RHS of floordiv is expected to be positive according to the spec. I'm assuming that's the reason why the code originally assumed the -1 RHS implied multiplication.  Maybe we should fix the affine verifier instead to disallow negative constants with floordiv/ceildiv/mod.

bondhugula: > The RHS of floordiv is expected to be positive according to the spec. I'm assuming that's the reason why the code originally assumed the -1 RHS implied multiplication. 

No, that was due to another bug. The change in lib/IR/AsmPrinter.cpp in this commit makes it clear: d0 floordiv -1 or d0 ceildiv -1 was being turned into -d0 due to a missing guard!

> Maybe we should fix the affine verifier instead to disallow negative constants with floordiv/ceildiv/mod.

Negative constants can appear later in the IR as a result of simplification and they could appear on branches that are never taken (basically dead code that is not yet marked dead). As such, they have to be just preserved: this includes floordiv/ceildiv/mod w.r.t zero as well besides negative values. Such expressions are left as is and aren't simplified or folded.

bondhugula:
Anything more on this? All test cases pass for me, and there isn't detail on what's failing internally. @ftynse

River707:
Just fixed it, should be out soon. Apologies for the delay.

Issue #174, opened 2019-10-08
[spirv] Define cast ops


antiagainst:
We need to support common cast ops. See the spec ["3.32.11. Conversion Instructions"](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_conversion_a_conversion_instructions) for the list. Right now we can ignore those requiring `Kernel` capability.

denis0x0D:
@antiagainst @MaheshRavishankar does anyone work on it?

antiagainst:
No. Please feel free to take it. :)

denis0x0D:
@antiagainst thanks!

PR #208, opened 2019-10-24
[spirv] Add cast operations.

denis0x0D:
This patch addresses https://github.com/tensorflow/mlir/issues/174

Add cast operations: OpConvertFToU, OpConvertFToS,
OpConvertSToF, OpConvertUToF, OpUConvert, OpSConvert, OpSConvert.

Thanks!

### DIFF
--- a/include/mlir/Dialect/SPIRV/SPIRVCastOps.td
+++ b/include/mlir/Dialect/SPIRV/SPIRVCastOps.td
@@ -1,0 +1,336 @@
+//===-- SPIRVCastOps.td - MLIR SPIR-V Cast Ops -------*- tablegen -*-------===//
+//
+// Copyright 2019 The MLIR Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+//
+// This file contains cast ops for the SPIR-V dialect. It corresponds
+// to "3.32.11. Convertion Instructions" of the SPIR-V specification.
+//
+//===----------------------------------------------------------------------===//
+
+#ifdef SPIRV_CAST_OPS
+#else
+#define SPIRV_CAST_OPS
+
+#ifdef SPIRV_BASE
+#else
+include "mlir/SPIRV/SPIRVBase.td"
+#endif // SPIRV_BASE
+
+class SPV_CastOp<string mnemonic, Type resultType, Type operandType,
+                 list<OpTrait> traits = []> :
+      SPV_Op<mnemonic,
+             !listconcat(traits,
+                         [NoSideEffect, SameOperandsAndResultShape])> {
+  let arguments = (ins
+    SPV_ScalarOrVectorOf<operandType>:$operand
+  );
+
+  let results = (outs
+    SPV_ScalarOrVectorOf<resultType>:$result
+  );
+
+  let parser = [{ return mlir::impl::parseCastOp(parser, result); }];
+  let printer = [{ mlir::impl::printCastOp(this->getOperation(), p); }];
+  let verifier = [{ return verifyCastOp(this->getOperation()); }];
+}
+
+// -----
+
+def SPV_BitcastOp : SPV_Op<"Bitcast", [NoSideEffect]> {
+  let summary = "Bit pattern-preserving type conversion.";
+
+  let description = [{
+    Result Type must be an OpTypePointer, or a scalar or vector of
+    numerical-type.
+
+    Operand must have a type of OpTypePointer, or a scalar or vector of
+    numerical-type. It must be a different type than Result Type.
+
+    If either Result Type or Operand is a pointer, the other must be a
+    pointer (diverges from the SPIR-V spec).
+
+    If Result Type has a different number of components than Operand, the
+    total number of bits in Result Type must equal the total number of bits
+    in Operand. Let L be the type, either Result Type or Operand’s type,
+    that has the larger number of components. Let S be the other type, with
+    the smaller number of components. The number of components in L must be
+    an integer multiple of the number of components in S. The first
+    component (that is, the only or lowest-numbered component) of S maps to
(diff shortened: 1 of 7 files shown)


include/mlir/Dialect/SPIRV/SPIRVCastOps.td
```diff
@@ -0,0 +1,281 @@
+//===-- SPIRVCastOps.td - MLIR SPIR-V Cast Ops -*- tablegen -*-===//
```
antiagainst: Nit: Fill the line to 80 characters. :)

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -154,6 +154,43 @@ printMemoryAccessAttribute(LoadStoreOpTy loadStoreOp, OpAsmPrinter &printer,
   elidedAttrs.push_back(spirv::attributeName<spirv::StorageClass>());
 }
 
+template <typename CastOp>
+static LogicalResult verifyCastOp(CastOp castOp,
+                                  bool sameOperandAndResultBitWidth = true) {
```
antiagainst: Nit: "requireSameBitwidth"?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -154,6 +154,43 @@ printMemoryAccessAttribute(LoadStoreOpTy loadStoreOp, OpAsmPrinter &printer,
...
+  const auto operandTypeBitWidth = operandType.getIntOrFloatBitWidth();
+  const auto resultTypeBitWidth = resultType.getIntOrFloatBitWidth();
+  const auto isSameBitWidth = operandTypeBitWidth == resultTypeBitWidth;
+
+  if (sameOperandAndResultBitWidth) {
+    if (!isSameBitWidth) {
+      return castOp.emitOpError(
+                 "expected the same bit width for operand type and result "
+                 "type, but provided ")
+             << operandTypeBitWidth << "-bit width for operand type, "
```
antiagainst: Nit: just print `operandType` and `resultType` here .

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -154,6 +154,43 @@ printMemoryAccessAttribute(LoadStoreOpTy loadStoreOp, OpAsmPrinter &printer,
...
+             << resultTypeBitWidth << "-bit width for result type";
+    }
+    return success();
+  }
+
+  if (isSameBitWidth) {
+    return castOp.emitOpError(
+               "expected the different bit width for operand type and result "
+               "type, but provided ")
+           << operandTypeBitWidth << "-bit width for operand type, "
```
antiagainst: Same here.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -154,6 +154,43 @@ printMemoryAccessAttribute(LoadStoreOpTy loadStoreOp, OpAsmPrinter &printer,
...
+                 "type, but provided ")
+             << operandTypeBitWidth << "-bit width for operand type, "
+             << resultTypeBitWidth << "-bit width for result type";
+    }
+    return success();
+  }
+
+  if (isSameBitWidth) {
+    return castOp.emitOpError(
+               "expected the different bit width for operand type and result "
```
antiagainst: expected different bit widths ...

include/mlir/Dialect/SPIRV/SPIRVCastOps.td
```diff
@@ -0,0 +1,281 @@
...
+                         [NoSideEffect, SameOperandsAndResultShape])> {
+  let arguments = (ins
+    SPV_ScalarOrVectorOf<operandType>:$operand
+  );
+
+  let results = (outs
+    SPV_ScalarOrVectorOf<resultType>:$result
+  );
+
+  let parser = [{ return mlir::impl::parseCastOp(parser, result); }];
```
antiagainst: This is inconsistent with spv.Bitcast. Can you upate spv.Bitcast to also use this?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -154,6 +154,43 @@ printMemoryAccessAttribute(LoadStoreOpTy loadStoreOp, OpAsmPrinter &printer,
   elidedAttrs.push_back(spirv::attributeName<spirv::StorageClass>());
 }
 
+template <typename CastOp>
```
antiagainst: Do we need a template here? I think just using a `Operation *op` should be fine here?

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -154,6 +154,43 @@ printMemoryAccessAttribute(LoadStoreOpTy loadStoreOp, OpAsmPrinter &printer,
...
+  if (auto vectorType = operandType.dyn_cast<VectorType>()) {
+    operandType = vectorType.getElementType();
+    resultType = resultType.cast<VectorType>().getElementType();
+  }
+
+  const auto operandTypeBitWidth = operandType.getIntOrFloatBitWidth();
+  const auto resultTypeBitWidth = resultType.getIntOrFloatBitWidth();
+  const auto isSameBitWidth = operandTypeBitWidth == resultTypeBitWidth;
+
+  if (sameOperandAndResultBitWidth) {
```
antiagainst: I don't think the spec requires `OpConvertFToS`/etc. to be of the same bitwidth?

denis0x0D: @antiagainst thanks for review, I was looking into DirectXShaderCompiler for this ops https://github.com/microsoft/DirectXShaderCompiler/blob/master/tools/clang/lib/SPIRV/SpirvEmitter.cpp#L6623 , as far as I understood those code, before creating an instruction `OpConvertFToS` it explicitly converts a bit width of `fromType` into bit width of `toType`  by `OpFConvert` instruction.

> // First convert the source to the bitwidth of the destination if necessary.

please fix me if I'm wrong.
Thanks.

MaheshRavishankar: The spec itself does not seem to have this restriction. But I think it is better that in the SPIR-V dialect we enforce the bit width to be same for the conversion and use OpFConvert/OpSConvert to model the bitwidth conversions.

antiagainst: @denis0x0D: haha, good find! Actually I cannot recall exactly why we did that in DXC... It might be the spec wording changed or something. What Mahesh said makes sense. Fine for me to keep it.

denis0x0D:
squashed into one coomit and rebased on master.

PR #209, opened 2019-10-24
Identified structs

schweitzpgi:
Add constructors for LLVM identified struct types.  This is to support recursive type references.  The interface follows that found in LLVM:;StructType's public interface with create and setBody functions.

The test for this is commented out because roundtripping does not work.  However, using the interface does generate the expected LLVM code.


ftynse:
What specifically does not work with roundtripping? Is it the parser?

River707:
@ftynse yes, we can't use the standard LLVM type parser to support named structs. It relies on the fact that the struct types have already been printed/parsed and elides the body. To support these we will likely need to re-implement the parser in MLIR that supports inline parsing/printing.

schweitzpgi:
> What specifically does not work with roundtripping? Is it the parser?

To echo River's response, yes, the problem is with the parser.

PR #211, opened 2019-10-27
Bugfix: Keep worklistMap in sync with worklist in GreedyPatternRewriter

dcaballe:
When we removed a pattern, we removed it from worklist but not from
worklistMap. Then, when we tried to add a new pattern on the same Operation
again, the pattern wasn't added since it already existed in the
worklistMap (but not in the worklist).


joker-eph:
Do you see a way to exercise this in a test?

dcaballe:
> Do you see a way to exercise this in a test?

I couldn't find a simple way. This bug was exposed by https://github.com/tensorflow/mlir/pull/212 and only one test failed out of multiple tests using that feature. At least, that test is indirectly covering it. IIRC, the failure only happened when adding and removing patterns for 2 ops multiple times. Please, let me know if you have any idea.

Issue #199, opened 2019-10-18
Add a method to apply specified patterns alone - only on the matching ops?


bondhugula:
It doesn't look like there is a way to apply a specific rewrite pattern in isolation, i.e., without triggering folding and dead code elimination everywhere in the op (its regions), which would include processing ops that don't match and have nothing to do with the specified pattern rewrites. Would it be meaningful to have a flag on applyPatternsGreedily or another version of it that instead will not perform any unrelated folding and dead code elimination? (It can of course remove any ops that become dead as a result of the specified pattern rewrites). The issue here is really the logic of GreedyPatternRewriteDriver::simplify(...). https://github.com/tensorflow/mlir/blob/51a9eb030c8dee92f1ff2669378bec19c4418788/lib/Transforms/Utils/GreedyPatternRewriteDriver.cpp#L149

Having such a version is necessary to easily write test cases for specific patterns, and to be more efficient with the rewriting (not rewrite large parts of the IR unrelated to the pattern rewrites that have been supplied to the driver) -- this is especially the case when called on the function ops (most common) or ops with regions.

joker-eph:
If you have a use-case for it, what about adding a flag/option on the GreedyPatternRewriteDriver to not fold?

bondhugula:
> If you have a use-case for it, what about adding a flag/option on the GreedyPatternRewriteDriver to not fold?

We actually want it to fold (and eliminate dead ops), but only in conjunction with the matched patterns. The issue is that it adds *all* ops (under the supplied op) to the worklist irrespective of whether they'll be matched.

PR #212, opened 2019-10-27
Implement removeOpIfDeadResults in PatternRewriter

dcaballe:
'replaceOp' and 'replaceOpWithNewOp' accept an list of `valuesToRemoveIfDead`.
However, such removal functionality wasn't implemented.

There are quite a few tests using this feature already:
```
    MLIR :: AffineOps/canonicalize.mlir
    MLIR :: Dialect/FxpMathOps/lower-uniform-casts.mlir
    MLIR :: Dialect/FxpMathOps/lower-uniform-real-math-addew.mlir
    MLIR :: Dialect/FxpMathOps/lower-uniform-real-math-mulew.mlir
    MLIR :: Dialect/GPU/canonicalize.mlir
    MLIR :: Dialect/QuantOps/canonicalize.mlir
    MLIR :: Dialect/QuantOps/convert-const.mlir
    MLIR :: Dialect/QuantOps/convert-fakequant.mlir
    MLIR :: Dialect/SPIRV/canonicalize.mlir
    MLIR :: Examples/Toy/Ch4/shape_inference.mlir
    MLIR :: Examples/Toy/Ch5/affine-lowering.mlir
    MLIR :: Examples/Toy/Ch5/shape_inference.mlir
    MLIR :: Examples/Toy/Ch6/affine-lowering.mlir
    MLIR :: Examples/Toy/Ch6/llvm-lowering.mlir
    MLIR :: Examples/Toy/Ch6/shape_inference.mlir
    MLIR :: Quantizer/matmul.mlir
    MLIR :: Quantizer/remove-instrumentation.mlir
    MLIR :: Transforms/Vectorize/lower_vector_transfers.mlir
    MLIR :: Transforms/canonicalize.mlir
    MLIR :: Transforms/inlining.mlir
    MLIR :: mlir-tblgen/pattern.mlir
```
However, maybe adding a couple of tests to cover Operation with multiple results could be a good idea. Any suggestion of a multi-result Operation and could trigger this feature?

This patch also needs https://github.com/tensorflow/mlir/pull/211 to land. Otherwise, `Transforms/canonicalize.mlir` will fail.


lib/IR/PatternMatch.cpp
```diff
@@ -79,6 +82,44 @@ PatternRewriter::~PatternRewriter() {
...
+  valuesToRemoveMap.insert(valuesToRemoveIfDead.begin(),
+                           valuesToRemoveIfDead.end());
+  for (auto *toRemove : valuesToRemoveIfDead) {
+    auto toRemoveIt = valuesToRemoveMap.find(toRemove);
+    if (toRemoveIt == valuesToRemoveMap.end())
+      // This Value was already processed.
+      continue;
+
+    auto *opToRemove = toRemove->getDefiningOp();
+    auto opResults = opToRemove->getResults();
```
River707: This seems overly complex, and potentially incorrect. We shouldn't require that all results be in the `valuesToRemoveIfDead` set, and we can simplify by just checking that the operation is use_empty. This isn't correct though, because we can't opaquely remove operations that have side effects. Orthogonally to that, this isn't running until fixed point, so there will be no propagation of deadness to the operands of the newly dead operations. I wonder if it's just better to add these to the worklist(in the GreedyPatternRewriter case) so that we don't have to reimplement the driver and deadness checking logic.

River707 (changes requested):

lib/IR/PatternMatch.cpp
dcaballe: Thanks for the feedback!

> This seems overly complex, and potentially incorrect. We shouldn't require that all results be in the valuesToRemoveIfDead set, and we can simplify by just checking that the operation is use_empty.

I implemented multiple flavors of this. The first one was this simplified version that you mentioned. However, I thought it would be problematic for Operations with multiple uses. For example, imagine we have an Operation that produces two results but only one of them is in `valuesToRemoveIfDead`. Are we sure we want to remove the Operation if both results are dead? What if the client was planning to do something with the other result in the future? I think that maybe we should be more conservative here since we are doing an "in-place DCE" in the middle of a code rewrite. What do you think?

I have another flavor that changes `valuesToRemoveIfDead` from `ArrayRef<Value *> to `ArrayRef<Operation *> (https://github.com/dcaballe/mlir/tree/dcaballe/remove-dead). This change moves the responsibility of deciding if an Operation must be removed to the client. However, I didn't like it much after seeing the existing use cases.

Why do you think the current approach is potentially incorrect?

> This isn't correct though, because we can't opaquely remove operations that have side effects.

Nice, I could add that check as well.

> Orthogonally to that, this isn't running until fixed point, so there will be no propagation of deadness to the operands of the newly dead operations. I wonder if it's just better to add these to the worklist(in the GreedyPatternRewriter case) so that we don't have to reimplement the driver and deadness checking logic.

I guess it depends on how far we want to go here. Do we want full-blown DCE? I thought this feature was more like a cheap/best effort attempt to have cleaner code after the rewrite but, yeah, we can try to extend it a bit more if it makes sense.

bondhugula: > 
> I guess it depends on how far we want to go here. Do we want full-blown DCE? I thought this feature was more like a cheap/best effort attempt to have cleaner code after the rewrite but, yeah, we can try to extend it a bit more if it makes sense.

Ideally, we wouldn't want any DCE irrelevant to the supplied patterns happen on the IR. But this isn't currently the case with mlir::applyPatternsGreedily and I consider this a bug / undesired behavior - I had opened https://github.com/tensorflow/mlir/issues/199 on this. The reason for this behavior is I believe the greedy pattern rewriter initially being used for -canonicalize, which is expected to do DCE and folding on everything in the op that was passed. OTOH, applyPatternsGreedily has use cases where a top-level op or say a func op is provided as its argument, and the action is expected just around the matching ops corresponding to the few patterns in the supplied list.

dcaballe: Thanks for the context! I think removing "operands of the newly dead operations" would still fit into "DCE that is relevant to the supplied patterns" so it should be ok. However, I would suggest that we leave changes on `GreedyPatternRewriter` for a follow-up PR. If you could help me create a small test that faithfully represents what must be covered, I could spend some time on it later.

For this PR, I can address the side-effect issue. The remaining point is what we want to do about operations with multiple results. Based on the previous comments, should we remove an Operation if only one of its results is provided in the `valuesToRemoveIfDead` and all the results are dead? Do you think this could be problematic?

dcaballe: Hi @River707, it would be great if I could get your opinion about ^ before I address all the comments. Thanks!

River707: If we know the operation is dead we can just delete, whether all of the results were provided(unlikely) or not. I see this list as a hint to the driver that these operations are likely dead. The main problem though, is that this was added at the very beginning of the pattern framework and we built a lot of infrastructure that already takes care of many/most of the use cases that this would fire in: applyPatternsGreedily operates to a fixed-point, we already add the operands/users to the worklist for consideration in many cases, etc. The only test cases I can think of are ones that either have operands/users that won't get added automatically, or ones where we could remove the need to do an extra fixed point iteration. This can be difficult to construct though unless we have a real use case, but can be valuable.

The main reason I'd rather just have this add items to the greedy pattern worklist is because we(or future developers) will likely want to extend this when cases aren't being caught. At this point we will have two implementations of a fixed point, worklist based algorithm that cover the same thing, with this one being a subset of what the greedy driver is already doing.

Also, I agree with you @dcaballe  that this is fine regardless of the scenario that 'applyPatternsGreedily' is invoked in. This is something explicitly specified by the user, so we should always be able to act on it accordingly.

dcaballe: Thanks, @River707. I addressed all the feedback. Sorry for the delay but I wanted to make sure that this approach was enough for my use case. It turns out that my use case became a bit more complicated and I will need something more sophisticated, which actually supports your point :). I'll leave that discussion for the mailing list.

I agree this could be valuable for simple cases. There are some use cases already in the codebase. For example: 
- https://github.com/tensorflow/mlir/blob/master/lib/Dialect/QuantOps/Transforms/ConvertConst.cpp#L102
- https://github.com/tensorflow/mlir/blob/master/lib/Dialect/StandardOps/Ops.cpp#L447

I think the last one fits into the second category that you mentioned since it uses this mechanism to remove constants (not the best example, though, as canonicalization already takes care of dead code). We could leave more complicated cases for the worklist counterpart but if you feel this is going to be problematic in the future, I can go ahead and remove `removeValuesIfDead` from the API.

Thanks,
Diego

include/mlir/IR/PatternMatch.h
```diff
@@ -403,6 +403,11 @@ class PatternRewriter : public OpBuilder {
   virtual void notifyOperationRemoved(Operation *op) {}
 
 private:
+  /// This method checks if Values in 'valuesToRemoveIfDead' are dead and
+  /// removes their defining Operations. An Operation is removed only if all its
+  /// Value results are dead and has no side effect.
```
kiszk: nit: `has` -> `have`

dcaballe:
@kiszk, thanks for the feedback (not sure why I can't see your comment in github). `has no side effect` refers to `Operation`. Maybe I can reword that to make it clearer.

@River707, what do you think about this PR after the Stripe use case? Should we proceed with this implementation or remove `valuesToRemoveIfDead` from the API? I think it still makes sense for the simple cases I pointed out before but I'll be happy to remove it if you think it doesn't make sense. I just don't want to leave `valuesToRemoveIfDead` in the API without implementation :)

River707:
I apologize for the tardiness, can you share the use case you have for Stripe?

dcaballe:
Sorry, I wasn't clear. Stripe use case is the one that evolved into the BlockArgument conversion problem (https://groups.google.com/a/tensorflow.org/forum/#!topic/mlir/Jpn5eX2Bcpg) so I no longer have a simple case that can benefit from this. However, I still see some value, as I mentioned before.

River707:
Sorry for the delay. I may be biased, but I say we just remove it until there is a clear story in how it fits in with the rest of the infra.

dcaballe:
That should be it!

dcaballe:
Moved to https://reviews.llvm.org/D72545

PR #214, opened 2019-10-30
Allow dialect to create friendly names for block arguments

jbruestle:
Provide a callback similar to getOpResultName but for block arguments.

### DIFF
--- a/include/mlir/IR/OpImplementation.h
+++ b/include/mlir/IR/OpImplementation.h
@@ -594,6 +594,9 @@ public:
   /// Get a special name to use when printing the given operation. The desired
   /// name should be streamed into 'os'.
   virtual void getOpResultName(Operation *op, raw_ostream &os) const {}
+  /// Get a special name to use when printing the block arguments. The desired
+  /// name should be streamed into 'os'.
+  virtual void getBlockArgumentName(BlockArgument *arg, raw_ostream &os) const {}
 };
 
 } // end namespace mlir
--- a/include/mlir/IR/OpImplementation.h
+++ b/include/mlir/IR/OpImplementation.h
@@ -594,9 +594,10 @@ public:
   /// Get a special name to use when printing the given operation. The desired
   /// name should be streamed into 'os'.
   virtual void getOpResultName(Operation *op, raw_ostream &os) const {}
-  /// Get a special name to use when printing the block arguments. The desired
-  /// name should be streamed into 'os'.
-  virtual void getBlockArgumentName(BlockArgument *arg, raw_ostream &os) const {}
+  /// Get a special name to use when printing the entry block arguments for any
+  /// region contained by an Op in this dialect.  The desired name should be
+  /// streamed into 'os'.  If the result is empty, the default name will be used.
+  virtual void getRegionArgumentName(BlockArgument *arg, raw_ostream &os) const {}
 };
 
 } // end namespace mlir
--- a/lib/IR/AsmPrinter.cpp
+++ b/lib/IR/AsmPrinter.cpp
@@ -1518,10 +1518,19 @@ void OperationPrinter::numberValueID(Value *value) {
   llvm::raw_svector_ostream specialName(specialNameBuffer);
 
   // Check to see if this value requested a special name.
-  auto *op = value->getDefiningOp();
-  if (state && op) {
-    if (auto *interface = state->getOpAsmInterface(op->getDialect()))
-      interface->getOpResultName(op, specialName);
+  if (state) {
+    if (auto *op = value->getDefiningOp()) {
+      // If the value is the result of an Op, ask the Op's dialect for a name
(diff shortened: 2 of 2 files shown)


lib/IR/AsmPrinter.cpp
```diff
@@ -1518,10 +1518,19 @@ void OperationPrinter::numberValueID(Value *value) {
...
-  auto *op = value->getDefiningOp();
-  if (state && op) {
-    if (auto *interface = state->getOpAsmInterface(op->getDialect()))
-      interface->getOpResultName(op, specialName);
+  if (state) {
+    if (auto *op = value->getDefiningOp()) {
+      // If the value is the result of an Op, ask the Op's dialect for a name
+      if (auto *interface = state->getOpAsmInterface(op->getDialect()))
+        interface->getOpResultName(op, specialName);
+    } else if (auto* arg = mlir::dyn_cast<mlir::BlockArgument>(value)) {
```
River707: This should be a cast. There are only two types of values.

Also, the hook should only be called on entry block arguments.

Also, drop all of the 'mlir::'.

include/mlir/IR/OpImplementation.h
```diff
@@ -594,6 +594,9 @@ class OpAsmDialectInterface
   /// Get a special name to use when printing the given operation. The desired
   /// name should be streamed into 'os'.
   virtual void getOpResultName(Operation *op, raw_ostream &os) const {}
+  /// Get a special name to use when printing the block arguments. The desired
+  /// name should be streamed into 'os'.
+  virtual void getBlockArgumentName(BlockArgument *arg, raw_ostream &os) const {}
```
River707: BlockArgument -> RegionArgument

include/mlir/IR/OpImplementation.h
```diff
@@ -594,6 +594,9 @@ class OpAsmDialectInterface
   /// Get a special name to use when printing the given operation. The desired
   /// name should be streamed into 'os'.
   virtual void getOpResultName(Operation *op, raw_ostream &os) const {}
+  /// Get a special name to use when printing the block arguments. The desired
```
River707: This should be updated specifically mention that these are only entry block arguments.

lib/IR/AsmPrinter.cpp
jbruestle: So I was doing a dyn_cast on the off chance there would some day be another subclass of Value, but I'm happy to change it to not do that.  Also, I will drop the pointless name-spacing, that was an oversight.  However, what is an 'entry' block argument?  A block argument on the first block in a region?  Is the idea that only those block arguments should be related to parent op's dialect?  Also, are non-entry block arguments even used anywhere?

River707: 'entry' block argument is a block argument on the entry block of a region. We only really allow attaching semantics of any kind to these arguments.

Non-entry block arguments are used to represent phi values, which come into play in any place that has uses of non-structured control flow: llvm, standard, spirv, etc:
https://github.com/tensorflow/mlir/blob/master/g3doc/Rationale.md#block-arguments-vs-phi-nodes

jbruestle: Thanks, we don't use much non-structured flow control, so I didn't realize this, but I love the use of block arguments instead of PHI nodes.  Should be fixed now.

jbruestle:
Thanks for the comments, think I fixed everything.

lib/IR/AsmPrinter.cpp
```diff
@@ -1518,10 +1518,22 @@ void OperationPrinter::numberValueID(Value *value) {
   llvm::raw_svector_ostream specialName(specialNameBuffer);
 
   // Check to see if this value requested a special name.
-  auto *op = value->getDefiningOp();
-  if (state && op) {
-    if (auto *interface = state->getOpAsmInterface(op->getDialect()))
-      interface->getOpResultName(op, specialName);
+  if (state) {
+    if (auto *op = value->getDefiningOp()) {
+      // If the value is the result of an Op, ask the Op's dialect for a name
```
bondhugula: Nit: missing full stop at the end here, and for the one further below.

flaub:
I'm in the process of updating this PR. Part of the challenge is that every time we move LLVM/MLIR forward, we also need to move our BUILD files forward (we also use bazel). How does your team manage this? Is the process to have a separate cmake environment and then transition to bazel? Vice versa?

River707:
Internally at google we maintain both separately+simultaneously. If it helps your team, the TensorFlow copy of the MLIR BUILD files can be viewed at:
https://github.com/tensorflow/tensorflow/blob/master/third_party/mlir/BUILD

flaub:
Yep we use that one (and the autogenerated LLVM one), thanks!

jbruestle:
Updated version submitted by @flaub, closing.

PR #215, opened 2019-11-01
[spirv] Add bit ops

denis0x0D:
This CL added op definitions for a few bit operations:

* OpShiftLeftLogical
* OpShiftRightArithmetic
* OpShiftRightLogical
* OpBitCount
* OpBitReverse
* OpNot

Also moved the definition of spv.BitwiseAnd to follow the
lexicographical order.

@antiagainst @MaheshRavishankar can you please take a look? Thanks!

### DIFF
--- a/lib/Dialect/SPIRV/SPIRVOps.cpp
+++ b/lib/Dialect/SPIRV/SPIRVOps.cpp
@@ -449,6 +449,40 @@ static void printLogicalOp(Operation *logicalOp, OpAsmPrinter &printer) {
   printer << " : " << logicalOp->getOperand(0)->getType();
 }
 
+static ParseResult parseShiftOp(OpAsmParser &parser, OperationState &state) {
+  SmallVector<OpAsmParser::OperandType, 2> operandInfo;
+  Type baseType;
+  Type shiftType;
+  auto loc = parser.getCurrentLocation();
+
+  if (parser.parseOperandList(operandInfo, 2) || parser.parseColon() ||
+      parser.parseType(baseType) || parser.parseComma() ||
+      parser.parseType(shiftType) ||
+      parser.resolveOperands(operandInfo, {baseType, shiftType}, loc,
+                             state.operands)) {
+    return failure();
+  }
+  state.addTypes(baseType);
+  return success();
+}
+
+static void printShiftOp(Operation *op, OpAsmPrinter &printer) {
+  Value *base = op->getOperand(0);
+  Value *shift = op->getOperand(1);
+  printer << op->getName() << ' ' << *base << ", " << *shift << " : "
+          << base->getType() << ", " << shift->getType();
+}
+
+static LogicalResult verifyShiftOp(Operation *op) {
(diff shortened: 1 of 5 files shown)


lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -449,6 +449,34 @@ static void printLogicalOp(Operation *logicalOp, OpAsmPrinter &printer) {
...
+
+static void printShiftOp(Operation *op, OpAsmPrinter &printer) {
+  Value *base = op->getOperand(0);
+  Value *shift = op->getOperand(1);
+  printer << op->getName() << ' ' << *base << ", " << *shift << " : "
+          << base->getType() << ", " << shift->getType();
+}
+
+static LogicalResult verifyShiftOp(Operation *op) {
+  return success(op->getOperand(0)->getType() == op->getResult(0)->getType());
```
antiagainst: Print an error message on failure and also add a test for this? :)

denis0x0D: @antiagainst oh, sorry, I forgot about an error message;
by the way, it seems like an error could be generated only when manually creating this op with builder, not from parser, because while parsing it reuses `baseType` https://github.com/tensorflow/mlir/pull/215/files#diff-1a7b8ff4af6a0a814aa170d35a45ba8bR465
please fix me if I'm wrong.
Thanks.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -474,7 +474,13 @@ static void printShiftOp(Operation *op, OpAsmPrinter &printer) {
 }
 
 static LogicalResult verifyShiftOp(Operation *op) {
-  return success(op->getOperand(0)->getType() == op->getResult(0)->getType());
+  if (op->getOperand(0)->getType() != op->getResult(0)->getType()) {
+    return op->emitError("expected the same types for the first operand and "
```
antiagainst: the same type

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -474,7 +474,13 @@ static void printShiftOp(Operation *op, OpAsmPrinter &printer) {
...
+                         "result, but provided ")
+           << op->getOperand(0)->getType() << ", "
```
antiagainst: s/,/and/

lib/Dialect/SPIRV/SPIRVOps.cpp
antiagainst: Yes you are right. Please add a test with the generic assembly form. Thanks! :)

denis0x0D: @antiagainst thanks for mention generic assembly form! I didn't know about it.

antiagainst:
Bots failure unrelated this this and should be fixed by 799273236a9e44ff55a10f58d98edacd58be0a94.

PR #216, opened 2019-11-01
add LLVM IR to build the fneg instruction

schweitzpgi:
Adding the fneg instruction to the LLVM IR dialect.  We want to target this LLVM instruction from FIR.


ftynse:
Sure, don't hesitate to ping me on CLs immediately (emails tend to drown otherwise)

PR #217, opened 2019-11-01
fix link-time duplicate symbol error

schweitzpgi:
We ran into a duplicate symbols link-time error when calling markAllAnalysesPreserved() in our pass.

This change eliminates the linker error.

### DIFF
--- a/include/mlir/Pass/AnalysisManager.h
+++ b/include/mlir/Pass/AnalysisManager.h
@@ -70,7 +70,7 @@ public:
 
 private:
   /// An identifier used to represent all potential analyses.
-  constexpr static AnalysisID allAnalysesID = {};
+  const static AnalysisID allAnalysesID;
 
   /// The set of analyses that are known to be preserved.
   SmallPtrSet<const void *, 2> preservedIDs;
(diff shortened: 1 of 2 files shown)


joker-eph:
I don't understand this change, can you elaborate why is it needed / how is it fixing it?
Are you using C++17 maybe?

schweitzpgi:
> I don't understand this change, can you elaborate why is it needed / how is it fixing it?
> Are you using C++17 maybe?

Yes. The Flang project uses C++17.

joker-eph:
> > I don't understand this change, can you elaborate why is it needed / how is it fixing it?
> > Are you using C++17 maybe?
> 
> Yes. The Flang project uses C++17.

OK, static constexpr are inline variable now in C++17, but that still does not explain the failure though (or I don't understand the source of the failure).
What is your platform/toolchain version?

schweitzpgi:
> > > I don't understand this change, can you elaborate why is it needed / how is it fixing it?
> > > Are you using C++17 maybe?
> > 
> > 
> > Yes. The Flang project uses C++17.
> 
> OK, static constexpr are inline variable now in C++17, but that still does not explain the failure though (or I don't understand the source of the failure).
> What is your platform/toolchain version?

The problem was on linux when compiling with g++ 8.3.0.

jpienaar:
> It appears to me that this issue may have arisen in a build where some object files where being compiled with C++14 and others with C++17.

Could you retry building fresh with only C++17 set? (if that is what is required for other parts)

schweitzpgi:
It looks like the better fix is to build everything with C++17.  At least that solves this link-time failure.

PR #219, opened 2019-11-02
Enable out-of-tree projects to use MLIR.

sscalpone:
Add CMake rules to copy various tools and files into the install directory.  Installs the include and td sources, as well as generated inc files, in include/mlir.  Installs the binaries mlir-cpu-runner, mlir-cuda-runner, mlir-opt, mlir-translate, and mlir-tblgen in bin.


sscalpone:
@googlebot I signed it!

joker-eph:
Seems like you modified a lot of the include folder, can this be done at the top-level like in LLVM?
See here: https://github.com/llvm/llvm-project/blob/master/llvm/CMakeLists.txt#L1007-L1031

tools/mlir-opt/CMakeLists.txt
```diff
@@ -61,3 +61,14 @@ add_llvm_executable(mlir-opt
...
+install(TARGETS mlir-opt
+  ${export_to_llvmexports}
+  COMPONENT mlir-opt
+  RUNTIME DESTINATION ${LLVM_TOOLS_INSTALL_DIR})
+
+if(NOT LLVM_ENABLE_IDE)
+  add_llvm_install_targets("install-mlir-opt"
+    DEPENDS mlir-opt
+  COMPONENT mlir-opt)
+endif()
```
joker-eph: The LLVM tools don't need this, is this because they are using the `add_llvm_tool` macro instead of add_llvm_executable? Can we align?

This repeating pattern seems like it would be better behind a macro instead of spread all around.

sscalpone:
We're testing the recommended approach in @schweitzpgi branch.

PR #223, opened 2019-11-05
typos

Robertorosmaninho:
The mnemonic "sgt" and "ult" are used twice! The second "slt" should be "sge" for "signed greater than or equal" and the second "ult" should be "ule"  for "unsigned less than or equal".


antiagainst:
Super nit: could you update the commit message to say "Fix typos" and turn the message body into sentences? :)

Robertorosmaninho:
> Super nit: could you update the commit message to say "Fix typos" and turn the message body into sentences? :)

Done! :D

PR #226, opened 2019-11-08
Add integer bit-shift operations to the standard dialect.

ombre5733:
Rename the 'shlis' operation in the standard dialect to 'shift_left'. Add tests
for this operation (these have been missing so far) and add a lowering to the
'shl' operation in the LLVM dialect.

Add also 'shift_right_signed' (lowered to LLVM's 'ashr') and 'shift_right_unsigned'
(lowered to 'lshr').

The original plan was to name these operations 'shift.left', 'shift.right.signed'
and 'shift.right.unsigned'. This works if the operations are prefixed with 'std.'
in MLIR assembly. Unfortunately during import the short form is ambigous with
operations from a hypothetical 'shift' dialect. The best solution seems to omit
dots in standard operations for now.


ombre5733:
This is actually a re-write of #153. Instead of using 'shli', 'shris' and 'shriu' for the shift operations, the more readable names 'shift_left', 'shift_right_signed' and 'shift_right_unsigned' have been chosen.

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -943,8 +943,45 @@ def SignExtendIOp : Std_Op<"sexti",
...
+    The shift_left operation shifts an integer value to the left by a variable
+    amount. The low order bits are filled with zeros.
+
+      %1 = constant 5 : i8                       // %1 is 0b00000101
+      %2 = constant 3 : i8
+      %3 = shift_left %1, %2 : (i8, i8) -> i8    // %3 is 0b00101000
+  }];
+}
+
+def SignedShiftRightOp : IntArithmeticOp<"shift_right_signed"> {
```
River707: Can we keep the operation names similar to the mnemonic, i.e. ShiftRightSignedOp/ShiftRightUnsignedOp

River707:
I'd rather not have multiple conventions for naming signed/unsigned operations. If this gets submitted we should change Div/Rem to have the same spelling.

include/mlir/Dialect/StandardOps/Ops.td
ombre5733: Ok, I can make this change.

ombre5733:
@River707 The question is how to name the div and rem operations. We have `divf`, `divis` and `diviu` (and similar for rem). Should it be `divi_signed` and `divi_unsigned` (or better `div_int_signed`)?

River707:
I don't have a preference for the exact naming for these(I'm somewhat partial to just simply divi/divf), I'm more concerned about keeping them uniform.

ombre5733:
@River707 I have now renamed divis, diviu, remis and remiu to divi_signed, divi_unsigned, remi_signed and remi_unsigned to be consistent with the shift operations. Is this what you had in mind?

test/Transforms/constant-fold.mlir
```diff
@@ -288,63 +288,63 @@ func @simple_diviu() -> (i32, i32, i32) {
...
   %1 = constant dense<2> : tensor<4xi32>
 
   // CHECK-DAG: [[C3:%.+]] = constant dense<3> : tensor<4xi32>
-  %2 = diviu %0, %1 : tensor<4xi32>
+  %2 = divi_unsigned %0, %1 : tensor<4xi32>
 
   %3 = constant dense<-2> : tensor<4xi32>
 
-  // Unsigned division interprets -2 as 2^32-2, so the result is 0.
+  // Unsigned divi_signedion interprets -2 as 2^32-2, so the result is 0.
```
River707: accidental changes?

test/Transforms/lower-affine.mlir
```diff
@@ -436,7 +436,7 @@ func @args_ret_affine_apply(index, index) -> (index, index) {
 }
 
 //===---------------------------------------------------------------------===//
-// Test lowering of Euclidean (floor) division, ceil division and modulo
+// Test lowering of Euclidean (floor) divi_signedion, ceil divi_signedion and modulo
```
River707: Same here and a few other I believe.

test/Transforms/constant-fold.mlir
```diff
@@ -212,74 +212,74 @@ func @mulf_splat_tensor() -> tensor<4xf32> {
...
   %1 = constant 2 : i32
 
   // CHECK-DAG: [[C3:%.+]] = constant 3 : i32
-  %2 = diviu %0, %1 : i32
+  %2 = divi_unsigned %0, %1 : i32
 
   %3 = constant -2 : i32
 
-  // Unsigned division interprets -2 as 2^32-2, so the result is 0.
+  // Unsigned divi_signedion interprets -2 as 2^32-2, so the result is 0.
```
River707: accidental change

test/Transforms/constant-fold.mlir
ombre5733: Yeah, I did the renaming with a script and only skimmed over the changes. I'll fix those.

ombre5733: Fixed.

test/Transforms/lower-affine.mlir
ombre5733: Fixed.

test/Transforms/constant-fold.mlir
ombre5733: Fixed.

ombre5733:
The accidental renamings have been fixed. Can you resolve the conflict in `ConvertStandardToLLVM.cpp` or do I have to take care of this?

ombre5733:
@River707 I also fixed the last conflict with the recent upstream changes. From my point of view, the commit is fine.

ombre5733:
Can I help somehow to unblock the merge?

joker-eph:
Sorry for the delay, it is on us (actually you didn't run `git clang-format` apparently, which blocks the auto-integration, I did it and it should get through soon)

ombre5733:
Oh, sorry for not clang-formatting and thanks for fixing it!

Mehdi Amini <notifications@github.com> schrieb am Sa., 21. Dez. 2019, 06:16:

PR #228, opened 2019-11-09
Add a short TableGen HowTo to tutorial chapter 2.

ombre5733:
Add a note to chapter 2 of the Toy tutorial stating how to invoke
mlir-tblgen to check the generated C++ code. IMHO this is incredibly
useful when getting acquainted with TableGen/ODS.


River707:
Thanks for the improvement!

joker-eph:
I agree: looking at TableGen output is incredibly helpful! I thought I included it in the first version of the tutorial, maybe it was lost during the refresh.
Thanks for the contribution :)

PR #237, opened 2019-11-14
Add getRemappedValue to ConversionPatternRewriter

dcaballe:
ConversionPatternRewritter::getRemappedValue returns the remapped value of an original value that has been previously replaced using ConversionPatternRewritter. This mechanism allows the implementation of M->N conversions where we need to retrieve remapped values of M which aren't necessarily present/accessible in the replacement of N.


include/mlir/Transforms/DialectConversion.h
```diff
@@ -382,6 +382,9 @@ class ConversionPatternRewriter final : public PatternRewriter {
   /// Return a reference to the internal implementation.
   detail::ConversionPatternRewriterImpl &getImpl();
 
+  /// Remap the given value to that with potentially different type.
```
River707: Can you reword this a bit? Maybe mention that this grabs the converted value that replaced 'key'.

Can you also move this to above the `PatternRewriter Hooks` section. This part is intended solely for the implementations of the `PatternRewriter` hooks.

River707:
I would expand the description a bit because it doesn't quite capture the real use case, I believe. It should mention that this is for `M->N->1` conversions where you want the remapped values of M which aren't necessarily present/accessible in the replacement of N. 

Also, can you add a simple test in the TestDialect if possible?

dcaballe:
I addressed the feedback and implemented a simple test that exercises getRemappedValue by using an existing instruction in TestDialect and adding a dummy test pass. I also reworded the PR description but I’m not sure I understand the `M->M->1` in your comment so I used `M->N` instead. Please, let me know if something else is needed.

test/lib/Transforms/TestRemappedValue.cpp
```diff
@@ -0,0 +1,90 @@
+//===- TestInlining.cpp - Pass to inline calls in the test dialect --------===//
+//
+// Copyright 2019 The MLIR Authors.
```
River707: Can you just insert this in the existing TestPatterns.cpp pass? It's under TestDialect/

test/Transforms/test-remapped-value.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: mlir-opt %s -test-remapped-value | FileCheck %s
+
+// Simple test that exercises ConvertPatternRewriter::getRemappedValue.
+func @remap_input_1_to_1(%arg0: i32) {
+  %0 = "test.one_variadic_out_one_variadic_in1"(%arg0) : (i32) -> i32
+  %1 = "test.one_variadic_out_one_variadic_in1"(%0) : (i32) -> i32
+  "test.return"() : () -> ()
+}
+// CHECK-LABEL: func @remap_input_1_to_1
+// CHECK-NEXT: "test.one_variadic_out_one_variadic_in1"(%arg0, %arg0)
```
River707: Can you avoid hard coding the SSA value names here? Can you also add '-legalize' after the 'test' part of the file name? Otherwise it isn't clear what this is testing.

PR #244, opened 2019-11-18
Add OpenMP dialect to the dialect registry

DavidTruby:
This dialect will be used by flang for OpenMP code generation

### DIFF
--- a/include/mlir/IR/DialectSymbolRegistry.def
+++ b/include/mlir/IR/DialectSymbolRegistry.def
@@ -29,6 +29,7 @@ DEFINE_SYM_KIND_RANGE(QUANTIZATION)
 DEFINE_SYM_KIND_RANGE(IREE) // IREE stands for IR Execution Engine
 DEFINE_SYM_KIND_RANGE(LINALG) // Linear Algebra Dialect
 DEFINE_SYM_KIND_RANGE(FIR) // Flang Fortran IR Dialect
+DEFINE_SYM_KIND_RANGE(OPENMP) // OpenMP IR Dialect
 DEFINE_SYM_KIND_RANGE(TOY) // Toy language (tutorial) Dialect
 DEFINE_SYM_KIND_RANGE(SPIRV) // SPIR-V dialect
 


DavidTruby:
Arm has signed the CLA.

@googlebot I signed it!

joker-eph:
Just to confirm: is this dialect gonna be "abstract" or "generic" OpenMP or is it gonna be "Fortran Specific" OpenMP? (for instance you may let things like types creep in). If the later then I rather see it reflected in the name.

joker-eph:
> Just to confirm: is this dialect gonna be "abstract" or "generic" OpenMP or is it gonna be "Fortran Specific" OpenMP? (for instance you may let things like types creep in). If the later then I rather see it reflected in the name.

Also: if this is a generic OpenMP dialect and not Fortran specific, I would expect it to be submitted upstream and not live in the flang tree? Can you clarify your plans about this?

DavidTruby:
The current plan is to make this a generic OpenMP IR without Fortran specific features as we would like to leave open the possibility for clang to use this if(/when?) they move to an mlir-based code generator. We are also currently planning to submit this upstream in mlir and not in the flang tree.

It's possible of course that when we start implementing this we might find we need Fortran specific things anyway, but the plan is to avoid anything Fortran specific if at all possible.

joker-eph:
Great! Thanks, I'd encourage you to send an RFC with the dialect design and the first "milestone" you'd like to reach: it'll likely smooth out patch review :)

DavidTruby:
>I'd encourage you to send an RFC with the dialect design and the first "milestone" you'd like to reach: it'll likely smooth out patch review :) 

Most of the discussion on dialect design has been happening in the flang community. We have some design walkthroughs, I'll see if we can get some cross-conversation going here with the MLIR community.

I have also been preparing a patch to add an OpenMP dialect and a single operation `omp.barrier` so that we can start work on the dialect. This probably qualifies as the first milestone we're aiming for :)

joker-eph:
> Most of the discussion on dialect design has been happening in the flang community. We have some design walkthroughs, I'll see if we can get some cross-conversation going here with the MLIR community.

Thanks! We will definitely need something like this before reviewing patches I think

nicolasvasilache:
I'd be quite interested in also seeing what the end to end path to
execution looks like.

It would be great if some execution flow could be made available early in
the MLIR repo to experiment.

DavidTruby:
Is this waiting on more design info to be merged or can we get this initial registration in before then?

joker-eph:
> Is this waiting on more design info to be merged or can we get this initial registration in before then?

No this is all good, for some reason our integration script failed and I didn't notice. I restarted it and hopefully it should go through! Sorry for the delay.

PR #245, opened 2019-11-18
add some new parser hooks to the parser to support FIR types

schweitzpgi:
allow access to '?', '*', and optional '{' and '}' tokens.

### DIFF
--- a/include/mlir/IR/DialectImplementation.h
+++ b/include/mlir/IR/DialectImplementation.h
@@ -188,6 +188,12 @@ public:
   /// Parse a `}` token.
   virtual ParseResult parseRBrace() = 0;
 
+  /// Parse a '{' token if present
+  virtual ParseResult parseOptionalLBrace() = 0;
+
+  /// Parse a `}` token if present
+  virtual ParseResult parseOptionalRBrace() = 0;
+
   /// Parse a `:` token.
   virtual ParseResult parseColon() = 0;
 
@@ -263,6 +269,12 @@ public:
 
   /// Parse a `...` token if present;
   virtual ParseResult parseOptionalEllipsis() = 0;
+
+  /// Parse a `*` token.
+  virtual ParseResult parseOptionalStar() = 0;
+
+  /// Parse a `?` token.
+  virtual ParseResult parseOptionalQuestion() = 0;
 
   //===--------------------------------------------------------------------===//
   // Attribute Parsing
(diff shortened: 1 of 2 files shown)


include/mlir/IR/DialectImplementation.h
```diff
@@ -188,6 +188,12 @@ class DialectAsmParser {
   /// Parse a `}` token.
   virtual ParseResult parseRBrace() = 0;
 
+  /// Parse a '{' token if present
+  virtual ParseResult parseOptionalLBrace() = 0;
+
```
River707: Can you keep the same ordering as the .cpp file?

include/mlir/IR/DialectImplementation.h
```diff
@@ -264,6 +270,12 @@ class DialectAsmParser {
   /// Parse a `...` token if present;
   virtual ParseResult parseOptionalEllipsis() = 0;
 
+  /// Parse a `*` token.
+  virtual ParseResult parseOptionalStar() = 0;
+
```
River707: Can you try to keep these sorted alphabetically?

lib/Parser/Parser.cpp
```diff
@@ -596,6 +606,16 @@ class CustomDialectAsmParser : public DialectAsmParser {
     return success(parser.consumeIf(Token::r_square));
   }
 
+  /// Parses a '*' if present.
+  ParseResult parseOptionalStar() override {
```
River707: There is a somewhat loose attempt to keep these sorted alphabetically.

include/mlir/IR/DialectImplementation.h
schweitzpgi: You bet.

lib/Parser/Parser.cpp
schweitzpgi: Can you clarify this a bit more? I'm seeing a list of open-thingy, close-thingy pairs with non-optional and optional variants intermingled. But Less, Greater, LParen, RParen, ... Ellipsis doesn't seem to be alphabetical to me.

Maybe just put Question before Star?

Thanks.

River707: Yes. That's why I said "loose". Ellipsis is definitely in the wrong spot.

In general I keep the parseL and parseR next to each other, same with the optional. I'll clean up the outliers in a followup, thanks for pointing them out!

schweitzpgi:
OK, done.

PR #246, opened 2019-11-18
Add some cmake rules for installing headers, mlir-tblgen, and mlir-opt

schweitzpgi:
This is a re-work of the pull-request #219 that reduces the size of the overall change, etc.

The purpose is to install MLIR header files in the `include` tree of the installation and facilitate using MLIR in out-of-tree built projects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,25 +62,20 @@ if (NOT LLVM_INSTALL_TOOLCHAIN_ONLY)
     DESTINATION include
     COMPONENT mlir-headers
     FILES_MATCHING
-    PATTERN "*.def"
     PATTERN "*.h"
-    PATTERN "*.td"
     PATTERN "*.inc"
     PATTERN "LICENSE.TXT"
-    PATTERN ".svn" EXCLUDE
     )
 
   install(DIRECTORY ${MLIR_INCLUDE_DIR}/mlir ${MLIR_INCLUDE_DIR}/mlir-c
     DESTINATION include
     COMPONENT mlir-headers
     FILES_MATCHING
-    PATTERN "*.def"
     PATTERN "*.h"
     PATTERN "*.gen"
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -56,3 +56,36 @@ add_subdirectory(test)
...
 endif()
+
+if (NOT LLVM_INSTALL_TOOLCHAIN_ONLY)
+  install(DIRECTORY include/mlir include/mlir-c
+    DESTINATION include
+    COMPONENT mlir-headers
+    FILES_MATCHING
+    PATTERN "*.def"
+    PATTERN "*.h"
+    PATTERN "*.td"
```
joker-eph: Do you know why the .td are installed?

CMakeLists.txt
```diff
@@ -56,3 +56,36 @@ add_subdirectory(test)
...
+    DESTINATION include
+    COMPONENT mlir-headers
+    FILES_MATCHING
+    PATTERN "*.def"
+    PATTERN "*.h"
+    PATTERN "*.gen"
+    PATTERN "*.inc"
+    PATTERN "CMakeFiles" EXCLUDE
+    PATTERN "config.h" EXCLUDE
+    PATTERN ".svn" EXCLUDE
```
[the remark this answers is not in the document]
schweitzpgi: Yes. This was copied from LLVM.

CMakeLists.txt
schweitzpgi: I don't.

joker-eph: Let's skip until we have a use-case then?
It'll avoid creating a "Chesterton's fence" :)

(and let's remove .svn as well as .def files, unless we have .def files?)

schweitzpgi: Heh.

Removed .svn. .td, and .def.

PR #247, opened 2019-11-18
SmallVector->SmallVectorImpl

tjingrant:
LLVM ProgrammersManual indicates that `SmallVectorImpl` is preferred over `SmallVector`. Conforming to the manual may be a good idea.


joker-eph:
I took liberty to rewrite the commit message to "Use SmallVectorImpl instead of SmallVector for function parameters (NFC)" when landing, hope you won't mind!

tjingrant:
@joker-eph thanks! Not at all. Will take note to write better commit messages next time!

PR #252, opened 2019-11-20
Make isValidSymbol more powerful

bondhugula:
The check in isValidSymbol, as far as a DimOp result went, checked if
the dim op was on a top-level memref. However, any alloc'ed, view, or
subview memref would be fine as long as the corresponding dimension of
that memref is either a static one or was in turn created using a valid
symbol in the case of dynamic dimensions.

Reported-by: Jose Gomez

### DIFF
--- a/include/mlir/Dialect/StandardOps/Ops.td
+++ b/include/mlir/Dialect/StandardOps/Ops.td
@@ -208,6 +208,9 @@ def AllocOp : Std_Op<"alloc"> {
     operand_range getSymbolicOperands() {
       return {operand_begin() + getType().getNumDynamicDims(), operand_end()};
     }
+
+    /// Returns the dynamic sizes for this alloc operation if specified.
+    operand_range getDynamicSizes() { return getOperands(); }
   }];
 
   let hasCanonicalizer = 1;
--- a/lib/Dialect/AffineOps/AffineOps.cpp
+++ b/lib/Dialect/AffineOps/AffineOps.cpp
@@ -105,8 +105,9 @@ static bool isFunctionRegion(Region *region) {
 }
 
 /// A utility function to check if a value is defined at the top level of a
-/// function. A value defined at the top level is always a valid symbol.
-bool mlir::isTopLevelSymbol(Value *value) {
+/// function. A value of index type defined at the top level is always a valid
+/// symbol.
+bool mlir::isTopLevelValue(Value *value) {
   if (auto *arg = dyn_cast<BlockArgument>(value))
     return isFunctionRegion(arg->getOwner()->getParent());
   return isFunctionRegion(value->getDefiningOp()->getParentRegion());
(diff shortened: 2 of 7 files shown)


lib/Dialect/AffineOps/AffineOps.cpp
```diff
@@ -130,13 +131,44 @@ bool mlir::isValidDim(Value *value) {
...
+/// Returns true if the result of the dim op is a valid symbol.
+static bool isDimOpValidSymbol(DimOp dimOp) {
+  // The dim op is okay if its operand memref/tensor is defined at the top
+  // level.
+  if (isTopLevelValue(dimOp.getOperand()))
+    return true;
+  // The dim op is also okay if its operand memref/tensor is a view/subview
+  // whose corresponding size is a valid symbol.
+  if (auto viewOp = dyn_cast<ViewOp>(dimOp.getOperand()->getDefiningOp()))
+    return isMemRefSizeValidSymbol<ViewOp>(viewOp, dimOp.getIndex());
```
nicolasvasilache: factor out `dimOp.getIndex()`?

lib/Dialect/AffineOps/AffineOps.cpp
```diff
@@ -130,13 +131,44 @@ bool mlir::isValidDim(Value *value) {
...
+  if (!ShapedType::isDynamic(memRefType.getDimSize(index)))
+    return true;
+  // Get the position of the dimension among dynamic dimensions;
+  unsigned dynamicDimPos = memRefType.getDynamicDimIndex(index);
+  return isValidSymbol(
+      *(memrefDefOp.getDynamicSizes().begin() + dynamicDimPos));
+}
+
+/// Returns true if the result of the dim op is a valid symbol.
+static bool isDimOpValidSymbol(DimOp dimOp) {
```
nicolasvasilache: While this function is correct, let me point out that when used with affine transformations, one will probably also need to ensure preconditions on offsets and strides of the `ViewOp` and `SubViewOp`.

This is outside of the scope of this PR but still useful to raise.

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -207,6 +207,11 @@ def AllocOp : Std_Op<"alloc"> {
     operand_range getSymbolicOperands() {
       return {operand_begin() + getType().getNumDynamicDims(), operand_end()};
     }
+
+    /// Returns the dynamic sizes for this alloc operation if specified.
+    operand_range getDynamicSizes() {
+      return {operand_begin(), operand_end()};
```
River707: getOperands()

lib/IR/StandardTypes.cpp
```diff
@@ -152,6 +152,17 @@ int64_t ShapedType::getDimSize(int64_t i) const {
   return getShape()[i];
 }
 
+unsigned ShapedType::getDynamicDimIndex(unsigned absoluteIndex) const {
+  assert(absoluteIndex < getRank() && "invalid index");
+  assert(ShapedType::isDynamic(getDimSize(absoluteIndex)) && "invalid index");
+  unsigned dynamicDimPos = 0;
+  for (unsigned i = 0; i < absoluteIndex; ++i) {
```
River707: return llvm::count_if(getShape().take_front(absoluteIndex), ShapedType::isDynamic);

bondhugula: Thanks, done.

nicolasvasilache:
Blocked by #254, fix is in flight.

nicolasvasilache:
Fix has landed, please rebase and I'll push it internally (may be some delay this is the first time I try this part of the flow).

bondhugula:
Anything more here? The build is failing due to an issue unrelated to this PR.

PR #257, opened 2019-11-21
Unify vector op names with other dialects.

Kayjukh:
Change vector op names from `VectorFooOp` to `Vector_FooOp` and from `vector::VectorFooOp` to `vector::FooOp`.

### DIFF
--- a/include/mlir/Dialect/VectorOps/VectorOps.td
+++ b/include/mlir/Dialect/VectorOps/VectorOps.td
@@ -53,7 +53,7 @@ class Vector_Op<string mnemonic, list<OpTrait> traits = []> :
 // with operators other than the current set: {*, +}.
 // TODO(andydavis) Consider using AffineMaps to express contracting, batch
 // and free dimension pairs.
-def VectorContractionOp :
+def Vector_ContractionOp :
   Vector_Op<"contract", [NoSideEffect]>,
     Arguments<(ins AnyVector:$lhs, AnyVector:$rhs, AnyVector:$acc,
                Variadic<TupleOf<[Index]>>:$masks)>,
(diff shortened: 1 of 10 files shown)


Kayjukh:
I just followed what was done in commit 1e9bdfc for the GPU dialect. It might be a better fit since it avoids the redundant dialect name in the operation.

nicolasvasilache:
Please rebase on 3f37f5278b304fdf79b9f56d625b39e9c9d21b96 and I'll start the integration process.

Kayjukh:
I just rebased on the latest commit on master c203743.

nicolasvasilache:
can you please rebase and make sure to run clang-format?
The following 3 files are flagged:
include/mlir/EDSC/Intrinsics.h
lib/Conversion/VectorConversions/VectorToVector.cpp
lib/Transforms/MaterializeVectors.cpp

Kayjukh:
I just rebased and ran `clang-format`, but it does not find anything to format. Could this be a version mismatch? I'm using
```
clang-format version 9.0.0 (tags/RELEASE_900/final)
```

ftynse:
I ran `clang-format -i -style=file` on your changes, there was a bunch of those. On the general basis, we are using clang-format compiled from LLVM's HEAD.

Kayjukh:
@nicolasvasilache Interestingly enough, I cannot manage to get any format changes from the files you pointed out to me.

@ftynse However, if I run
```sh
find . -name "*.cpp" | xargs clang-format -i -style=file
find . -name "*.h" | xargs clang-format -i -style=file
```
in the root of the repository, it changes a lot of files, most of which I did not even touch. A quick `git diff` shows that these changes are mostly related to the [`AlwaysBreakTemplateDeclarations: Yes`](https://github.com/tensorflow/mlir/blob/master/.clang-format#L2) in the `.clang-format` file. Are we using the same `.clang-format`?

PR #162, opened 2019-10-01
Add support for some multi-store cases in affine loop fusion

dcaballe:
Hello!

We've been giving affine loop fusion a try and we are pretty happy with the initial results! Great work! We are seeing quite a few loop nests being fused in our models!

We noticed that currently only single-store producer loops are supported and would like to contribute a fix for that since it seems to be an important limitation. Even though our loop nests usually have a single store, this limitation is exposed when several single-store loop nests can be fused. The following example is a snippet of a model we are working on, where 4 loop nests could be fused into a single one:

```
func @main(%in : memref<1048576x256xf32>, %out : memref<1048576x256xf32>) {
  %cst = constant 0.000000e+00 : f32

  %6 = alloc() : memref<1048576x256xf32>
  affine.for %arg7 = 0 to 1048576 {
    affine.for %arg8 = 0 to 256 {
      %13 = affine.load %in[%arg7, %arg8] : memref<1048576x256xf32>
      %14 = affine.load %in[%arg7, %arg8] : memref<1048576x256xf32>
      %15 = mulf %14, %13 : f32
      affine.store %15, %6[%arg7, %arg8] : memref<1048576x256xf32>
    }
  }
  %7 = alloc() : memref<1048576x256xf32>
  affine.for %arg7 = 0 to 1048576 {
    affine.for %arg8 = 0 to 256 {
      %13 = affine.load %6[%arg7, %arg8] : memref<1048576x256xf32>
      %14 = cmpf "ogt", %13, %cst : f32
      %15 = select %14, %13, %cst : f32
      affine.store %15, %7[%arg7, %arg8] : memref<1048576x256xf32>
    }
  }
  %8 = alloc() : memref<1048576x256xf32>
  affine.for %arg7 = 0 to 1048576 {
    affine.for %arg8 = 0 to 256 {
      %13 = affine.load %7[%arg7, %arg8] : memref<1048576x256xf32>
      %14 = affine.load %7[%arg7, %arg8] : memref<1048576x256xf32>
      %15 = mulf %14, %13 : f32
      affine.store %15, %8[%arg7, %arg8] : memref<1048576x256xf32>
    }
  }
  affine.for %arg7 = 0 to 1048576 {
    affine.for %arg8 = 0 to 256 {
      %13 = affine.load %8[%arg7, %arg8] : memref<1048576x256xf32>
      %14 = cmpf "ogt", %13, %cst : f32
      %15 = select %14, %13, %cst : f32
[...]

PR #259, opened 2019-11-22
AffineLoopFusion: Prevent fusion of multi-out-edge producer loops

dcaballe:
In https://github.com/tensorflow/mlir/pull/162 I introduced a bug that
incorrectly allowed fusion of producer loops with multiple outgoing
edges. This commit fixes that problem. It also introduces a new flag to
disable sibling loop fusion so that we can test producer-consumer fusion
in isolation.

### DIFF
--- a/lib/Transforms/LoopFusion.cpp
+++ b/lib/Transforms/LoopFusion.cpp
@@ -57,6 +57,11 @@ static llvm::cl::opt<bool>
                         llvm::cl::desc("Enables maximal loop fusion"),
                         llvm::cl::cat(clOptionsCategory));
 
+static llvm::cl::opt<bool> clSiblingFusion(
+    "sibling-fusion", llvm::cl::init(true),
+    llvm::cl::desc("Enables fusion of sibling loops (enabled by default)"),
+    llvm::cl::cat(clOptionsCategory));
+
 /// A threshold in percent of additional computation allowed when fusing.
 static llvm::cl::opt<double> clFusionAddlComputeTolerance(
     "fusion-compute-tolerance",
@@ -1005,17 +1010,23 @@ static Value *createPrivateMemRef(AffineForOp forOp, Operation *srcStoreOpInst,
 
 // Checks if node 'srcId' can be safely fused into node 'dstId'. Node 'srcId'
 // may write to multiple memrefs but it is required that only one of them,
-// 'srcLiveOutStoreOp', have an output edge.
+// 'srcLiveOutStoreOp', has output edges.
 // Returns true if 'dstNode's read/write region to 'memref' is a super set of
-// 'srcNode's write region to 'memref'.
+// 'srcNode's write region to 'memref' and 'srcId' has only one output edge.
 // TODO(andydavis) Generalize this to handle more live in/out cases.
 static bool canFuseSrcWhichWritesToLiveOut(unsigned srcId, unsigned dstId,
                                            AffineStoreOp srcLiveOutStoreOp,
                                            MemRefDependenceGraph *mdg) {
   assert(srcLiveOutStoreOp && "Expected a valid store op");
-  assert(mdg->getOutEdgeCount(srcId) == 1 && "Expected only one output edge");
   auto *dstNode = mdg->getNode(dstId);
   Value *memref = srcLiveOutStoreOp.getMemRef();
+  unsigned memrefNumOutEdges = mdg->getOutEdgeCount(srcId, memref);
+  assert(mdg->getOutEdgeCount(srcId) == memrefNumOutEdges &&
+         "Expected only outgoing edges from memref in srcId");
+
+  // Return false if 'srcNode' has more than one output edge on 'memref'.
+  if (memrefNumOutEdges != 1)
+    return false;
 
   // Compute MemRefRegion 'srcWriteRegion' for 'srcStoreOp' on 'memref'.
   MemRefRegion srcWriteRegion(srcLiveOutStoreOp.getLoc());
@@ -1459,8 +1470,10 @@ public:
   // *) Third pass fuses any remaining producer nodes into their users.
   void run() {
     // TODO(andydavis) Run this repeatedly until a fixed-point is reached.
-    fuseProducerConsumerNodes(/*maxSrcUserCount=*/1);
-    fuseSiblingNodes();
+    if (clSiblingFusion) {
+      fuseProducerConsumerNodes(/*maxSrcUserCount=*/1);
+      fuseSiblingNodes();
+    }
     fuseProducerConsumerNodes(
         /*maxSrcUserCount=*/std::numeric_limits<unsigned>::max());
     eraseUnusedMemRefAllocations();
--- a/lib/Transforms/LoopFusion.cpp
+++ b/lib/Transforms/LoopFusion.cpp
@@ -1015,12 +1015,8 @@ static bool canFuseSrcWhichWritesToLiveOut(unsigned srcId, unsigned dstId,
   assert(srcLiveOutStoreOp && "Expected a valid store op");
   auto *dstNode = mdg->getNode(dstId);
   Value *memref = srcLiveOutStoreOp.getMemRef();
-  unsigned memrefNumOutEdges = mdg->getOutEdgeCount(srcId, memref);
-  assert(mdg->getOutEdgeCount(srcId) == memrefNumOutEdges &&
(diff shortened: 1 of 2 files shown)


test/Transforms/loop-fusion.mlir
```diff
@@ -2388,6 +2389,29 @@ func @mul_add_0(%arg0: memref<3x4xf32>, %arg1: memref<4x3xf32>, %arg2: memref<3x
...
   // CHECK-NEXT: }
   // CHECK-NEXT: return
+  return
+}
+
+// Verify that 'fuseProducerConsumerNodes' doesn't fuse a producer loop with
+// multiple outgoing edges. Sibling loop fusion is disabled to properly test
+// producer-consumer fusion in isolation.
+// CHECK-LABEL: func @should_not_fuse_multi_outgoing_edge_store_producer
+func @should_not_fuse_multi_outgoing_edge_store_producer() {
```
bondhugula: Producer consumer fusion by itself is expected to fuse everything here. You shouldn't need sibling fusion to be turned on for fusion here. As such, the desired output even with -sibling-fusion=false would be the fused nest. While the fused nest could be seen as a sibling fusion followed by producer consumer fusion (which is probably what you had in mind?), it can also be viewed just as a sequence of two producer consumer fusions. 

Disabling sibling fusion would be expected to stop fusion in other cases like these.

```
for %i =  ...
   affine.load %A[%i]
   affine.store %B[%i]

for %j = ...
   affine.load %A[%i]
   affine.store %C[%i]
```

There is no producer consumer relationship or reuse, but "sibling fusion" (being for input/RAR reuse) will fuse these.

lib/Transforms/LoopFusion.cpp
```diff
@@ -1459,8 +1470,10 @@ struct GreedyFusion {
   // *) Third pass fuses any remaining producer nodes into their users.
   void run() {
     // TODO(andydavis) Run this repeatedly until a fixed-point is reached.
-    fuseProducerConsumerNodes(/*maxSrcUserCount=*/1);
-    fuseSiblingNodes();
+    if (clSiblingFusion) {
+      fuseProducerConsumerNodes(/*maxSrcUserCount=*/1);
```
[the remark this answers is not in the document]
dcaballe: I assumed that if sibling fusion was disabled, only one `fuseProducerConsumerNodes` pass was needed and I left the one with the highest `maxSrcUserCount`. I now see comment in line 1472 and I'm not sure if a single pass will address all the producer-consumer cases. Could a first pass expose new producer-consumer cases that are not handled by the own pass?

test/Transforms/loop-fusion.mlir
dcaballe: Yes, that makes sense to me. A mature producer-consumer fusion algorithm should fuse loops below. However, what I'm trying to test here is that the current producer-consumer algorithm is not fusing a producer with more than one outgoing edge because that is not supported currently (note the *three* `NOSIBLING` checks). If I didn't disable sibling fusion, producer-consumer algorithm wouldn't fuse the loops but sibling fusion would and I couldn't be sure who did it.

dcaballe:
Thanks for the comments, @bondhugula!
I can't add reviewers to MLIR PRs probably because I don't have permission. Maybe @andydavis1 is also interested?

andydavis1:
Hi Diego, thanks for contributing! I can take a look at this...

lib/Transforms/LoopFusion.cpp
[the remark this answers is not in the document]
andydavis1: Yes, please put call the "fuseSiblingNodes()" under the flag "clSiblingFusion", otherwise I think users will get confused. If you need to disable producer-consumer fusion, then it should be under a separate flag.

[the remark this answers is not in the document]
andydavis1: However, we may not need the flag at all, if you can construct a test case which is illegal for sibling fusion.

// producer loop
for ...
  store %v, %a[%i]

// Consumer loop 1
for ...
  load %a[%i]
  store %v, %b[%i]

// Consumer loop 2
for ...
  load %a[%i]
  load %b[%i]

The dependence on '%b' between the two consumer nodes should prevent sibling fusion

test/Transforms/loop-fusion.mlir
andydavis1: Can you try a unit test as follows. This test would not allow sibling fusion of the two consumer loop nests because of the dependence on '%b'. 

// producer loop
for ...
store %v, %a[%i]

// Consumer loop 1
for ...
load %a[%i]
store %v, %b[%i]

// Consumer loop 2
for ...
load %a[%i]
load %b[%i]

dcaballe: Thanks! I’ve been playing with the suggested test case:
```
func @should_not_fuse_multi_outgoing_edge_store_producer2(%a : memref<1xf32>, %b : memref<1xf32>) {
    %cst = constant 0.000000e+00 : f32
    affine.for %arg0 = 0 to 1 {
      affine.store %cst, %a[%arg0] : memref<1xf32>
    }
 
    affine.for %arg0 = 0 to 1 {
      %0 = affine.load %a[%arg0] : memref<1xf32>
      affine.store %cst, %b[%arg0] : memref<1xf32>
    }

    affine.for %arg0 = 0 to 1 {
      %0 = affine.load %a[%arg0] : memref<1xf32>
      %1 = affine.load %b[%arg0] : memref<1xf32>
    }
    return
  } 
```
Producer-consumer fusion fuses the three loops. It fuses 2nd and 3rd first, then 1st becomes a producer with a single outgoing edge and it’s fused later so we cannot use this test.
However, I noticed that if I make ‘%a’ and external memref in my original test case, I can test exactly what I want and sibling fusion doesn’t kick in. I think we can proceed this way and remove `clSiblingFusion` for now, although I kind of like that flag. It's useful for debugging/independent testing.

bondhugula:
> Modified original test and removed `clSiblingFusion`.
> I also added a new test (and fixes related to it) that it asserted now. It’s a producer with multiple outgoing edges where there is a single store with a single outgoing edge.
> 
> Please, let me know if you have any other comments.
> Thanks!

I'm a bit confused here. Why should the fusion be prevented here in cases like these where it's valid (referring to the test case you added) - where you have a single producer and multiple consumers? I recall this was what you had addressed in your earlier commit. Are you introducing the limitation back? :-)

dcaballe:
> I'm a bit confused here. Why should the fusion be prevented here in cases like these where it's valid (referring to the test case you added) - where you have a single producer and multiple consumers? I recall this was what you had addressed in your earlier commit. Are you introducing the limitation back? :-)

My previous commit introduced support for multi-store producers with a single outgoing edge within the function. That is still working fine. The test I added is intact: https://github.com/tensorflow/mlir/pull/259/files/b8e7f532a846e8a4f76b04473db21e9ffb7d15f7#diff-ea810dda41c98feaab3d122f33648fb5R2327. However, as part of that change, I asserted on that producers reaching `canFuseSrcWhichWritesToLiveOut` should have only one outgoing edge (https://github.com/tensorflow/mlir/pull/162/files#diff-5b47f3417e354b0773b092edc5d8740cR1014). That is not true because a single store may have multiple outgoing edges and loads can also have outgoing edges. That is the piece I’m reverting and adding test for it, which is a bit complicated due to the producer-consumer fusion and sibling fusion interaction.

Does it make sense now? :)

bondhugula:
> > I'm a bit confused here. Why should the fusion be prevented here in cases like these where it's valid (referring to the test case you added) - where you have a single producer and multiple consumers? I recall this was what you had addressed in your earlier commit. Are you introducing the limitation back? :-)
> 
> My previous commit introduced support for multi-store producers with a single outgoing edge within the function. That is still working fine. The test I added is intact: https://github.com/tensorflow/mlir/pull/259/files/b8e7f532a846e8a4f76b04473db21e9ffb7d15f7#diff-ea810dda41c98feaab3d122f33648fb5R2327. However, as part of that change, I asserted on that producers reaching `canFuseSrcWhichWritesToLiveOut` should have only one outgoing edge (https://github.com/tensorflow/mlir/pull/162/files#diff-5b47f3417e354b0773b092edc5d8740cR1014). That is not true because a single store may have multiple outgoing edges and loads can also have outgoing edges. That is the piece I’m reverting and adding test for it, which is a bit complicated due to the producer-consumer fusion and sibling fusion interaction.
> 
> Does it make sense now? :)

I see - sounds good then.

PR #268, opened 2019-11-25
Rename MemRefType to RankedMemRefType - NFC

nmostafa:
This change renames MemRefType to RankedMemRefType to enable #261 to introduce base MemRefType class and Unranked/RankedMemRefType sub-classes.


nicolasvasilache:
Thanks for doing this! 
This one is not going to be fun as conflicts will arise often if we don't push it quick..
Can you rebase please as it's already conflicting?

nicolasvasilache:
I'll try to iterate quickly on it, please also mark it "- NFC" and rebase so I can start our internal tests.

bondhugula:
To keep names and lines short and because the ranked MemRefType is far more widespread in use and utility across the codebase, would it make sense to just continue to use MemRefType for a ranked memref and use RankErasedMemRefType or UnrankedMemRefType for an unranked one? Note that the use case for an unranked tensor type (eg. shape inference, etc. in more abstract levels of the IR) are quite different from that of an unranked memref type(?) and so it isn't strictly needed to maintain naming consistency across the two.  (IIUC, unranked memref types are more for interfacing.)

nicolasvasilache:
@bondhugula see @River707's comments in #261 which prompted this change.

bondhugula:
> @bondhugula see @River707's comments in #261 which prompted this change.

Thanks for the context - I'll post there.

nmostafa:
@nicolasvasilache, seems master was broken yesterday. Updated branch again.

nmostafa:
Agreed that a rename is not necessary. Closing.

Issue #269, opened 2019-11-26
[spirv] Running canonicalize on spirv-module leads to an assertion failure


MaheshRavishankar:
Running canonicalization on this module leads to an assertion failure

```
spv.module "Logical" "GLSL450" {
  spv.globalVariable @__builtin_var_NumWorkgroups__ built_in("NumWorkgroups") : !spv.ptr<vector<3xi32>, Input>
  spv.globalVariable @__builtin_var_WorkgroupId__ built_in("WorkgroupId") : !spv.ptr<vector<3xi32>, Input>
  func @fmul_kernel() {
    %3 = spv.constant 12 : i32
    %4 = spv.constant 32 : i32
    %5 = spv.constant 4 : i32
    %6 = spv._address_of @__builtin_var_WorkgroupId__ : !spv.ptr<vector<3xi32>, Input>
    %7 = spv.Load "Input" %6 : vector<3xi32>
    %8 = spv.CompositeExtract %7[0 : i32] : vector<3xi32>
    %9 = spv._address_of @__builtin_var_WorkgroupId__ : !spv.ptr<vector<3xi32>, Input>
    %10 = spv.Load "Input" %9 : vector<3xi32>
    %11 = spv.CompositeExtract %10[1 : i32] : vector<3xi32>
    %18 = spv._address_of @__builtin_var_NumWorkgroups__ : !spv.ptr<vector<3xi32>, Input>
    %19 = spv.Load "Input" %18 : vector<3xi32>
    %20 = spv.CompositeExtract %19[0 : i32] : vector<3xi32>
    %21 = spv._address_of @__builtin_var_NumWorkgroups__ : !spv.ptr<vector<3xi32>, Input>
    %22 = spv.Load "Input" %21 : vector<3xi32>
    %23 = spv.CompositeExtract %22[1 : i32] : vector<3xi32>
    %30 = spv.IMul %11, %4 : i32
    %31 = spv.IMul %23, %4 : i32
    spv.loop {
      spv.Branch ^bb1(%30 : i32)
    ^bb1(%32: i32):     // 2 preds: ^bb0, ^bb2
      %33 = spv.SLessThan %32, %3 : i32
      spv.BranchConditional %33, ^bb2, ^bb3
    ^bb2:       // pred: ^bb1
      %34 = spv.IMul %8, %5 : i32
      %35 = spv.IMul %20, %5 : i32
      spv.loop {
        spv.Branch ^bb1(%34 : i32)
[...]

PR #281, opened 2019-11-28
[spirv] Check that operand of `spirv::CompositeExtractOp` is constant while folding.

denis0x0D:
This patch addresses an issue https://github.com/tensorflow/mlir/issues/269
Add a check that operand of `spirv::CompositeExtractOp` is constant while folding.
@antiagainst @MaheshRavishankar can you please take a look, thanks!

### DIFF
--- a/test/Dialect/SPIRV/canonicalize.mlir
+++ b/test/Dialect/SPIRV/canonicalize.mlir
@@ -126,6 +126,17 @@ func @extract_array_interm() -> (vector<2xi32>) {
 
 // -----
 
+// CHECK-LABEL: extract_from_not_constant
+func @extract_from_not_constant() -> i32 {
+  %0 = spv.Variable : !spv.ptr<vector<3xi32>, Function>
+  %1 = spv.Load "Function" %0 : vector<3xi32>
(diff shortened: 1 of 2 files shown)


test/Dialect/SPIRV/canonicalize.mlir
```diff
@@ -126,6 +126,19 @@ func @extract_array_interm() -> (vector<2xi32>) {
 
 // -----
 
+// CHECK-LABEL: extract_from_not_constant
+func @extract_from_not_constant() -> i32 {
+  // CHECK: %[[VAR:.*]] = spv.Variable
+  // CHECK-NEXT: %[[COMPOSITE:.*]] = spv.Load "Function" %[[VAR]]
+  // CHECK-NEXT spv.CompositeExtract %[[COMPOSITE]]
```
antiagainst: Missing `:` after `CHECK-NEXT`. Actually I think we just need to check the `spv.CompositeExtract` op exists, so no need to check `spv.Variable` and `spv.Load`. :)

denis0x0D: @antiagainst thanks for review, fixed!

PR #282, opened 2019-11-29
Fixed difference in docs tutorial #1

JKIsaacLee:
annotation explains only transpose(b), however, transpose(a) also contains.


JKIsaacLee:
I'm not sure, is this syntax only transposes b<2,3> to b<3,2>?

joker-eph:
Fix looks right to me at the moment

PR #284, opened 2019-12-02
NFC - update doc, comments, vim syntax file

bondhugula:
- for the symbol rules, the code was updated but the doc wasn't.

### DIFF
--- a/utils/vim/syntax/mlir.vim
+++ b/utils/vim/syntax/mlir.vim
@@ -31,7 +31,11 @@ syn match mlirType /x\s*\zsvector/
 " Operations.
 " Core ops (not exhaustive yet).
 " TODO: the list is not exhaustive.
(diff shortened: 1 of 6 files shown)


utils/vim/syntax/mlir.vim
```diff
@@ -31,7 +31,11 @@ syn match mlirType /x\s*\zsvector/
 " Operations.
 " Core ops (not exhaustive yet).
```
ftynse: I'd rather consider introducing some regexp that try to maintain a list of defined (std) ops. Something like

```
symbol ::= ^\s+[A-Za-z][A-Za-z0-9_]
dialect-prefix ::= ^\s+[A-Za-z][A-Za-z0-9_]*\.
dialect-symbol ::= dialect-prefix symbol
```

could work given that we tend to start new Ops from new lines.

bondhugula: But we wouldn't want highlighting to work for typos like std.for etc. For now, I just included the more commonly used ones.

Issue #271, opened 2019-11-26
[Linalg] Add permutation information to tiling


nicolasvasilache:
Once #270 is landed, we can add an optional permutation map to declarative tiling transformations. 

There is a C++ component and a Tablegen component to this PR.

nicolasvasilache:
Done.

PR #288, opened 2019-12-03
[Linalg] Add permutation information to tiling

tetuante:
This patch closes issue #271.
It adds an optional permutation map to declarative tiling transformations.
The map is expressed as a list of integers.

Change-Id: I4af69f359f0e55399fa8f107526dfea26977bcad

### DIFF
--- a/include/mlir/Dialect/Linalg/Transforms/LinalgTransformPatterns.td
+++ b/include/mlir/Dialect/Linalg/Transforms/LinalgTransformPatterns.td
@@ -57,9 +57,13 @@ class TileAndFuseLinalgOp<
 // In the future, tile sizes should be derived from op properties + machine
 // description but we do not need to wait on this to start having useful
 // patterns.
-class TileLinalgOp<list<int> sizes, string value> : NativeCodeCall<
+// `permutation` is an optional parameter to specify the ordering of the
+// tiled loops. If provided, it must be a list of integers with the same number
+// of elements as `sizes`.
+class TileLinalgOp<list<int> sizes, string value, list<int> permutation=[]> : NativeCodeCall<
   "if (failed(tileLinalgOpAndSetMarker($_builder, $0, {" #
-  StrJoinInt<sizes>.result # "}, \"" # value # "\")))" #
+  StrJoinInt<sizes>.result # "}, \"" # value # "\", {" #
+  StrJoinInt<permutation>.result # "})))" #
   "  return matchFailure();">;
 
 //===----------------------------------------------------------------------===//
--- a/include/mlir/Dialect/Linalg/Utils/Utils.h
+++ b/include/mlir/Dialect/Linalg/Utils/Utils.h
@@ -134,23 +134,43 @@ struct TiledLinalgOp {
 };
 
 /// Performs standalone tiling of a single LinalgOp by `tileSizes`.
-/// Returns a struct containing the tiled loops and the cloned op if successful,
-/// llvm::None otherwise.
+/// and permute the loop nest according to `permutation`
+/// The permutation is expressed as a list of integers that specify
+/// the new ordering of the loop nest. The length of `permutation`
+/// must be equal to the length of `tileSizes`.
+/// E.g. the permutation `(i,j,k) -> (j,k,i)` will be expressed with
+/// `permutation = [1,2,0]`. All values in `permutation` must be
+/// integers, in the range 0..`tileSizes.size()` without duplications
+/// (i.e. `[1,1,2]` is an invalid permutation). An empty list
+/// states for the identity permutation.
+/// Returns a struct containing the tiled loops in the specified order
+/// and the cloned op if successful, llvm::None otherwise.
 /// When non-null, the optional pointer `folder` is used to call into the
 /// `createAndFold` builder method. If `folder` is null, the regular `create`
 /// method is called.
 llvm::Optional<TiledLinalgOp> tileLinalgOp(OpBuilder &b, LinalgOp op,
                                            ArrayRef<Value *> tileSizes,
+                                           ArrayRef<unsigned> permutation = {},
                                            OperationFolder *folder = nullptr);
 
 /// Performs standalone tiling of a single LinalgOp by constant `tileSizes`.
-/// Returns a struct containing the tiled loops and the cloned op if successful,
-/// llvm::None otherwise.
+/// and permute the loop nest according to `permutation`
+/// The permutation is expressed as a list of integers that specify
+/// the new ordering of the loop nest. The length of `permutation`
+/// must be equal to the length of `tileSizes`.
+/// E.g. the permutation `(i,j,k) -> (j,k,i)` will be expressed with
+/// `permutation = [1,2,0]`. All values in `permutation` must be
+/// integers, in the range 0..`tileSizes.size()` without duplications
+/// (i.e. `[1,1,2]` is an invalid permutation). An empty list
+/// states for the identity permutation.
+/// Returns a struct containing the tiled loops in the specified order
+/// and the cloned op if successful, llvm::None otherwise.
 /// When non-null, the optional pointer `folder` is used to call into the
 /// `createAndFold` builder method. If `folder` is null, the regular `create`
 /// method is called.
 llvm::Optional<TiledLinalgOp> tileLinalgOp(OpBuilder &b, LinalgOp op,
                                            ArrayRef<int64_t> tileSizes,
+                                           ArrayRef<unsigned> permutation = {},
                                            OperationFolder *folder = nullptr);
 
 template <typename... Args>
--- a/lib/Dialect/Linalg/Transforms/LinalgTransforms.cpp
+++ b/lib/Dialect/Linalg/Transforms/LinalgTransforms.cpp
@@ -33,11 +33,11 @@ using namespace mlir::linalg;
 const StringLiteral mlir::linalg::LinalgTransforms::kLinalgTransformMarker =
     "__internal_linalg_transform__";
 
-LogicalResult mlir::linalg::tileLinalgOpAndSetMarker(PatternRewriter &rewriter,
-                                                     Operation *op,
-                                                     ArrayRef<int64_t> sizes,
-                                                     StringRef linalgMarker) {
-  auto tileRes = tileLinalgOperation(rewriter, op, sizes);
+LogicalResult mlir::linalg::tileLinalgOpAndSetMarker(
+    PatternRewriter &rewriter, Operation *op, ArrayRef<int64_t> sizes,
+    StringRef linalgMarker, ArrayRef<unsigned> permutation) {
+  assert(permutation.size() == 0 || permutation.size() == sizes.size());
+  auto tileRes = tileLinalgOperation(rewriter, op, sizes, permutation);
   if (!tileRes)
     return failure();
   tileRes->op.setAttr(LinalgTransforms::kLinalgTransformMarker,
--- a/lib/Dialect/Linalg/Transforms/Tiling.cpp
+++ b/lib/Dialect/Linalg/Transforms/Tiling.cpp
@@ -215,10 +215,17 @@ makeTiledViews(OpBuilder &b, Location loc, LinalgOp linalgOp,
   return res;
 }
 
-llvm::Optional<TiledLinalgOp>
-mlir::linalg::tileLinalgOp(OpBuilder &b, LinalgOp op,
-                           ArrayRef<Value *> tileSizes,
-                           OperationFolder *folder) {
+void applyPermutationToLoopRanges(SmallVector<SubViewOp::Range, 4> &loopRanges,
+                                  ArrayRef<unsigned> permutation) {
+  SmallVector<SubViewOp::Range, 4> auxVec(loopRanges.size());
+  for (unsigned i = 0; i < permutation.size(); ++i)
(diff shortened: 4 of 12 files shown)


include/mlir/Dialect/Linalg/Transforms/LinalgTransformPatterns.td
```diff
@@ -57,9 +57,10 @@ class TileAndFuseLinalgOp<
 // In the future, tile sizes should be derived from op properties + machine
 // description but we do not need to wait on this to start having useful
 // patterns.
```
nicolasvasilache: Can you plz describe the permutation parameter?

include/mlir/Dialect/Linalg/Utils/Utils.h
```diff
@@ -134,23 +134,27 @@ struct TiledLinalgOp {
 };
 
 /// Performs standalone tiling of a single LinalgOp by `tileSizes`.
-/// Returns a struct containing the tiled loops and the cloned op if successful,
-/// llvm::None otherwise.
+/// and permute the loop nest according to `permutation`
```
nicolasvasilache: Can you plz explain what the permutation is and give an example of valid / invalid permutation?
Here and in other places that extend the API.

lib/Dialect/Linalg/Transforms/LinalgTransforms.cpp
```diff
@@ -36,8 +36,10 @@ const StringLiteral mlir::linalg::LinalgTransforms::kLinalgTransformMarker =
 LogicalResult mlir::linalg::tileLinalgOpAndSetMarker(PatternRewriter &rewriter,
                                                      Operation *op,
                                                      ArrayRef<int64_t> sizes,
-                                                     StringRef linalgMarker) {
-  auto tileRes = tileLinalgOperation(rewriter, op, sizes);
+                                                     StringRef linalgMarker,
+                                                     ArrayRef<int64_t> permutation) {
+  assert(permutation.size() == 0 || permutation.size() == sizes.size());
```
nicolasvasilache: Please document these preconditions in the documentation.

lib/Dialect/Linalg/Transforms/Tiling.cpp
```diff
@@ -248,6 +259,18 @@ mlir::linalg::tileLinalgOp(OpBuilder &b, LinalgOp op,
...
+
+    // If we have to apply a permutation to the tiled loop nest, we have to reorder the induction
+    // variables
+    // This permutation is the right one assuming that loopRanges have previously been permuted by (i,j,k)->(k,i,j)
+    // So this permutation should be the inversePermutation of that one: (d0,d1,d2)->(d2,d0,d1)
+    if (permutation.size() > 0) {
+      SmallVector<AffineExpr, 4> perm;
+      for (auto index : permutation)
+        perm.push_back(getAffineDimExpr(index, ScopedContext::getContext()));
+      auto invPermutationMap = inversePermutation(AffineMap::get(permutation.size(), 0, perm));
```
nicolasvasilache: Let's lift this up to the top, check `invPermutationMap` for emptiness (implicitly false) and exit gracefully with `llvm::None` please.

tetuante: Ok to lift the map building of the AffineMap to the top. But `permutation` could be legally empty to denote the identity permutation (i.e. no permutation).  So we should not exit tileLinalgOp() in that case. Then, should we then explicitly  create the identity map to always apply it  in the lambda function? Or we can guard the applyMapToValues() with an if condition?

nicolasvasilache: If the permutation is empty, you could just set the map using this: https://github.com/tensorflow/mlir/blob/master/lib/IR/AffineMap.cpp#L109

If it is not empty then use your loop based version.

Then pass that through `inversePermutation` ?

include/mlir/Dialect/Linalg/Transforms/LinalgTransformPatterns.td
tetuante: Added some further description of the paremeter

include/mlir/Dialect/Linalg/Utils/Utils.h
tetuante: Added explanations with examples and include a validation function

lib/Dialect/Linalg/Transforms/Tiling.cpp
tetuante: Done!

lib/Dialect/Linalg/Transforms/LinalgTransforms.cpp
tetuante: Added documentation in LinalgTransforms.h including these preconditions

lib/Dialect/Linalg/Transforms/Tiling.cpp
```diff
@@ -215,9 +215,28 @@ makeTiledViews(OpBuilder &b, Location loc, LinalgOp linalgOp,
...
 }
 
+bool isPermutationValid(const ArrayRef<uint64_t>& permutation) {
+  SmallVector<bool, 4> present(permutation.size(), false);
+  for (unsigned i = 0; i < permutation.size(); ++i) {
+    if (permutation[i] >= permutation.size())
+      return false;
+    present[i] = true;
+  }
+  return std::all_of(present.begin(), present.end(), [](bool v) { return v; });
```
River707: What is this doing? When are any of these ever going to be false?

lib/Dialect/Linalg/Transforms/Tiling.cpp
```diff
@@ -226,6 +245,16 @@ mlir::linalg::tileLinalgOp(OpBuilder &b, LinalgOp op,
                  op.getNumWindowLoops() ==
              tileSizes.size() &&
          "expected matching number of tile sizes and loops");
+
+  // If permutation is empty, use the identity. Build the permutation map otherwise.
+  auto invPermutationMap = AffineMap::getMultiDimIdentityMap(tileSizes.size(),ScopedContext::getContext());
```
River707: This file needs a run of clang-format.

lib/Dialect/Linalg/Transforms/Tiling.cpp
```diff
@@ -215,9 +215,28 @@ makeTiledViews(OpBuilder &b, Location loc, LinalgOp linalgOp,
   return res;
 }
 
+bool isPermutationValid(const ArrayRef<uint64_t>& permutation) {
```
nicolasvasilache: MLIR generally uses `unsigned`.

River707: ArrayRef should be passed by value, not by reference.

lib/Dialect/Linalg/Transforms/Tiling.cpp
tetuante: Check is required in case of duplicated entries in the permutation array (which is not a valid permutation). Added comments to clarify both here and in Utils.h.
Would it be better to write the loop instead of using std:all_of?
Also, is there a way to enforce these restrictions at compile time in TableGen?

lib/Dialect/Linalg/Transforms/Tiling.cpp
tetuante: Done, using clang-format-8. Is that alright?

lib/Dialect/Linalg/Transforms/Tiling.cpp
tetuante: Changed

tetuante: My bad! Done.

lib/Dialect/Linalg/Transforms/Tiling.cpp
River707: Did you mean to use present[permutation[i]]?

Also if you don't use SmallBitVector like suggested above, the std::all_of should be changed to either llvm::all_of(...) or !llvm::is_contained(present, false);

River707: This seems to point to a lack of proper test coverage.

nicolasvasilache: ok, I refrained for suggesting this earlier but given @River707's comment can we just drop this function, turn into an AffineMap (add a helper in AffineMap.cpp if necessary) and just use `isProjectedPermutationMap`?
See: https://github.com/tensorflow/mlir/blob/master/lib/IR/AffineMap.cpp#L244

ftynse:
Please run clang-format on the patch (https://github.com/tensorflow/mlir/blob/master/g3doc/DeveloperGuide.md#style-guide)

lib/Dialect/Linalg/Transforms/Tiling.cpp
ftynse: We are running the latest from LLVM repository since we depend on LLVM anyway.

lib/Dialect/Linalg/Transforms/Tiling.cpp
tetuante: The explicit for loop  was checking for out of range values in `permutation`, while marking the presence of valid values. But, if duplicates were present, some `present` elements could remain `false` after the loop.
For example, if `permutation=[0,3,1]`, the `if`condition in the  `for` loop will prevent an invalid access (since `3 > permutation.size()` ) and return false. But, if  `permutation=[0,1,1]`, then `present[2]==false` after the for loop, so std::all_of will return false.

Regarding the use of `isProjectedPermutationMap` (or `isPermutation`), there is no explicit check for out-of-bound accesses and it lets `SmallVector` fail, obscuring the reason for failure. I.e. if `permutation=[0,3,1]`, the AffineMap is successfully created, but  checking `if (seen[dim.getPosition()])` in `isProjectedPermutationMap` will access out-of-bounds triggering an `assert` in SmallVector.h.

I am happy either way, but I thought I should make you aware of this. Is it worth adding the extra guard to `isProjectedPermutationMap`?

nicolasvasilache: There is a bug in the creation of the AffineMap:  you should not do 
`AffineMap::get(permutation.size(), 0, perm)` but
`AffineMap::get(getMax(permutation), 0, perm)`

The AffineMap you describe should be:
`(d0,  d1, d2) -> (d0, d3, d1)` which is an invalid map (`d3` is an undefined coordinate dimension).

If this is the case, the potential fact that it is not caught by some verifier at the type level should be fixed.

In the meantime, this seems like an easy enough pitfall to fall into (and to miss when reviewing): let's factor out this code and have a permutation builder helper in AffineMap.h that would build a valid map from a permutation array.

Thanks!

tetuante: You are right, thank you. We have implemented a builder helper in AffineMap.h.

lib/Dialect/Linalg/Transforms/Tiling.cpp
tetuante: Thanks. Done

lib/Dialect/Linalg/Transforms/Tiling.cpp
```diff
@@ -215,17 +215,50 @@ makeTiledViews(OpBuilder &b, Location loc, LinalgOp linalgOp,
...
 
-llvm::Optional<TiledLinalgOp>
-mlir::linalg::tileLinalgOp(OpBuilder &b, LinalgOp op,
-                           ArrayRef<Value *> tileSizes,
-                           OperationFolder *folder) {
+// A permutation, expressed as list of unsigned, is not valid if:
+// - There are duplicates in the list
+// - Any element of the list is larger or equal the list size
+bool isPermutationValid(const ArrayRef<unsigned> permutation) {
+  SmallVector<bool, 4> present(permutation.size(), false);
```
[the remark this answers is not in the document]
tetuante: After comments above on this function, we have removed it.

lib/Dialect/Linalg/Transforms/Tiling.cpp
```diff
@@ -215,17 +215,34 @@ makeTiledViews(OpBuilder &b, Location loc, LinalgOp linalgOp,
...
   assert(op.getNumParallelLoops() + op.getNumReductionLoops() +
                  op.getNumWindowLoops() ==
              tileSizes.size() &&
          "expected matching number of tile sizes and loops");
+
+  // If permutation is empty, use the identity. Build the permutation map
+  // otherwise.
+  auto invPermutationMap = AffineMap::getMultiDimIdentityMap(
+      tileSizes.size(), ScopedContext::getContext());
+  if (permutation.size() > 0) {
```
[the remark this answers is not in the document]
nicolasvasilache: addressing internally while merging

lib/Dialect/Linalg/Transforms/Tiling.cpp
```diff
@@ -239,7 +256,8 @@ mlir::linalg::tileLinalgOp(OpBuilder &b, LinalgOp op,
   auto loopRanges =
       makeTiledLoopRanges(b, scope.getLocation(), viewSizesToLoopsMap,
                           viewSizes, tileSizes, folder);
-
+  if (permutation.size() > 0)
```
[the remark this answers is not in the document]
nicolasvasilache: addressing internally while merging

PR #299, opened 2019-12-06
[spirv] Reorder `erase` and `emplace` to avoid "invalid iterator access"

denis0x0D:


lib/Dialect/SPIRV/Serialization/Deserializer.cpp
```diff
@@ -1875,8 +1875,8 @@ LogicalResult ControlFlowStructurizer::structurizeImpl() {
       if (Block *mappedTo = mapper.lookupOrNull(newMerge))
         newMerge = mappedTo;
 
-      blockMergeInfo.try_emplace(newHeader, newMerge, newContinue);
       blockMergeInfo.erase(it);
+      blockMergeInfo.try_emplace(newHeader, newMerge, newContinue);
```
antiagainst: Could you add a comment before `erase` to remind future changes?

" // The iterator should be erased before adding a new entry into blockMergeInfo to avoid iterator invalidation."

denis0x0D: thanks, fixed!

PR #301, opened 2019-12-06
fix vectorize examples in comments

AlexandreEichenberger:
As I was reading the lib/Transforms/Vectorize.cpp to understand simdization in MLIR, I came across some outdated examples in the comments. I fixed the example and ran them to double check that they work.

Changes are only in comments. Hope this will be helpful to others

### DIFF
--- a/lib/Transforms/Vectorize.cpp
+++ b/lib/Transforms/Vectorize.cpp
@@ -306,10 +306,10 @@ using namespace mlir;
 /// terminal processing out of the use-def chains starting from loads. In the
 /// following snippet, there is simply no load::
 /// ```mlir
-/// mlfunc @fill(%A : memref<128xf32>) -> () {
+/// func @fill(%A : memref<128xf32>) -> () {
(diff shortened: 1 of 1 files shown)


nicolasvasilache:
Thanks much @AlexandreEichenberger.

Note that I did this work 1 year ago now and MLIR was still very young.
It certainly would benefit from a dust off and potentially major revisions.
Atm we are pushing on the vector dialect to have better abstractions that will simplify vectorization.
Still the objectives and motivations described in the comments are still very much on top of my mind.

Please let me know if you'd be interested in discussing further :)

AlexandreEichenberger:
@nicolasvasilache Definitely interested in vectorization. We have one that works quite well for conv/matmul outside of MLIR and are evaluating migrating it to MLIR. Our scheme works also on "partial" tiles where we materialize the padding on the fly (free with permute ops) as well as partial stores.  I understand that there are benefit in migrating that handling of partial tiles into the read / write from / to original input / output tensors to buffers for tiles, so not sure yet what is best.

PR #308, opened 2019-12-09
[spirv] Add CompositeConstruct operation.

denis0x0D:
This PR adds a `spv.CompositeConstruct` operation.

@antiagainst @MaheshRavishankar can you please take a look, thanks!

### DIFF
--- a/lib/Dialect/SPIRV/SPIRVOps.cpp
+++ b/lib/Dialect/SPIRV/SPIRVOps.cpp
@@ -1054,6 +1054,73 @@ static LogicalResult verify(spirv::BranchConditionalOp branchOp) {
 }
 
 //===----------------------------------------------------------------------===//
+// spv.CompositeConstruct
+//===----------------------------------------------------------------------===//
+
+static ParseResult parseCompositeConstructOp(OpAsmParser &parser,
+                                             OperationState &state) {
+  SmallVector<OpAsmParser::OperandType, 4> operands;
+  Type type;
+  auto loc = parser.getCurrentLocation();
+
+  if (parser.parseOperandList(operands) || parser.parseColonType(type)) {
+    return failure();
+  }
+  auto cType = type.dyn_cast<spirv::CompositeType>();
+  if (!cType) {
+    return parser.emitError(
+               loc, "result type must be a composite type, but provided ")
+           << type;
+  }
+
+  if (operands.size() != cType.getNumElements()) {
+    return parser.emitError(loc, "has incorrect number of operands: expected ")
+           << cType.getNumElements() << ", but provided " << operands.size();
+  }
+  // TODO: Add support for constructing a vector type from the vector operands.
+  // According to the spec: "for constructing a vector, the operands may
+  // also be vectors with the same component type as the Result Type component
+  // type".
+  SmallVector<Type, 4> elementTypes;
+  elementTypes.reserve(cType.getNumElements());
+  for (auto index : llvm::seq<uint32_t>(0, cType.getNumElements())) {
+    elementTypes.push_back(cType.getElementType(index));
+  }
+  state.addTypes(type);
+  return parser.resolveOperands(operands, elementTypes, loc, state.operands);
+}
+
+static void print(spirv::CompositeConstructOp compositeConstructOp,
+                  OpAsmPrinter &printer) {
+  printer << spirv::CompositeConstructOp::getOperationName() << " ";
+  printer.printOperands(compositeConstructOp.constituents());
+  printer << " : " << compositeConstructOp.getResult()->getType();
+}
+
+static LogicalResult verify(spirv::CompositeConstructOp compositeConstructOp) {
+  auto cType = compositeConstructOp.getType().cast<spirv::CompositeType>();
+
+  SmallVector<Value *, 4> constituents(compositeConstructOp.constituents());
+  if (constituents.size() != cType.getNumElements()) {
+    return compositeConstructOp.emitError(
+               "has incorrect number of operands: expected ")
+           << cType.getNumElements() << ", but provided "
+           << constituents.size();
+  }
+
+  for (auto index : llvm::seq<uint32_t>(0, constituents.size())) {
+    if (constituents[index]->getType() != cType.getElementType(index)) {
+      return compositeConstructOp.emitError(
+                 "operand type mismatch: expected operand type ")
+             << cType.getElementType(index) << ", but provided "
+             << constituents[index]->getType();
+    }
+  }
+
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
 // spv.CompositeExtractOp
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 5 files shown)


lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -1053,6 +1053,67 @@ static LogicalResult verify(spirv::BranchConditionalOp branchOp) {
...
+  // According to the spec: "for constructing a vector, the operands may
+  // also be vectors with the same component type as the Result Type component
+  // type".
+  SmallVector<Type, 4> elementTypes;
+  for (uint32_t i = 0, e = cType.getNumElements(); i < e; ++i) {
+    elementTypes.push_back(cType.getElementType(i));
+  }
+  return failure(
+      parser.resolveOperands(operands, elementTypes, loc, state.operands) ||
+      parser.addTypesToList(type, state.types));
```
River707: Swap the order of these:

state.addTypes(elementTypes);
return parser.resolveOperands(...);

`addTypesToList` always returns success.

test/Dialect/SPIRV/Serialization/composite-op.mlir
```diff
@@ -6,4 +6,9 @@ spv.module "Logical" "GLSL450" {
     %0 = spv.CompositeInsert %arg1, %arg0[1 : i32, 0 : i32] : !spv.array<4xf32> into !spv.struct<f32, !spv.struct<!spv.array<4xf32>, f32>>
     spv.ReturnValue %0: !spv.struct<f32, !spv.struct<!spv.array<4xf32>, f32>>
   }
+  func @composite_construct_vector(%arg0: f32, %arg1: f32, %arg2 : f32) -> vector<3xf32> {
+    // CHECK: {{%.*}} = spv.CompositeConstruct {{%.*}}, {{%.*}}, {{%.*}} : vector<3xf32>
```
River707: Drop the `{{.*}} = `

River707: Here and in the other file.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -1053,6 +1053,67 @@ static LogicalResult verify(spirv::BranchConditionalOp branchOp) {
...
+
+static void print(spirv::CompositeConstructOp compositeConstructOp,
+                  OpAsmPrinter &printer) {
+  printer << spirv::CompositeConstructOp::getOperationName() << " ";
+  printer.printOperands(compositeConstructOp.constituents());
+  printer << " : " << compositeConstructOp.getResult()->getType();
+}
+
+static LogicalResult verify(spirv::CompositeConstructOp compositeConstructOp) {
+  auto cType = compositeConstructOp.getType().dyn_cast<spirv::CompositeType>();
```
antiagainst: check null `cType` first ;)

test/Dialect/SPIRV/Serialization/composite-op.mlir
denis0x0D: thanks, fixed!

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
antiagainst: Actually, because of the verification generated by ODS, we should be fine to directly `cast` here.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -1053,6 +1053,67 @@ static LogicalResult verify(spirv::BranchConditionalOp branchOp) {
...
+static ParseResult parseCompositeConstructOp(OpAsmParser &parser,
+                                             OperationState &state) {
+  SmallVector<OpAsmParser::OperandType, 4> operands;
+  Type type;
+  auto loc = parser.getCurrentLocation();
+
+  if (parser.parseOperandList(operands) || parser.parseColonType(type)) {
+    return failure();
+  }
+  auto cType = type.dyn_cast<spirv::CompositeType>();
```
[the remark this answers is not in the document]
denis0x0D: I thought the check for composite type automatically provided by ODS,
https://github.com/tensorflow/mlir/pull/308/files#diff-e16ea79756bbf3cffd9d7410f3af4001R74
can you please fix me if I wrong.
Thanks.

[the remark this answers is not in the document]
antiagainst: That's for the verifier. For the parser the logic generated there is not invoked. :)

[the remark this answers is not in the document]
denis0x0D: thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -1053,6 +1053,67 @@ static LogicalResult verify(spirv::BranchConditionalOp branchOp) {
...
+  if (operands.size() != cType.getNumElements()) {
+    return parser.emitError(loc, "has incorrect number of operands: expected ")
+           << cType.getNumElements() << ", but provided " << operands.size();
+  }
+  // TODO: Add support for constructing a vector type from the vector operands.
+  // According to the spec: "for constructing a vector, the operands may
+  // also be vectors with the same component type as the Result Type component
+  // type".
+  SmallVector<Type, 4> elementTypes;
```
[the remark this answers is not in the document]
denis0x0D: thanks, fixed!

lib/Dialect/SPIRV/SPIRVOps.cpp
denis0x0D: thanks, fixed.

lib/Dialect/SPIRV/SPIRVOps.cpp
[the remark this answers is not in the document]
denis0x0D: I was not able to reproduce that fail, because of using generic parser instead of spirv, sorry about it.

lib/Dialect/SPIRV/SPIRVOps.cpp
```diff
@@ -1053,6 +1053,73 @@ static LogicalResult verify(spirv::BranchConditionalOp branchOp) {
...
+  elementTypes.reserve(cType.getNumElements());
+  for (uint32_t i = 0, e = cType.getNumElements(); i < e; ++i) {
```
MaheshRavishankar: Not required with this change, but I think the recommendation is to use llvm::seq<uint32_t>(cType.getNumElements)

denis0x0D: thanks, fixed. Also added a new test for result type.

PR #314, opened 2019-12-11
Fix OSS build

nicolasvasilache:


nicolasvasilache:
This has been landed internally.

Issue #295, opened 2019-12-05
[Linalg] Expose subview promotion as a declarative pattern


nicolasvasilache:

tetuante:
Working on it!

nicolasvasilache:
Done.

PR #315, opened 2019-12-12
[Linalg] Expose subview promotion as a declarative pattern

tetuante:
This PR targets issue #295. It exposes the already existing
 subiew promotion pass as a declarative pattern.

### DIFF
--- a/include/mlir/Dialect/Linalg/Transforms/LinalgTransformPatterns.td
+++ b/include/mlir/Dialect/Linalg/Transforms/LinalgTransformPatterns.td
@@ -43,6 +43,13 @@ class AffineMapDomainHasDim<int n> : CPred<[{
   $0.getAttrOfType<ArrayAttr>(getIndexingMapsAttrName()).getValue()[0].
   cast<AffineMapAttr>().getValue().getNumDims() ==}] # n # [{}]>;
 
+class HasOperandsOfType<string type>: CPred<[{
+    llvm::any_of($0.getOperands(),
+        [](Value* v) {
+          return dyn_cast_or_null<}] # type # [{>(v->getDefiningOp());
+        })
+}]>;
+
 //===----------------------------------------------------------------------===//
 // Linalg fusion patterns.
 //===----------------------------------------------------------------------===//
@@ -101,4 +108,10 @@ class PermuteGenericLinalgOp<list<int> permutation, string value> :
     StrJoinInt<permutation>.result # "}, \"" # value # "\"))) " #
     "  return matchFailure();">;
 
+//===----------------------------------------------------------------------===//
+// Linalg promote subview operands.
+//===----------------------------------------------------------------------===//
+class LinalgOpPromoteSubviews<string OpType> : NativeCodeCall<
+  "if (failed(linalgOpPromoteSubviews($_builder, $0))) " #
+  "  return matchFailure();">;
 #endif // LINALG_TRANSFORMS
--- a/include/mlir/Dialect/Linalg/Transforms/LinalgTransforms.h
+++ b/include/mlir/Dialect/Linalg/Transforms/LinalgTransforms.h
@@ -95,6 +95,10 @@ LogicalResult vectorizeGenericOp(PatternRewriter &rewriter, Operation *op);
 LogicalResult permuteGenericLinalgOp(PatternRewriter &rewriter, Operation *op,
                                      ArrayRef<unsigned> permutation,
                                      StringRef linalgMarker);
+
+/// Promote std.subviews feeding linalg operations
+LogicalResult linalgOpPromoteSubviews(PatternRewriter &rewriter, Operation *op);
+
 } // namespace linalg
 } // namespace mlir
 
--- a/include/mlir/Dialect/Linalg/Utils/Utils.h
+++ b/include/mlir/Dialect/Linalg/Utils/Utils.h
@@ -23,6 +23,8 @@
 #include "mlir/Dialect/StandardOps/Ops.h"
 #include "mlir/EDSC/Helpers.h"
 
+#include "llvm/ADT/SetVector.h"
+
 namespace mlir {
 class AffineExpr;
 class AffineMap;
@@ -217,6 +219,17 @@ void applyPermutationToVector(SmallVector<T, N> &inVec,
     auxVec[i] = inVec[permutation[i]];
   inVec = auxVec;
 }
+
+/// Prepares the SubView promotion later performed by `promoteSubViews`
+/// (where most of the transformation happens). It arranges the new
+/// operands for `LinalgOp op` and deallocates the new buffer(s)
+/// It is the entry point for declarative transformation
+/// Returns the cloned `LinalgOp` with the new operands
+LinalgOp promoteSubViewOperands(OpBuilder &b, LinalgOp op,
+                                llvm::SetVector<Value *> subViews,
+                                bool dynamicBuffers = false,
+                                OperationFolder *folder = nullptr);
+
 } // namespace linalg
 } // namespace mlir
 
--- a/lib/Dialect/Linalg/Transforms/LinalgTransforms.cpp
+++ b/lib/Dialect/Linalg/Transforms/LinalgTransforms.cpp
@@ -43,6 +43,7 @@ using namespace mlir::linalg;
 using namespace mlir::linalg::intrinsics;
 
 using llvm::dbgs;
+using llvm::SetVector;
 
 // Marker used as attribute name in generated Linalg rewriting transformations.
 const StringLiteral mlir::linalg::LinalgTransforms::kLinalgTransformMarker =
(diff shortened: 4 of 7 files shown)


test/lib/DeclarativeTransforms/TestLinalgTransformPatterns.td
```diff
@@ -126,4 +125,26 @@ def : Pat<(IndexedGenericOp:$op $_1, $_2, $_3, $_4, $_5, $_6, $_7, $_8),
...
 
+//===----------------------------------------------------------------------===//
+// Linalg subview operands promotion.
+//===----------------------------------------------------------------------===//
+def : Pat<(MatmulOp:$op $A, $B, $C),
+          (LinalgOpPromoteSubviews<"MatmulOp"> $op),
+          [(Constraint<HasOperandsOfType<"SubViewOp">> $op),
+          (Constraint<HasLinalgTransformMarker<"_promote_views_">> $op)]>;
+
+def : Pat<(MatvecOp:$op $A, $b, $c),
```
nicolasvasilache: These patterns are not used, could we drop them?

test/Dialect/Linalg/transform-patterns.mlir
```diff
@@ -315,3 +315,53 @@ func @matmul_perm(%A: memref<?x?xf32, offset: ?, strides: [?, 1]>,
...
+  %c0 = constant 0 : index
+  %c1 = constant 1 : index
+  %0 = dim %arg0, 0 : memref<?x?xf32, offset: ?, strides: [?, 1]>
+  %1 = dim %arg0, 1 : memref<?x?xf32, offset: ?, strides: [?, 1]>
+  %2 = dim %arg1, 1 : memref<?x?xf32, offset: ?, strides: [?, 1]>
+  loop.for %arg3 = %c0 to %0 step %c2000 {
+    loop.for %arg4 = %c0 to %2 step %c3000 {
+      loop.for %arg5 = %c0 to %1 step %c4000 {
+        %3 = std.subview %arg0[%arg3, %arg5][%c2000, %c4000][%c1, %c1] :
+             memref<?x?xf32, offset: ?, strides: [?, 1]> to memref<?x?xf32, offset: ?, strides: [?, ?]>
```
nicolasvasilache: The subview result type can keep the same strides as the original type.

include/mlir/Dialect/Linalg/Transforms/LinalgTransformPatterns.td
```diff
@@ -43,6 +43,13 @@ class AffineMapDomainHasDim<int n> : CPred<[{
   $0.getAttrOfType<ArrayAttr>(getIndexingMapsAttrName()).getValue()[0].
   cast<AffineMapAttr>().getValue().getNumDims() ==}] # n # [{}]>;
 
+class HasOperandsOfType<string type>: CPred<[{
+    std::any_of($0.getOperands().begin(), $0.getOperands().end(),
```
River707: llvm::any_of

include/mlir/Dialect/Linalg/Transforms/LinalgTransforms.h
```diff
@@ -95,6 +95,10 @@ LogicalResult vectorizeGenericOp(PatternRewriter &rewriter, Operation *op);
 LogicalResult permuteGenericLinalgOp(PatternRewriter &rewriter, Operation *op,
                                      ArrayRef<unsigned> permutation,
                                      StringRef linalgMarker);
+
+// Promote std.subviews feeding linalg operations
```
River707: // -> ///

include/mlir/Dialect/Linalg/Utils/Utils.h
```diff
@@ -23,6 +23,10 @@
 #include "mlir/Dialect/StandardOps/Ops.h"
 #include "mlir/EDSC/Helpers.h"
 
+#include "llvm/ADT/SetVector.h"
+
+using llvm::SetVector;
```
River707: Don't do this, don't add using directives in a header. Use it in a fully resolved context, or add this to LLVM.h

lib/Dialect/Linalg/Transforms/LinalgTransforms.cpp
```diff
@@ -230,3 +230,19 @@ mlir::linalg::permuteGenericLinalgOp(PatternRewriter &rewriter, Operation *op,
...
+LogicalResult mlir::linalg::linalgOpPromoteSubviews(PatternRewriter &rewriter,
+                                                    Operation *op) {
+  LinalgOp linOp = dyn_cast<LinalgOp>(op);
+  SetVector<Value *> subViews;
+  for (auto it : linOp.getInputsAndOutputs())
+    if (auto sv = dyn_cast_or_null<SubViewOp>(it->getDefiningOp()))
+      subViews.insert(sv);
+  if (!subViews.empty()) {
+    auto resOp = promoteSubViewOperands(rewriter, linOp, subViews);
+    if (!resOp)
```
River707: return success(resOp); ?

lib/Dialect/Linalg/Transforms/Promotion.cpp
```diff
@@ -189,20 +189,22 @@ static void promoteSubViewOperands(LinalgOp op, SetVector<Value *> subViews,
...
+        res.getIndexOfOutput(viewAndPartialLocalView.first).hasValue();
     if (isOutput)
       copy(viewAndPartialLocalView.second, viewAndPartialLocalView.first);
   }
 
   // 4. Dealloc local buffers.
   for (const auto &pi : promotedBufferAndViews)
     dealloc(pi.buffer);
+
+  return res; // std::move?
```
River707: Drop the comment here.

test/Dialect/Linalg/transform-patterns.mlir
tetuante: When we tried that (`strides : [?,1]` for the output of the  `subview`), I got the error:
`transform-patterns.mlir:333:14: error: expected result type to have dynamic strides` 
Am I missing something ?

MaheshRavishankar: THe current specification of subview op doesnt allow mixing and matching static and dynamic strides. Either they should be all static or all dynamic.  So you cannot have [?, ?].
In this case even though the stride operands are constants, the base memref has dynamic stride. So the subview op result must have all strides dynamic.
To fix this, the specification of subview op has to change to allow for this more general representation.

lib/Dialect/Linalg/Transforms/Promotion.cpp
tetuante: Done. Thanks

lib/Dialect/Linalg/Transforms/LinalgTransforms.cpp
tetuante: Much better.Thanks

include/mlir/Dialect/Linalg/Transforms/LinalgTransforms.h
tetuante: Done

include/mlir/Dialect/Linalg/Transforms/LinalgTransformPatterns.td
tetuante: Much better. Thank you.

include/mlir/Dialect/Linalg/Utils/Utils.h
tetuante: Oooops! Done

test/lib/DeclarativeTransforms/TestLinalgTransformPatterns.td
tetuante: Is it worth keeping them and adding some tests for them?

test/Dialect/Linalg/transform-patterns.mlir
nicolasvasilache: ah yes, forgot about that :) thanks!

test/lib/DeclarativeTransforms/TestLinalgTransformPatterns.td
nicolasvasilache: I wouldn't think so, it's very easy to copy-paste the MatmulOp and adapt.

tetuante: OK. Done.

nicolasvasilache:
@River707 I believe your comments have been addressed.

lib/Dialect/Linalg/Transforms/Promotion.cpp
```diff
@@ -189,20 +189,22 @@ static void promoteSubViewOperands(LinalgOp op, SetVector<Value *> subViews,
...
   opViews.append(operands.begin(), operands.end());
-  op.clone(b, op.getLoc(), opViews);
+  LinalgOp res = op.clone(b, op.getLoc(), opViews);
 
   // 3. Emit write-back for the promoted output views: copy the partial view.
   for (auto viewAndPartialLocalView : writebackViews) {
     // Note: use the old op to determine whether the operand view is an output.
     bool isOutput =
-        op.getIndexOfOutput(viewAndPartialLocalView.first).hasValue();
+        res.getIndexOfOutput(viewAndPartialLocalView.first).hasValue();
```
nicolasvasilache: This is a bug that prevent writebacks from happening.
As the comment right above says, you want to check the old op not the newly cloned one.
I'll make it into a WARNING.

Fixing internally for submitting.

Issue #318, opened 2019-12-14
Build errors - NVVMConversions, ROCDLConversions


bondhugula:
The trunk doesn't build (both release and debug builds).  (LLVM trunk: ed83942bc02501c62b55f77cba688263ea35221a )
```
/usr/lib64/ccache/clang++  -DGTEST_HAS_RTTI=0 -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -Iprojects/mlir/lib/Target -I/home/uday/llvm-project/llvm/projects/mlir/lib/Target -I/usr/include/libxml2 -Iinclude -I/home/uday/llvm-project/llvm/include -I/home/uday/llvm-project/llvm/projects/mlir/include -Iprojects/mlir/include -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wstring-conversion -fdiagnostics-color -ffunction-sections -fdata-sections -O2 -DNDEBUG    -fno-exceptions -fno-rtti -std=c++14 -MD -MT projects/mlir/lib/Target/CMakeFiles/MLIRTargetNVVMIR.dir/LLVMIR/ConvertToNVVMIR.cpp.o -MF projects/mlir/lib/Target/CMakeFiles/MLIRTargetNVVMIR.dir/LLVMIR/ConvertToNVVMIR.cpp.o.d -o projects/mlir/lib/Target/CMakeFiles/MLIRTargetNVVMIR.dir/LLVMIR/ConvertToNVVMIR.cpp.o -c /home/uday/llvm-project/llvm/projects/mlir/lib/Target/LLVMIR/ConvertToNVVMIR.cpp
/home/uday/llvm-project/llvm/projects/mlir/lib/Target/LLVMIR/ConvertToNVVMIR.cpp:51:55: error: no member named 'nvvm_shfl_sync_bfly_f32p' in namespace 'llvm::Intrinsic'
    return resultType->isFloatTy() ? llvm::Intrinsic::nvvm_shfl_sync_bfly_f32p
                                     ~~~~~~~~~~~~~~~~~^
[...]

PR #319, opened 2019-12-15
Fix Issue #318

Robertorosmaninho:
Fix bug on the build of the trunk version. The contents of Instrisic namespace were not being found.


joker-eph:
Thanks for the contribution!
At the moment, we fix these discrepancy internally when we bring a new version of LLVM. We are usually behind LLVM HEAD for one or two days at most.

PR #320, opened 2019-12-15
fix a typo in OpDefinitions doc

jinmingjian:
[{ matched with }], rather than ]}


jinmingjian:
@googlebot I signed it!

PR #328, opened 2019-12-18
1.the type of op.getIndex() is unsigned, the value is always greater …

leike666666:
…than or equal to 0

2.the function hasConstantoffsetSizesAndStrides is defined but not used


joker-eph:
Thanks for the fixes!
In general please split unrelated changes in multiple pull-requests.

It seems that `getIndex() ` was changed to return a signed-type just yesterday though, can you just drop this change from this PR? (and update the description accordingly)

PR #329, opened 2019-12-18
Allow dialect to create friendly names for region arguments

flaub:
This supersedes PR #214

### DIFF
--- a/include/mlir/IR/OpImplementation.h
+++ b/include/mlir/IR/OpImplementation.h
@@ -567,7 +567,8 @@ public:
   virtual ParseResult parseColonType(Type &result) = 0;
 
   /// Parse a colon followed by a type of a specific kind, e.g. a FunctionType.
-  template <typename TypeType> ParseResult parseColonType(TypeType &result) {
+  template <typename TypeType>
+  ParseResult parseColonType(TypeType &result) {
     llvm::SMLoc loc = getCurrentLocation();
 
     // Parse any kind of type.
@@ -655,6 +656,13 @@ public:
   /// OpAsmInterface.td#getAsmResultNames for usage details and documentation.
   virtual void getAsmResultNames(Operation *op,
                                  OpAsmSetValueNameFn setNameFn) const {}
+
+  /// Get a special name to use when printing the entry block arguments of the
+  /// region contained by an operation in this dialect.  The desired name should
+  /// be streamed into 'os'.  If the result is empty, the default name will be
+  /// used.
+  virtual void getRegionArgumentName(BlockArgument *arg,
+                                     raw_ostream &os) const {}
 };
 
 //===--------------------------------------------------------------------===//
--- a/include/mlir/IR/OpImplementation.h
+++ b/include/mlir/IR/OpImplementation.h
@@ -658,11 +658,9 @@ public:
                                  OpAsmSetValueNameFn setNameFn) const {}
 
   /// Get a special name to use when printing the entry block arguments of the
-  /// region contained by an operation in this dialect.  The desired name should
-  /// be streamed into 'os'.  If the result is empty, the default name will be
-  /// used.
-  virtual void getRegionArgumentName(BlockArgument *arg,
-                                     raw_ostream &os) const {}
+  /// region contained by an operation in this dialect.
+  virtual void getAsmBlockArgumentNames(Block *block,
+                                        OpAsmSetValueNameFn setNameFn) const {}
 };
 
 //===--------------------------------------------------------------------===//
--- a/lib/IR/AsmPrinter.cpp
+++ b/lib/IR/AsmPrinter.cpp
@@ -1619,28 +1619,30 @@ void OperationPrinter::numberValuesInRegion(Region &region) {
 }
 
 void OperationPrinter::numberValuesInBlock(Block &block) {
+  auto setArgNameFn = [&](Value *arg, StringRef name) {
+    assert(!valueIDs.count(arg) && "arg numbered multiple times");
+    assert(cast<BlockArgument>(arg)->getOwner() == &block &&
+           "arg not defined in 'block'");
+    setValueName(arg, name);
+  };
+
   bool isEntryBlock = block.isEntryBlock();
+  if (isEntryBlock && state) {
+    if (auto *op = block.getParentOp()) {
+      if (auto dialectAsmInterface = state->getOpAsmInterface(op->getDialect()))
+        dialectAsmInterface->getAsmBlockArgumentNames(&block, setArgNameFn);
+    }
+  }
 
   // Number the block arguments. We give entry block arguments a special name
   // 'arg'.
-  const OpAsmDialectInterface *interface = nullptr;
-  if (isEntryBlock && state) {
-    if (auto *op = block.getParentOp()) {
-      interface = state->getOpAsmInterface(op->getDialect());
-    }
-  }
-  for (auto *arg : block.getArguments()) {
-    SmallString<32> specialNameBuffer;
+  for (auto arg : block.getArguments()) {
+    SmallString<32> specialNameBuffer(isEntryBlock ? "arg" : "");
     llvm::raw_svector_ostream specialName(specialNameBuffer);
-    if (isEntryBlock) {
-      if (interface) {
-        interface->getRegionArgumentName(arg, specialName);
-      }
-      if (specialName.str().empty()) {
-        specialName << "arg" << nextArgumentID++;
-      }
-    }
-    setValueName(arg, specialName.str());
+    if (isEntryBlock)
+      specialName << nextArgumentID++;
+    if (!valueIDs.count(arg))
+      setValueName(arg, specialName.str());
   }
 
   // Number the operations in this block.
--- a/lib/IR/AsmPrinter.cpp
+++ b/lib/IR/AsmPrinter.cpp
@@ -1636,13 +1636,15 @@ void OperationPrinter::numberValuesInBlock(Block &block) {
 
   // Number the block arguments. We give entry block arguments a special name
   // 'arg'.
-  for (auto arg : block.getArguments()) {
+  SmallString<32> specialNameBuffer(isEntryBlock ? "arg" : "");
+  llvm::raw_svector_ostream specialName(specialNameBuffer);
+  for (auto *arg : block.getArguments()) {
     if (valueIDs.count(arg))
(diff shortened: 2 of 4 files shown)


include/mlir/IR/OpImplementation.h
```diff
@@ -567,7 +567,8 @@ class OpAsmParser {
   virtual ParseResult parseColonType(Type &result) = 0;
 
   /// Parse a colon followed by a type of a specific kind, e.g. a FunctionType.
-  template <typename TypeType> ParseResult parseColonType(TypeType &result) {
+  template <typename TypeType>
+  ParseResult parseColonType(TypeType &result) {
```
River707: This seems like an unrelated format change, likely caused by a different version of clang-format.

include/mlir/IR/OpImplementation.h
```diff
@@ -655,6 +656,13 @@ class OpAsmDialectInterface
   /// OpAsmInterface.td#getAsmResultNames for usage details and documentation.
   virtual void getAsmResultNames(Operation *op,
                                  OpAsmSetValueNameFn setNameFn) const {}
+
+  /// Get a special name to use when printing the entry block arguments of the
+  /// region contained by an operation in this dialect.  The desired name should
+  /// be streamed into 'os'.  If the result is empty, the default name will be
+  /// used.
+  virtual void getRegionArgumentName(BlockArgument *arg,
+                                     raw_ostream &os) const {}
```
River707: Can we set this up with the same structure as the result names? i.e. allow groups of names. That would make things more consistent and extensible.

lib/IR/AsmPrinter.cpp
```diff
@@ -1623,12 +1623,22 @@ void OperationPrinter::numberValuesInBlock(Block &block) {
 
   // Number the block arguments. We give entry block arguments a special name
   // 'arg'.
-  SmallString<32> specialNameBuffer(isEntryBlock ? "arg" : "");
-  llvm::raw_svector_ostream specialName(specialNameBuffer);
+  const OpAsmDialectInterface *interface = nullptr;
+  if (isEntryBlock && state) {
+    if (auto *op = block.getParentOp()) {
```
River707: Drop trivial braces.

include/mlir/IR/OpImplementation.h
River707: It would also be nice to add a hook to the OpAsmOpInterface, but that requires support for default implementations which is something I need to add. So I can do that as a follow up.

include/mlir/IR/OpImplementation.h
flaub: Which version of clang-format should I be using?

include/mlir/IR/OpImplementation.h
flaub: It's not clear to me how to group block argument names, but I did modify the hook to be more consistent with `getAsmResultNames`

lib/IR/AsmPrinter.cpp
```diff
@@ -1619,18 +1619,30 @@ void OperationPrinter::numberValuesInRegion(Region &region) {
...
-    if (isEntryBlock) {
-      specialNameBuffer.resize(strlen("arg"));
+  for (auto arg : block.getArguments()) {
+    SmallString<32> specialNameBuffer(isEntryBlock ? "arg" : "");
+    llvm::raw_svector_ostream specialName(specialNameBuffer);
+    if (isEntryBlock)
       specialName << nextArgumentID++;
-    }
-    setValueName(arg, specialName.str());
+    if (!valueIDs.count(arg))
```
River707: This check should be hoisted to the top of the loop:

for (BlockArgument *arg : block.getArguments()) {
  if (!...)
    continue;

test/lib/TestDialect/TestDialect.cpp
```diff
@@ -41,6 +41,19 @@ struct TestOpAsmInterface : public OpAsmDialectInterface {
     if (auto asmOp = dyn_cast<AsmDialectInterfaceOp>(op))
       setNameFn(asmOp, "result");
   }
+
+  void getAsmBlockArgumentNames(Block *block,
+                                OpAsmSetValueNameFn setNameFn) const final {
+    auto op = block->getParentOp();
+    if (auto arrayAttr = op->getAttrOfType<ArrayAttr>("arg_names")) {
+      auto args = block->getArguments();
+      for (unsigned i = 0; i < std::min(arrayAttr.size(), args.size()); i++) {
```
River707: Cache the end iterator to avoid recomputation, and prefer preincrement

e = std::min(...); i != e; ++i

test/lib/TestDialect/TestDialect.cpp
```diff
@@ -41,6 +41,19 @@ struct TestOpAsmInterface : public OpAsmDialectInterface {
     if (auto asmOp = dyn_cast<AsmDialectInterfaceOp>(op))
       setNameFn(asmOp, "result");
   }
+
+  void getAsmBlockArgumentNames(Block *block,
+                                OpAsmSetValueNameFn setNameFn) const final {
+    auto op = block->getParentOp();
+    if (auto arrayAttr = op->getAttrOfType<ArrayAttr>("arg_names")) {
```
River707: Prefer early exit:

auto arrayAttr = ...;
if (!arrayAttr)
  return;

test/lib/TestDialect/TestDialect.cpp
```diff
@@ -41,6 +41,19 @@ struct TestOpAsmInterface : public OpAsmDialectInterface {
...
+      auto args = block->getArguments();
+      for (unsigned i = 0; i < std::min(arrayAttr.size(), args.size()); i++) {
+        if (auto strAttr = arrayAttr.getValue()[i].dyn_cast<StringAttr>()) {
```
River707: Drop trivial braces.

River707:
(Github won't let me respond to an outdated conversation, which is annoying).

>It's not clear to me how to group block argument names
What I had in mind for "grouping" is just how the interface with the 'setValueName' function works, not how the IR is actually printed. If we "grouped" block arguments, it would just mean that each argument should be prefixed with that name. Though, given that block arguments function a bit differently than operation results, it should be fine to have the behavior be slightly different. This difference should be clearly marked in a comment on the method though.

flaub:
I don’t get why performance would be a consideration for the TestDialect.

test/lib/TestDialect/TestDialect.cpp
flaub: Why?

test/lib/TestDialect/TestDialect.cpp
flaub: Why?

flaub:
> (Github won't let me respond to an outdated conversation, which is annoying).
> 
> > It's not clear to me how to group block argument names
> > What I had in mind for "grouping" is just how the interface with the 'setValueName' function works, not how the IR is actually printed. If we "grouped" block arguments, it would just mean that each argument should be prefixed with that name. Though, given that block arguments function a bit differently than operation results, it should be fine to have the behavior be slightly different. This difference should be clearly marked in a comment on the method though.

I haven’t found a need for this feature. Could we do this as a follow up PR?

include/mlir/IR/OpImplementation.h
flaub: I’m happy to fix this by using the same version as you. Could you share your version?

test/lib/TestDialect/TestDialect.cpp
River707: Same as the rest.

River707:
Coding standards apply across the entire codebase, the test dialect is no excuse to be inefficient/inconsistent.

test/lib/TestDialect/TestDialect.cpp
River707: This is part of the MLIR(LLVM) coding standards:
https://llvm.org/docs/CodingStandards.html

flaub:
Actually, all of this hooking is one way to achieve custom names. But really, all I want is a way for having custom names. Would it make more sense for MLIR to treat names as first class entities? What if a Value has an optional name? That’s the common piece between both this new hook and the existing op result hook. I don’t know where it would best be stored, but it at least makes sense for the textual representation.

include/mlir/IR/OpImplementation.h
flaub: This is caused by:
https://github.com/tensorflow/mlir/blob/master/.clang-format#L2

joker-eph:
> Coding standards apply across the entire codebase, the test dialect is no excuse to be inefficient/inconsistent.

I wouldn't try to be really anchored too much on performance for the test dialect, but I agree with the "consistency" argument: it is just easier if the whole codebase is following the same idioms everywhere (and when I review code I don't necessarily realize I am reading some testing only code sometimes).

lib/IR/AsmPrinter.cpp
```diff
@@ -1619,17 +1619,30 @@ void OperationPrinter::numberValuesInRegion(Region &region) {
...
-  SmallString<32> specialNameBuffer(isEntryBlock ? "arg" : "");
-  llvm::raw_svector_ostream specialName(specialNameBuffer);
-  for (auto *arg : block.getArguments()) {
-    if (isEntryBlock) {
-      specialNameBuffer.resize(strlen("arg"));
+  for (auto arg : block.getArguments()) {
+    if (valueIDs.count(arg))
+      continue;
+    SmallString<32> specialNameBuffer(isEntryBlock ? "arg" : "");
+    llvm::raw_svector_ostream specialName(specialNameBuffer);
```
River707: Seems like this loop can just be reverted modulo the `if (valueIds.count(...))` check at the beginning.

River707:
> > (Github won't let me respond to an outdated conversation, which is annoying).
> > > It's not clear to me how to group block argument names
> > > What I had in mind for "grouping" is just how the interface with the 'setValueName' function works, not how the IR is actually printed. If we "grouped" block arguments, it would just mean that each argument should be prefixed with that name. Though, given that block arguments function a bit differently than operation results, it should be fine to have the behavior be slightly different. This difference should be clearly marked in a comment on the method though.
> 
> I haven’t found a need for this feature. Could we do this as a follow up PR?

SGTM

lib/IR/AsmPrinter.cpp
flaub: OK, on it

PR #330, opened 2019-12-19
Free ReturnOp from being restricted to a FuncOp

bondhugula:
- allow return op to be used from the top level of any op with a region
  (as opposed to just from the top level of a func op)
- the return operand count/types should match the op results' (whenever
  it's not in a FuncOp's region).

This allows std.return to be used both in a declarative (FuncOp) as well as an
imperative (typical op) setting --- avoiding the need to create custom
return ops for what is standard, in cases where control is to be
transferred to the enclosing operation, and return values need to be
modeled in ops with regions.

This also enables affine.graybox to hold a list of blocks with terminators.

### DIFF
--- a/g3doc/Dialects/Standard.md
+++ b/g3doc/Dialects/Standard.md
@@ -72,10 +72,14 @@ Syntax:
 operation ::= `return` (ssa-use-list `:` type-list-no-parens)?
 ```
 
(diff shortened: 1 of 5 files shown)


bondhugula:
This was moved to MLIR in LLVM project.

PR #331, opened 2019-12-20
fix isValidDim for block arg case

bondhugula:

### DIFF
--- a/test/Transforms/slicing-utils.mlir
+++ b/test/Transforms/slicing-utils.mlir
@@ -222,13 +222,12 @@ func @slicing_test() {
 // FWDBWD-LABEL: slicing_test_2
 func @slicing_test_2() {
   %c0 = constant 0 : index
-  %c1 = constant 1 : index
   %c2 = constant 2 : index
   %c16 = constant 16 : index
-  loop.for %i0 = %c0 to %c16 step %c1 {
+  affine.for %i0 = %c0 to %c16 {
     affine.for %i1 = (i)[] -> (i)(%i0) to 10 {
       // BWD: matched: %[[b:.*]] {{.*}} backward static slice:
-      // BWD: loop.for {{.*}}
+      // BWD: affine.for {{.*}}
 
       // affine.for appears in the body of loop.for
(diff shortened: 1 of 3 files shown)


test/Transforms/slicing-utils.mlir
```diff
@@ -222,13 +222,12 @@ func @slicing_test() {
 // FWDBWD-LABEL: slicing_test_2
 func @slicing_test_2() {
   %c0 = constant 0 : index
-  %c1 = constant 1 : index
   %c2 = constant 2 : index
   %c16 = constant 16 : index
-  loop.for %i0 = %c0 to %c16 step %c1 {
```
ftynse: This change doesn't look relevant to the commit.

bondhugula: It is :-) With a loop.for %i0 isn't a valid dimensional identifier for anything inside (only affine.for IVs are).

bondhugula:
> Could you please describe what is the problem that you are fixing in the commit message? Thanks!

I was thinking the test case was sufficient - but have updated the commit message. Thanks!

test/Transforms/slicing-utils.mlir
ftynse: My bad, I missed it was used in the map in the nested `affine.for`.

ftynse:
> > Could you please describe what is the problem that you are fixing in the commit message? Thanks!
> 
> I was thinking the test case was sufficient - but have updated the commit message. Thanks!

I'm nitpicking, but as the scale of the project grows, I'm expecting some people following the development will read commit messages but will not take the time to look into the actual change. Thanks!

bondhugula:
> > > Could you please describe what is the problem that you are fixing in the commit message? Thanks!
> > 
> > 
> > I was thinking the test case was sufficient - but have updated the commit message. Thanks!
> 
> I'm nitpicking, but as the scale of the project grows, I'm expecting some people following the development will read commit messages but will not take the time to look into the actual change. Thanks!

Yes, you are right, I agree.

PR #332, opened 2019-12-20
Add missed syntax group name in Vim syntax

chelxom:
Currently some `syntax keyword` commands don't contain proper syntax group names, and the commands are ill-formed. This causes Vim to fail to read the syntax configuration of MLIR.

This PR fixes this by adding `mlirOps` as the group name for those syntax commands.


chelxom:
@googlebot I signed it!

bondhugula (approved):
Thanks for fixing this.

joker-eph:
Fixed in https://github.com/llvm/llvm-project/commit/d2a8e14177f0b2d48b38bd77004df0c99d675dd1

PR #333, opened 2019-12-20
Add zero_extendi and sign_extendi to intrinsic namespace

pruthvi1990:
This PR add's zero_extendi and sign_extendi to intrinsic namespace

### DIFF
--- a/test/EDSC/builder-api-test.cpp
+++ b/test/EDSC/builder-api-test.cpp
@@ -468,6 +468,72 @@ TEST_FUNC(insertion_in_block) {
   f.erase();
 }
 
+TEST_FUNC(zeroextendi_op_i1_to_i8) {
+  using namespace edsc;
+  using namespace edsc::intrinsics;
+  using namespace edsc::op;
+  auto i1Type = IntegerType::get(1, &globalContext());
+  auto i8Type = IntegerType::get(8, &globalContext());
+  auto memrefType = MemRefType::get({-1, -1}, i1Type, {}, 0);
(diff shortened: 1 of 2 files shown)


test/EDSC/builder-api-test.cpp
```diff
@@ -468,6 +468,72 @@ TEST_FUNC(insertion_in_block) {
...
+  auto f = makeFunction("sign_extendi_op", {}, {memrefType});
+
+  OpBuilder builder(f.getBody());
+  ScopedContext scope(builder, f.getLoc());
+  // clang-format off
+  ValueHandle zero = constant_index(0), one = constant_index(1);
+  MemRefView vA(f.getArgument(0));
+  IndexedValue A(f.getArgument(0));
+  IndexHandle i, j;
+  AffineLoopNestBuilder({&i, &j}, {zero, zero}, {one, one}, {1, 1})([&]{
```
ftynse: Could you make these tests more minimal? They end up testing the combination of AffineLoopNextBuilder, IndexedValue, accessor generator and the call you introduced. Building something like

```
func @f(%arg0: i1) -> i8 {
  %0 = sextend_i %arg0 : i1 to i8
  return %0 : i8
}
```

should be sufficient and would minimize the risk of this test being broken when irrelevant parts (affine loops, indexedValue) are changed.

ftynse (changes requested):

ftynse (approved):

pruthvi1990:
ported this PR to LLVM monorepo, https://reviews.llvm.org/D72298. 
So closing this PR

PR #335, opened 2019-12-22
Add uitofp to the standard dialect.

ombre5733:
The operation 'uitofp', which converts an unsigned integer to a floating-point
value was already part of the LLVM dialect. Contrary to its sibling 'sitofp' it
had no corresponding operation in the standard dialect.

This adds UIToFPOp to the standard dialect and a lowering to 'uitofp' in LLVM.
The work is mostly copy-paste of the existing SIToFPOp.

### DIFF
--- a/include/mlir/Dialect/StandardOps/Ops.td
+++ b/include/mlir/Dialect/StandardOps/Ops.td
@@ -1144,7 +1144,7 @@ def UnsignedShiftRightOp : IntArithmeticOp<"shift_right_unsigned"> {
 }
 
 def SIToFPOp : CastOp<"sitofp">, Arguments<(ins AnyType:$in)> {
-  let summary = "cast from integer type to floating-point";
+  let summary = "cast from signed integer type to floating-point";
   let description = [{
     Cast from a value interpreted as signed integer to the corresponding
     floating-point value. If the value cannot be exactly represented, it is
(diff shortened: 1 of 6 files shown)


include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -1532,6 +1532,24 @@ def TruncateIOp : Std_Op<"trunci", [NoSideEffect, SameOperandsAndResultShape]> {
...
+    supported.
+  }];
+
+  let extraClassDeclaration = [{
+    /// Return true if `a` and `b` are valid operand and result pairs for
+    /// the operation.
+    static bool areCastCompatible(Type a, Type b);
+  }];
+
+  let hasFolder = 0;
```
River707, line 1550: Why is this set to 0?

include/mlir/Dialect/StandardOps/Ops.td
```diff
@@ -1532,6 +1532,24 @@ def TruncateIOp : Std_Op<"trunci", [NoSideEffect, SameOperandsAndResultShape]> {
   }];
 }
 
+def UIToFPOp : CastOp<"uitofp">, Arguments<(ins AnyType:$in)> {
```
River707, line 1535: We should constrain the input to be an integer type, and the output to a floating point.

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -1984,6 +1984,15 @@ bool SIToFPOp::areCastCompatible(Type a, Type b) {
   return a.isa<IntegerType>() && b.isa<FloatType>();
 }
 
+//===----------------------------------------------------------------------===//
+// UIToFPOp
+//===----------------------------------------------------------------------===//
+
+// uitofp is applicable from integer types to float types.
```
River707, line 1991: Use /// for function comments.

lib/Dialect/StandardOps/Ops.cpp
```diff
@@ -1984,6 +1984,15 @@ bool SIToFPOp::areCastCompatible(Type a, Type b) {
   return a.isa<IntegerType>() && b.isa<FloatType>();
 }
 
+//===----------------------------------------------------------------------===//
```
River707, line 1987: Keep the operations sorted to match .td file.

River707 (changes requested):

ombre5733:
This PR was made back in the days when MLIR was not integrated into LLVM. Since submissions happen via Phabricator now, I'm closing this issue. I'll re-submit something once I have time to familiarize with Phabricator, arcanist and whatever more is needed.

Issue #309, opened 2019-12-10
llvm.noalias attribute on Memref arguments is not properly lowered


dcaballe:
We've been using 'llvm.noalias' attribute on Memrefs to propagate no-alias information of function arguments from nGraph to LLVM. This enabled some LLVM optimizations like vectorization and prevented unnecessary runtime checks on pointers, which were crucial for performance. Unfortunately, 'llvm.noalias' is not properly lowered after latest changes in Memref type. We are observing significant performance drops because of this. 

Initially, Memref descriptor in LLVM was just a pointer to the allocated data for that Memref so 'llvm.noalias' was propagated to that pointer. However, currently Memref descriptor in LLVM is a struct as follows:
```
// template <typename Elem, size_t Rank>
// struct {
//   Elem *allocatedPtr;
//   Elem *alignedPtr;
//   int64_t offset;
//   int64_t sizes[Rank]; // omitted when rank == 0
//   int64_t strides[Rank]; // omitted when rank == 0
// };
```

and 'llvm.noalias' is propagated to the pointer to Memref descriptor (struct) instead of to the actual pointers to data (allocatedPtr and alignedPtr). 

Example: 
```
func @main(%arg0 : memref<16xf32>{llvm.noalias = true}) {
  return
}
```

```
mlir-opt noalias.mlir -convert-std-to-llvm
module {
  llvm.func @main(%arg0: !llvm<"{ float*, float*, i64, [1 x i64], [1 x i64] }*"> {llvm.noalias = true}) {
    %0 = llvm.load %arg0 : !llvm<"{ float*, float*, i64, [1 x i64], [1 x i64] }*">
    llvm.return
  }
}
```

```
mlir-opt noalias.mlir -convert-std-to-llvm | mlir-translate --mlir-to-llvmir
define void @main({ float*, float*, i64, [1 x i64], [1 x i64] }* noalias %0) {
  %2 = load { float*, float*, i64, [1 x i64], [1 x i64] }, { float*, float*, i64, [1 x i64], [1 x i64] }* %0
  ret void
}
```

We would need no-alias information to be propagated to allocatedPtr/alignedPtr in Memref descriptor. I'm not sure what would be the best way to do that since `noalias` is a function argument attribute that cannot be applied to struct fields and `allocatedPtr` and `alignedPtr` do alias now. We would probably have to use alias scopes but I'm not very familiar with that. Any thoughts?

Thanks,
Diego

bondhugula:
[...]

PR #337, opened 2019-12-23
Custom lowering of MemRefs to LLVM

dcaballe:
This PR introduces the infrastructure to provide a custom Std-to-LLVM
lowering for MemRef type:
* `MemRefDescriptor` class is turned into an abstract API that defines
  the methods needed to perform the custom lowering.
* Existing `MemRefDesriptor` class is renamed to `DefaultMemRefDescriptor`.
  It provides default struct lowering implementation (NFC).
* `TestCustomMemRefLLVMLowering.cpp` implements a custom MemRef
  descriptor lowering and LLVM type converter with the basic
  functionality to lower MemRef type to a plain pointer to element type.
* `convert-memref-ops.mlir` is split into `convert-static-memref-ops.mlir`
  and `convert-dynamic-memref-ops.mlir` so that
  TestCustomMemRefLLVMLowering.cpp can be tested on all the static
  MemRef tests available.

Related discussion: https://github.com/tensorflow/mlir/issues/309

### DIFF
--- a/include/mlir/Conversion/StandardToLLVM/ConvertStandardToLLVM.h
+++ b/include/mlir/Conversion/StandardToLLVM/ConvertStandardToLLVM.h
@@ -35,6 +35,7 @@ class Type;
 
 namespace mlir {
 
+class MemRefDescriptor;
 class UnrankedMemRefType;
 
 namespace LLVM {
@@ -56,8 +57,9 @@ public:
   /// Convert a function type.  The arguments and results are converted one by
   /// one and results are packed into a wrapped LLVM IR structure type. `result`
   /// is populated with argument mapping.
-  LLVM::LLVMType convertFunctionSignature(FunctionType type, bool isVariadic,
-                                          SignatureConversion &result);
+  virtual LLVM::LLVMType convertFunctionSignature(FunctionType type,
+                                                  bool isVariadic,
+                                                  SignatureConversion &result);
 
   /// Convert a non-empty list of types to be returned from a function into a
   /// supported LLVM IR type.  In particular, if more than one values is
@@ -70,6 +72,20 @@ public:
 
   /// Returns the LLVM dialect.
   LLVM::LLVMDialect *getDialect() { return llvmDialect; }
+
+  /// Create a DefaultMemRefDescriptor object for 'value'.
+  virtual std::unique_ptr<MemRefDescriptor>
+  createMemRefDescriptor(ValuePtr value);
+
+  /// Builds IR creating an uninitialized value of the descriptor type.
+  virtual std::unique_ptr<MemRefDescriptor>
+  buildMemRefDescriptor(OpBuilder &builder, Location loc, Type descriptorType);
+  /// Builds IR creating a MemRef descriptor that represents `type` and
+  /// populates it with static shape and stride information extracted from the
+  /// type.
+  virtual std::unique_ptr<MemRefDescriptor>
+  buildStaticMemRefDescriptor(OpBuilder &builder, Location loc, MemRefType type,
+                              ValuePtr memory);
 
   /// Promote the LLVM struct representation of all MemRef descriptors to stack
   /// and use pointers to struct to avoid the complexity of the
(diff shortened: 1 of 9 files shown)


ftynse:
Wouldn't it be simpler to just provide a set of patterns that lower memref-related ops differently, here's a list https://github.com/llvm/llvm-project/blob/master/mlir/lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp#L2126

dcaballe:
> Wouldn't it be simpler to just provide a set of patterns that lower memref-related ops differently, here's a list https://github.com/llvm/llvm-project/blob/master/mlir/lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp#L2126

The problem with that approach is that it would require re-implementing half of LLVM lowering for every custom memref lowering. I see that being a problem in terms of code duplication and maintainability (every custom LLVM lowering will break every time something changes in a memref-related op or new ops are added). Basically, every custom lowering will have to be catching up with the default lowering every time that MLIR repo is moved forward in external projects. That would be really concerning. Also, based on the current implementation, I see that memref lowering is just a step in the type conversion mechanism. It's decoupled enough from the remaining part of the op lowering so that we can provide an independent abstraction for that. We shouldn't need to replicate the remaining part of the op lowering.

As of the current implementation, we could also pass `MemRefDescriptor` as a template instead of using an abstract base class. I think that would be cleaner and it wouldn't require dynamic allocation for memref descriptors.

ftynse:
It's a typical trade-off. If there is a new operation added to dialect X, somebody will have to implement conversions from it to other dialects. Whether it should be the owner of dialect X, or the owners of other dialects is up for negotiation between those owners. If the other dialects are not upstream, it's unlikely that the owner of dialect X will be able to do anything about it. (that being said, the conversion will not necessarily break, but may not support the newly added operation).

So it feels like your main concern is about the lowering being out-of-tree. If we did have an API for the descriptor, nothing would prevent the owner of the in-tree lowering from changing that API and still breaking external projects that rely on it. And it would be a compilation error, not just a properly reported "could not lower" runtime error.

bondhugula:
@dcaballe 
Normally, it'd be great to just have a single good lowering for memrefs for the LLVM dialect. I assume what you need out of tree is something different enough that the existing lowering can't be easily extended/augmented to handle? (Looks like you require a custom memref descriptor that carries more information.) It would be good to just find a way to augment the existing lowering so that the benefits are available to all paths. OTOH, if there are use cases for custom / different enough lowerings to exist separately (all still targeting the same LLVM dialect in MLIR), it'd be good to understand at least one (or better two) such use cases to be able to do the right refactoring or to decide if it'd be better off to just go with different pattern rewrites as @ftynse suggests (keeping the code isolated albeit with duplication).  Also, as @ftynse points out, even with a common API, breakage for external lowerings would still happen but the changes needed to make it working again are reduced.

dcaballe:
Thanks for the replies and sorry for the delay (coming back from break).
I'll also reply here to comments in https://github.com/tensorflow/mlir/issues/309 to unify the discussion.

> I assume what you need out of tree is something different enough that the existing lowering can't be easily extended/augmented to handle? (Looks like you require a custom memref descriptor that carries more information.)

Actually less information. I need to lower memrefs to bare pointers to be able to properly pass the `noalias` attribute to LLVM for function arguments until we have a model for aliasing in MLIR. It's similar to what we had before introducing support for strided memrefs and alignment. It's sad that we didn't have a test for that since this is impacting performance for everybody going to LLVM. Any plans on introducing performance tests using LLVM?

Also, more people in the mailing list mentioned that they would like to have this lowering for other purposes.


> This would shift the burden on whoever updates the core, not necessarily reduce it. But I understand the concern of getting out of sync.

IMO, it's usually easier to incrementally keep a feature working than to break it and try to fix it later. Actually, the issue with the `noalias` attribute is a good example to illustrate how complicate the latter can be :). 

>So I'm fine with upstreaming an additional lowering for memrefs to go to bare pointers when possible and failing otherwise. IMO, this can still be implemented as separate patterns producing simple code, sharing implementation helpers if reasonable. The choice between lowerings can be controlled by an option in the pass, similarly to what we do for malloc/alloca.

This sounds fair enough and will allow others to use it so I'll proceed with it if no objections. I'll move the discussion to Phabricator then. 

Thanks,
Diego

PR #338, opened 2019-12-24
Fix the wrong computation of dynamic strides for lowering AllocOp to LLVM

tungld:
The computation of **dynamic strides** in lowering AllocOp was wrong. 

Given a MemRefType with `sizes = [5, 10]`, it computed `strides = [5, 1]`, while the correct answer should be `strides = [10, 1]`.

### DIFF
--- a/lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
+++ b/lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
@@ -1054,14 +1054,15 @@ struct AllocOpLowering : public LLVMLegalizationPattern<AllocOp> {
     // Iterate strides in reverse order, compute runningStride and strideValues.
     auto nStrides = strides.size();
     SmallVector<Value, 4> strideValues(nStrides, nullptr);
-    for (auto indexedStride : llvm::enumerate(llvm::reverse(strides))) {
+    for (auto indexedStride : llvm::enumerate(strides)) {
       int64_t index = nStrides - 1 - indexedStride.index();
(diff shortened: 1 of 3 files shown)


lib/Conversion/StandardToLLVM/ConvertStandardToLLVM.cpp
```diff
@@ -1054,14 +1054,15 @@ struct AllocOpLowering : public LLVMLegalizationPattern<AllocOp> {
     // Iterate strides in reverse order, compute runningStride and strideValues.
     auto nStrides = strides.size();
     SmallVector<Value, 4> strideValues(nStrides, nullptr);
-    for (auto indexedStride : llvm::enumerate(llvm::reverse(strides))) {
+    for (auto indexedStride : llvm::enumerate(strides)) {
```
ftynse: Actually, `enumerate` is useless here, the body of the loop never accesses `indexedStride.value()`. Could you just rewrite the loop to iterate on `index` to remove the confusion that led to the bug in the first place?

ftynse (changes requested):
Thanks! There's one extra simplification that can be made to make the code cleaner.

tungld:
Thanks, @ftynse! I removed `enumerate`.

ftynse (approved):

ftynse:
Landed as https://github.com/llvm/llvm-project/commit/e5957ac3d7135eca95b4019eaa394703ad6112a6. 

Since MLIR moved to LLVM, please submit your following contributions through https://reviews.llvm.org. Thank you!

@joker-eph could you please close?

